PR #1, opened 2022-10-11
mDNS autoclustering

masnax:
Adds a basic CLI (from microcluster) and adds hooks for auto-clustering.

When starting a daemon, we simultaneously broadcast our expected cluster address and name, while searching for join tokens from another potential peer. When we find a matching join token, we stop the mdns lookup/broadcast and join the cluster.

When bootstrapping, we will collect any currently available mdns records and create join tokens. We will then initiate a new broadcast with a map of tokens by cluster name. After 30 seconds, we will stop broadcasting this message.

Currently there's a hard limit of 30 cluster members per broadcast loop. This is because the mdns lookup uses a channel that requires a specified length for collecting results. I'll try to see if I can make this behave a bit nicer.


stgraber:
Merging this as-is. I'll add the initial snap packaging and do some tests on this one later.

PR #5, opened 2022-10-24
Remove go-lxc dependencies

masnax:
Reimplements SetupTrust from LXD so we don't import too much from LXD. I think this is a simple enough function that we can duplicate it, not sure if we should bloat LXD's shared package with it.


microcloud/service/lxd_join.go
```diff
@@ -0,0 +1,75 @@
...
+		return fmt.Errorf("Failed generating trust certificate: %w", err)
+	}
+
+	post := api.CertificatesPost{
+		CertificatePut: cert.CertificatePut,
+		Password:       targetPassword,
+	}
+
+	err = target.CreateCertificate(post)
+	if err != nil && err.Error() != ErrCertificateExists.Error() {
```
tomponline, line 44: Does the lxd server response when certificate already exists use a specific error code as ideally we would detect it using api.StatusErrorCheck() here.

tomponline (changes requested):
Pr description is a little confusing as go-lxc is the go wrapper of liblxc not lxd's go client :)
Or is this an indirect dependency you're trying to avoid?

microcloud/service/lxd_join.go
masnax: Actually, looking into this more it looks like this error is long gone.

masnax:
> Pr description is a little confusing as go-lxc is the go wrapper of liblxc not lxd's go client :) Or is this an indirect dependency you're trying to avoid? Does this also need a go mod update?

Yeah I'm trying to avoid bringing in liblxc.

microcloud/service/lxd_join.go
masnax: Actually not, apparently I was grepping incorrectly :)

masnax: OK, I've updated the error to use `409 Conflict` if a certificate already exists.

tomponline:
Thanks, ive merged the lxd pr now.

stgraber:
@masnax I take it this needs an update for the LXD change as well as including a gomod update?

masnax:
@stgraber I've updated this now for the LXD changes. Also since the LXD changes include the new init types, I've gone ahead and added a commit changing the initialization to use the `lxd init --preseed` command, as discussed with @tomponline

tomponline (approved):
Lgtm

tomponline:
Good to see all of those indirect dependencies removed.

masnax:
Just fixed a small typo (The preseed yaml was both the last argument and being written to stdin)

PR #6, opened 2022-10-25
Init updates

masnax:
This addresses @stgraber's comments from IRC:

* There is now a helper to get the LXD client rather than calling `ConnectLXD` right away.
* Reverts initialization to use the API again. The helper here is basically `initDataNodeApply` from LXD with the project related parts stripped out.
* Uses `zfs` for the storage driver
* Bootstraps LXD with a `lxdfan0` network device

### DIFF
--- a/microcloud/service/lxd.go
+++ b/microcloud/service/lxd.go
@@ -1,22 +1,18 @@
 package service
 
 import (
-	"bytes"
-	"context"
 	"fmt"
 	"path/filepath"
 
 	"github.com/lxc/lxd/client"
+	"github.com/lxc/lxd/lxd/revert"
 	"github.com/lxc/lxd/lxd/util"
-	"github.com/lxc/lxd/shared"
 	"github.com/lxc/lxd/shared/api"
-	"gopkg.in/yaml.v2"
 )
 
 // LXDService is a LXD service.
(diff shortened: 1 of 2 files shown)


microcloud/service/lxd.go
```diff
@@ -48,30 +48,25 @@ func (s LXDService) Bootstrap() error {
 	profile := api.ProfilesPost{ProfilePut: api.ProfilePut{Devices: rootDisk}, Name: "default"}
 	storage := api.StoragePoolsPost{Name: "local", Driver: "dir"}
 
-	initData := api.InitPreseed{
-		Node: api.InitLocalPreseed{
-			ServerPut:    server,
-			StoragePools: []api.StoragePoolsPost{storage},
-			Profiles:     []api.ProfilesPost{profile},
-		},
+	initData := initDataNode{
```
tomponline: @masnax out of interest, why not use api.InitLocalPreseed here? They look the same?

tomponline (approved):
The `initDataNode` struct seems unnecessary, but I don't want to hold this PR up, so approving.

tomponline (approved):
Oh wait, you did use the newly exported structs, you just did a revert first. Makes sense.

PR #9, opened 2022-10-28
Proxy LXD and MicroCeph connections through /1.0/services

masnax:
This turned out a bit more involved than I initially thought.

Adds `/1.0/services/ceph/{rest}` and `/1.0/services/lxd/{rest}` endpoints that forward the request to `/1.0/{rest}` through the unix socket for MicroCeph and LXD, respectively. This was done so that we do not need to access the respective services' state directories from the client, and instead can request the daemon to do it for us.

There were some endpoints that required additional handling, and those have their own API endpoints on `/services/lxd/` and `/services/ceph` respectively. 

On the LXD side this was to allow handling of operations, and also to not needlessly duplicate any post-query formatting that's done in the LXD client helpers. 

On the MicroCeph side, some of the endpoints for bootstrapping and joining are not on the MicroCeph endpoint, and instead on the MicroCluster endpoint, so these endpoints also have their own handlers to forward to the correct endpoint. Additionally, some types were only accessible internally by MicroCluster, so those have been reimplemented here. We can make the MicroCluster ones public, alternatively.

---

On the database side, there is a new table `services` that holds the state directories for MicroCeph and LXD, so that the  APIs can access them.

---

The result of all of this is that the `microcloud` command now only has one directory flag `--cloud-dir`, and any connection it needs to make to LXD or MicroCeph will be made to the daemon, and then forwarded from there.


stgraber:
So a few small changes I think:
 - We should use `/1.0/services/microceph` rather than `/1.0/services/ceph` 
 - `/1.0/services/microceph/foo` should go to `/foo` on microceph, not `/1.0/foo` (makes it easier for the client and avoids the issue for internal endpoints)

microcloud/db/schema.go
```diff
@@ -0,0 +1,28 @@
+package db
```
stgraber, line 1: Why do we need a DB table for this?

The user won't be able to add new services, so if all we need is a map of service name to state directory, we can just bake that in microcloud directly as it will always be:
 - microceph => /var/snap/microceph/common/state/
 - lxd => /var/snap/lxd/common/lxd/

masnax: So just make them all constants? In that case should we do that for the MicroCloud state directory on MicroCloud and MicroCeph directory on MicroCeph as well?

masnax:
@stgraber I've now removed the table and added constants for the state directories of MicroCeph and LXD. I've also updated the endpoint paths.

Since neither the daemon nor the client now need any state directory other than their own, I've gone back to using `--state-dir` as the flag. 

Additionally, I've fixed up the logic for reporting when a peer has fully joined all services. Now the `microcloud boostrap` properly blocks until all clusters have formed.

microcloud/db/schema.go
stgraber: They should be constants in microcloud. MicroCloud will always be run as a snap, technically the same is true for MicroCeph too but not for LXD.

PR #26, opened 2023-01-04
Sleep between cluster role reshuffling when sending tokens

masnax:
On `init`, MicroCloud starts an mdns broadcast with all the join tokens for each service and peer as the text body. What this means is that all peers will attempt to join through that first initial member.

In some cases this can cause issues as a cluster rebalances roles as it grows. So this PR splits the broadcast in two. First we send out a broadcast with just 2 members, and wait for a few seconds before continuing with the rest.

### DIFF
--- a/microcloud/cmd/microcloud/main_init.go
+++ b/microcloud/cmd/microcloud/main_init.go
@@ -209,82 +209,61 @@ func Bootstrap(sh *service.ServiceHandler, peers map[string]string) error {
 		return err
 	}
 
-	tokensByName := make(map[string]map[string]string, len(peers))
+	// Initially add just 2 peers for each dqlite service to handle issues with role reshuffling while another
+	// node is joining the cluster.
+	initialSize := 2
+	var initialTokens map[string]map[string]string
+	var tokensByName map[string]map[string]string
+	if len(peers) > initialSize {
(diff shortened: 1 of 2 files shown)


tomponline:
> In some cases this can cause issues as a cluster rebalances roles as it grows.

What issues was this causing?

I ask because its my observation (from working on similar issues with the LXD test suite) that role re-balancing doesn't always occur within a set time limit, and thus adding a pause for a fixed time period may not fully solve the issue.

masnax:
> > In some cases this can cause issues as a cluster rebalances roles as it grows.
> 
> What issues was this causing?
> 
> I ask because its my observation (from working on similar issues with the LXD test suite) that role re-balancing doesn't always occur within a set time limit, and thus adding a pause for a fixed time period may not fully solve the issue.

In this case, the error was a "dqlite leader not found" error on exactly one peer. The join request seems to be completely dropped because the join token gets used up too. 

You're probably right that the sleep here is just a bandaid, but the error was pretty consistently reproducible with MicroCloud because every peer attempts to join at the almost exact same time.

I'm also not sure if it's exclusive to LXD's dqlite implementation, because I'm pretty sure the cluster got stuck joining MicroCeph too at one point.

tomponline:
I see thanks.

PR #38, opened 2023-02-24
Allow fan bridges on unusual networks

stgraber:


microcloud/service/lxd.go
```diff
@@ -89,10 +91,26 @@ func (s LXDService) Bootstrap() error {
...
 
+	// Setup networking.
+	underlay, _, err := network.DefaultGatewaySubnetV4()
+	if err != nil {
+		return fmt.Errorf("Couldn't determine Fan overlay subnet: %w", err)
+	}
+
+	underlaySize, _ := underlay.Mask.Size()
+	if underlaySize != 16 && underlaySize != 24 {
+		// Override to /16 as that will almost always lead to working Fan network.
```
tomponline, line 102: Is it worth changing the behavior in lxd itself if its usually fine?

PR #46, opened 2023-03-02
Rescan for additional cluster members

### LINKED ISSUE #31: Allow growing an existing cluster through `microcloud init`
The user should be able to re-run `microcloud init` to detect new servers and join them into the cluster.
If local storage is used, the joining servers will need to have at least one disk available for that.
Additional disks can then be added to Ceph if the user wants.

masnax:
Closes #31 

Adds `microcloud add` which allows for rescanning for new cluster members. The process works similarly to `init`. 

To facilitate communication with a system outside the cluster, `microcloud` now by default will listen on `:9443` and will serve thee `/1.0` and `/cluster/1.0` endpoints. This listener will get replaced with a proper one once the peer has joined.

Additionally the token broadcast has been updated to send a struct, which additionally includes a `LXDConfig` for assigning `MemberConfig` to the join request made by the peer. 

The `/services/disks` endpoint is now removed, to be replaced with the `source.wipe` config field for LXD storage pools.


microcloud/mdns/lookup.go
```diff
@@ -11,9 +11,16 @@ import (
...
 
 	"github.com/hashicorp/mdns"
+	"github.com/lxc/lxd/shared/api"
 	"github.com/lxc/lxd/shared/logger"
 )
 
+// JoinConfig represents configuration broadcast to signal MicroCloud to begin join operations.
+type JoinConfig struct {
+	Token     string
+	LXDConfig []api.ClusterMemberConfigKey
```
tomponline, line 21: @masnax is there any size limit in what can be broadcast?

stgraber:
Given we now have a proper listener in place, I think we should limit mDNS broadcast to just having the unconfigured servers advertise their existence over the network. The mDNS data should tell us:
 - Name
 - List of services
 - Certificate fingerprint

The sending of the join token and configuration should all be done through the temporary HTTPS listener, using the certificate fingerprint received over mDNS to give us a bit of extra security when sending out those tokens.

Does that make sense?

PR #48, opened 2023-03-08
Grow a MicroCloud cluster

### LINKED ISSUE #31: Allow growing an existing cluster through `microcloud init`

masnax:
Closes #31 

Adds microcloud add which allows for rescanning for new cluster members. The process works similarly to init.

To facilitate communication with a system outside the cluster, microcloud now by default will listen on :9443 and will serve thee /1.0 and /cluster/1.0 endpoints. This listener will get replaced with a proper one once the peer has joined.

Additionally the token broadcast has been updated to send a struct, which additionally includes a LXDConfig for assigning MemberConfig to the join request made by the peer.

The /services/disks endpoint is now removed, to be replaced with the source.wipe config field for LXD storage pools.

As per the comments in #46, I've also reworked the join logic to be through the temporary network listener instead of mDNS.


microcloud/cmd/microcloud/main_init.go
```diff
@@ -82,16 +82,21 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
...
 	services := []service.Service{*cloud, *lxd}
-	_, err = microcluster.App(context.Background(), microcluster.Args{StateDir: api.MicroCephDir})
+	app, err := microcluster.App(context.Background(), microcluster.Args{StateDir: api.MicroCephDir})
 	if err != nil {
-		logger.Info("Skipping MicroCeph service, could not detect state directory")
-	} else {
+		return err
+	}
+
+	_, err = os.Stat(app.FileSystem.ControlSocket().URL.Host)
```
tomponline, line 80: @masnax I'm interested/confused how a value containing "URL" and "Host" can be used to stat a directory.

It would make sense if the directory was named after the host, but then I'd expect it to be prefixed with a common path, but in this case it appears to be an absolute path stored in `.Host`. 

Am I missing something here?

microcloud/cmd/microcloud/main_init.go
```diff
@@ -272,25 +272,26 @@ func Bootstrap(sh *service.ServiceHandler, peers map[string]string) error {
...
-		return err
-	}
+}
 
-	tokens := map[service.ServiceType][]string{}
+func AddPeers(sh *service.ServiceHandler, peers map[string]string, localDisks map[string][]lxdAPI.ClusterMemberConfigKey) error {
+	joinConfig := map[service.ServiceType][]mdns.JoinConfig{}
 	for serviceType := range sh.Services {
-		tokens[serviceType] = make([]string, len(peers))
+		joinConfig[serviceType] = make([]mdns.JoinConfig, len(peers))
```
tomponline: I would encourage avoiding making slices a specific size, but instead make them a specific capacity (if you know it).

So here would be `joinConfig[serviceType] = make([]mdns.JoinConfig, 0, len(peers))` and then you can populate it below using `append()`. That way the code becomes less brittle if in the future an condition logic is added that means that not all items in the loop are added to the slice, as this would end up leaving zero-intialised empty item(s) at the end of the slice.

It also avoids initialising the slice with entities that are then going to get replaced immediately afterward in loop.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -550,23 +555,48 @@ func askLocalPool(auto bool, wipe bool, lxd service.LXDService) (map[string]stri
...
+		Key:    "source",
+	}
+
+	memberConfig := make(map[string][]lxdAPI.ClusterMemberConfigKey, len(selected))
+	for target, path := range selected {
+		if target == lxd.Name() {
+			if toWipe[target] != "" {
+				err := lxd.WipeDisk(target, filepath.Base(path))
+				if err != nil {
+					return nil, fmt.Errorf("Failed to add wipe disk %q on peer %q: %w", path, target, err)
```
tomponline: "disk to wipe" might be clearer to end users.

microcloud/cmd/microcloud/add.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func (c *cmdAdd) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add",
+		Short: "Begins scanning for new cluster members",
+		RunE:  c.Run,
+	}
+
+	cmd.Flags().BoolVar(&c.flagAuto, "auto", false, "Automatic setup with default configuration")
```
tomponline, line 33: Could we call this var flagAutoSetup (and any associated function arguments `autoSetup`) as earlier I saw a function that took an argument called `auto bool` and without a function comment it wasn't clear what that meant in the context of the function's name/signature.

microcloud/service/service_handler.go
```diff
@@ -182,11 +182,31 @@ func (s *ServiceHandler) Bootstrap(state *state.State) error {
 }
 
 // RunAsync runs the given hook asynchronously across all services.
-func (s *ServiceHandler) RunAsync(f func(s Service) error) error {
+func (s *ServiceHandler) RunAsync(microCloudFirst bool, f func(s Service) error) error {
```
tomponline, line 112: This arg could do with being explained in the docblock.

microcloud/api/services.go
```diff
@@ -0,0 +1,76 @@
...
+
+	serverCert, err := s.OS.ServerCert()
+	if err != nil {
+		return response.SmartError(fmt.Errorf("Failed to compare join request fingerprints: %w", err))
+	}
+
+	// Skip the join request if the fingerprint does not match what we expect.
+	if serverCert.Fingerprint() != req.Fingerprint {
+		fmt.Println("MAW")
+		//return response.SmartError(fmt.Errorf("Join request fingerprint does not match issuer config, ignoring join request"))
```
tomponline, line 49: Commented out line left in. Needs to be removed or reinstated.

tomponline (commented):
Do we need to add a version to the mdns advert/response so we can ignore responses from earlier/incompatible machines on the network (now that we are evolving the mdns advert format).

microcloud/cmd/microcloud/main_init.go
masnax:
```go
// ControlSocket returns the full path to the control.socket file that this daemon is listening on.
func (s *OS) ControlSocket() api.URL {
	return *api.NewURL().Scheme("http").Host(filepath.Join(s.StateDir, "control.socket"))
}
```

`LXD Cloud` uses this same method for passing around the control socket. It was mainly just a convenience to allow passing the control socket directly to a client function like a URL.

tomponline: I see, thank you.

microcloud/cmd/microcloud/main_init.go
masnax: This is actually gone from the final commit, but to clarify a bit about my intentions here, I actually deliberately made this a specific sized slice because this slice was populated in an asynchronous function. I believe an append would cause some race issues.

tomponline: Interesting. By async do you mean concurrently from multiple go routines? If yes, then yes appending concurrently would certainly be an issue. Do you have synchronisation to ensure that map isn't read until all possible updates have finished?

A comment here (and in similar places) would be useful to explain why this is so, as its not clear this would be called concurrently. Thanks

tomponline: Ah yes I see now that sh.RunAsync is actually running the function concurrently for all services.
Perhaps `RunAsync` could be renamed to clarify its actually `RunConcurrent`? (as async doesn't imply concurrency).

microcloud/service/microcloud.go
```diff
@@ -68,6 +73,48 @@ func (s CloudService) Join(joinConfig JoinConfig) error {
...
+				errors = append(errors, err)
+				mut.Unlock()
+				return
+			}
+
+			joinedChan <- struct{ Name string }{Name: peer}
+		}(peer, cfg)
+	}
+
+	for _, err := range errors {
```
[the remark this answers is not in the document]
masnax: Hmm, yeah now that I'm looking at this it really doesn't make much sense does it :)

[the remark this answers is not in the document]
tomponline: It makes sense if you don't want any errors ;)

microcloud/cmd/microcloud/main_init.go
masnax: Good point, yeah I often conflate the two but I shouldn't.

microcloud/service/microcloud.go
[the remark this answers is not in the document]
masnax: Ok, I've added a `joinResponse` struct here that contains `Name` and `Error` fields. Error handling was removed from this particular function, and instead errors are handled when the caller receives from the channel.

masnax:
@tomponline I've made some changes now according to your feedback. 

> Do we need to add a version to the mdns advert/response so we can ignore responses from earlier/incompatible machines on the network (now that we are evolving the mdns advert format).

I think this is a really good idea. I've added a version flag that's appended to the keyword that tells us an mDNS record is from MicroCloud (`_microcloud_v1.0`). Adding it to the payload wouldn't have worked as older systems would fail the `json.Unmarshal` call in the first place.

microcloud/service/service_handler.go
masnax: I've updated the comment and renamed the function to `RunConcurrent`

microcloud/cmd/microcloud/add.go
masnax: I've renamed the flag & associated variables to `autoSetup` and I've went ahead and done the same for `wipe` to `wipeAllDisks`

tomponline:
> Adding it to the payload wouldn't have worked as older systems would fail the `json.Unmarshal` call in the first place.

I don't think thats true. json.Unmarshal will ignore unknown fields unless https://pkg.go.dev/encoding/json#Decoder.DisallowUnknownFields is enabled.

tomponline:
> > > Adding it to the payload wouldn't have worked as older systems would fail the `json.Unmarshal` call in the first place.
> > 
> > 
> > I don't think thats true. json.Unmarshal will ignore unknown fields unless https://pkg.go.dev/encoding/json#Decoder.DisallowUnknownFields is enabled.
> 
> However in upstream MicroCloud, the response type is `map[string]string` which will fail to convert to `map[string]struct{}` with the error `json: cannot unmarshal string into Go value of type <type>`

Ah I see you've changed the type not just the fields. Is `map[string]struct{}` the old type?

masnax:
> > > > Adding it to the payload wouldn't have worked as older systems would fail the `json.Unmarshal` call in the first place.
> > > 
> > > 
> > > I don't think thats true. json.Unmarshal will ignore unknown fields unless https://pkg.go.dev/encoding/json#Decoder.DisallowUnknownFields is enabled.
> > 
> > 
> > However in upstream MicroCloud, the response type is `map[string]string` which will fail to convert to `map[string]struct{}` with the error `json: cannot unmarshal string into Go value of type <type>`
> 
> Ah I see you've changed the type not just the fields. Is `map[string]struct{}` the old type?

Oh actually forget everything I said, you're right. I thought I was still unmarshaling a whole map of structs (I was at some point but not anymore). It's actually one struct at a time now so I can totally just add Version to the `ServerInfo` struct.

microcloud/api/services.go
```diff
@@ -29,8 +29,9 @@ const MicroCephDir = "/var/snap/microceph/common/state"
 // Proxy returns a proxy endpoint with the given handler and access applied to all REST methods.
 func Proxy(name, path string, handler func(*state.State, *http.Request) response.Response) rest.Endpoint {
```
stgraber: Hmm, does this mean that any system with `microcloud` installed will just expose the entirety of the LXD and MicroCeph REST API to anyone who can guess its IP address?

That doesn't seem ideal. Even though the same is technically still possible by assembling a working cluster through MicroCloud and then accessing those APIs.

I'd prefer we limit the pre-init API to just the bare minimum we need to send join information.
I think we should also encode a server-specific secret in the mDNS message that needs to be transmitted when hitting the pre-init API. That way only someone on the same subnet can form use the API, anyone just randomly scanning the internet will find the listener but won't be able to use it without the secret.

When we add support for manually adding a server by address, we'll ask for both the address and that secret token to be given by the user.

masnax: For the manual adding case, I suppose we would need a command that generates some encoded string including the address and the certificate fingerprint on uninitialized systems, that could then be specified to the existing cluster to manually add it.

stgraber: > we already have the server certificate fingerprint in the mDNS broadcast which should precede all proxied connections, so we could just add that to the request context and then on the other side we can check that it matches before proxying the request to its service. Then once the service is clustered, we can switch to checking against the cluster cert instead?

That won't work from a security standpoint as anyone can grab the server certificate by simply initializing a TLS connection.
What we need is something that only someone listening for the mDNS would be able to get.

stgraber: > For the manual adding case, I suppose we would need a command that generates some encoded string including the address and the certificate fingerprint on uninitialized systems, that could then be specified to the existing cluster to manually add it.
> 
> EDIT: So basically a string encoded version of `ServerInfo` introduced here.
> 
> ```go
> // ServerInfo is information about the server that is broadcast over mDNS.
>  type ServerInfo struct {
>  	Version     string
>  	Name        string
>  	Address     string
>  	Services    []types.ServiceType
>  	Fingerprint string
>  }
> ```

Right, for the manual add, we can just use a mostly standard token type format, includes name, addresses and the secret.

masnax: I've now added an `AuthSecret` string that is created on each start of an uninitialized system and included in the broadcast. I didn't bother to store the secrets anywhere for now as MicroCloud doesn't currently do much after initialization. I'm thinking if we do want to do something there in the future, we would allow cluster members (recorded in the trust store) through without checking the AuthSecret, but currently it's always checked. 

The secret is stored in the request header at `X-MicroCloud-Auth` via the client proxies before the request is sent through.

tomponline:
thanks for the changes @masnax

stgraber:
@masnax oops, sorry, looks like I broke your branch with my gomod update

masnax:
@stgraber Fixed now.

stgraber:
@masnax needs a rebase post MicroOVN

masnax:
@stgraber Fixed

PR #54, opened 2023-03-16
Limit cluster formation to within a subnet

### LINKED ISSUE #34: Filter mDNS based on interface of the listening address`
When building a cluster, we should only consider machines that are connected to the same network that the initial server is listening on.
This should allow for effectively selecting a separate network/vlan for use of MicroCloud.

masnax:
Closes #34 

By default, all uninitialized machines will now include every address on every interface in their broadcast payload. They will also broadcast this over every one of those addresses. 

On the init side, we take the first record of a peer that we detect, and check if it is in a default or otherwise supplied subnet. The cluster will then be formed using these addresses for the peers. 

To facilitate this, there are some new flags and interactive bits in the CLI
* `--address` can be used to specify a listen address right away and avoid either the interactive step, or used with `--auto` instead of the default result of `NetworkInterfaceAddress()`. 
* `--subnet` works similarly. If a subnet is not otherwise given either with the flag or in the interactive CLI, a default `/24` or `/64` subnet will be used for whichever address is chosen.

The `mDNS` lookup stage now checks once per second, and at first only reports the names of detected systems. After user input, the scan is ended and if the used didn't supply `--subnet`, the user is prompted to do so, or use the default subnet. The user can also specify `all` to show all entries. The user will still be limited to just 1 selection for each peer, but the built-in table filtering can make it pretty easy to match on an interface or part of an IP.

### DIFF
--- a/microcloud/cmd/microcloud/add.go
+++ b/microcloud/cmd/microcloud/add.go
@@ -109,6 +109,26 @@ func (c *cmdAdd) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	var newPeers map[string]mdns.ServerInfo
+	askRetry("Retry selecting peers?", c.flagAutoSetup, func() error {
+		// Add the local member to the list of peers so we can select disks.
+		newPeers, err = selectPeers(s, subnet, c.flagAutoSetup, peers)
+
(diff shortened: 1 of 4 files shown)


masnax:
Oops, actually this is the wrong branch :p

PR #55, opened 2023-03-16
Limit cluster formation to within a subnet

masnax:
Closes https://github.com/canonical/microcloud/issues/34

By default, all uninitialized machines will now include every address on every interface in their broadcast payload. They will also broadcast this over every one of those addresses. These are stored in a `NetworkInfo` slice in the payload.

On the init side, we take the first record of a peer that we detect, and check if it is in a default or otherwise supplied subnet. The cluster will then be formed using these addresses for the peers. 

To facilitate this, there are some new flags and interactive bits in the CLI
* `--address` can be used to specify a listen address right away and avoid either the interactive step, or used with `--auto` instead of the default result of `NetworkInterfaceAddress()`. 
* `--subnet` works similarly. If a subnet is not otherwise given either with the flag or in the interactive CLI, a default `/24` or `/64` subnet will be used for whichever address is chosen.

The `mDNS` lookup stage now checks once per second, and at first only reports the names of detected systems. After user input, the scan is ended and if the used didn't supply `--subnet`, the user is prompted to do so, or use the default subnet. The user can also specify `all` to show all entries. The user will still be limited to just 1 selection for each peer, but the built-in table filtering can make it pretty easy to match on an interface or part of an IP.

### DIFF
--- a/microcloud/service/service_handler.go
+++ b/microcloud/service/service_handler.go
@@ -32,7 +32,7 @@ const (
 
 // ServiceHandler holds a set of services and an mdns server for communication between them.
 type ServiceHandler struct {
-	server *mdns.Server
+	servers []*mdns.Server
 
 	Services map[types.ServiceType]Service
 	Name     string
@@ -71,6 +71,7 @@ func NewServiceHandler(name string, addr string, stateDir string, debug bool, ve
 	}
 
 	return &ServiceHandler{
+		servers:  []*mdns.Server{},
 		Services: servicesMap,
 		Name:     name,
 		Address:  addr,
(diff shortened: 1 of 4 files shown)


microcloud/service/service_handler.go
```diff
@@ -97,32 +98,44 @@ func (s *ServiceHandler) Start(state *state.State) error {
...
+		info.Address = net.Address
+		bytes, err := json.Marshal(info)
+		if err != nil {
+			return fmt.Errorf("Failed to marshal server info: %w", err)
+		}
 
-	s.server, err = cloudMDNS.NewBroadcast(s.Name, s.Address, s.Port, bytes)
-	if err != nil {
-		return err
+		server, err := cloudMDNS.NewBroadcast(s.Name, net.Address, s.Port, bytes)
```
stgraber, line 121: Just a bit confused as to why we need to have info.Networks in the struct we send over mDNS.

It makes sense that we need to send one broadcast per address on the system, but I don't understand why the payload of the broadcast has to include the addresses too. Can't the receive just tell what address send the broadcast?

masnax: This way the mDNS lookup on just one network would report all of them. I suppose if there’s a lot of addresses it could be troublesome.

masnax: One thing is that the mDNS lookup seems to get a bit slow with that many entries. 2 additional peers with 5 addresses each takes about 2 seconds to report all of them.

masnax: I believe the mDNS records are slow because it keeps picking up previously-found records. If I'm right, then if we have a potential peer with significantly fewer addresses than the others, we might need to wait quite a while to get even a single record for it.

Maybe an API endpoint would be better suited for fetching the networks. So we just broadcast on every network and collect one record for each, then send an API request to each system to fetch all networks and filter down our peers based on that.

stgraber: Yeah, but if we only rely on the address we're receiving, then we're automatically filtering out all the addresses that we can't reach, that seems preferable to me.

stgraber: Otherwise we may end up showing and potentially attempting to use addresses which are not reachable or aren't directly on a shared subnet, which can then cause issues with OVN tunnels down the line.

PR #57, opened 2023-03-21
lxd: Wake up on initial startup

stgraber:
This allows for LXD to perform its initial startup, generate certificate, ... prior to MicroCloud needing it, making the bootstrap process much faster on slow systems.


stgraber:
@masnax kinda felt like I should have put something in `lxd.go` instead but wasn't sure of what exactly as I didn't want to have to instantiate a new object just to call this. Anyway, improvements welcome ;)

PR #79, opened 2023-03-29
Interactive address selection

masnax:
Now shows a table of addresses and their interfaces on the local machine. The user's selection will be used as the listen address and to lookup peers. 

After this step, the user will be prompted with a binary yes/no for whether or not to filter on the subnet of the listen address.

* If `--address` is given, that will be used in lieu of the interactive step.
* If `--auto`, the first address will be used and peer lookup will be limited to its subnet.
* If only one address is available, the interactive step will be skipped and we will go right to asking for subnet filtering.
* The `--subnet` flag is gone now

---

Additionally, if there are too few disks for distributed storage after selecting local storage, the distributed storage selection step will be entirely skipped with a warning about insufficient disks, rather than prompting the user to add disks as normal.


microcloud/cmd/microcloud/ask.go
```diff
@@ -69,6 +70,75 @@ func askMissingServices(services []types.ServiceType, stateDirs map[types.Servic
...
+
+	}
+
+	var subnet *net.IPNet
+	var err error
+	if net.ParseIP(listenAddr).To4() != nil {
+		_, subnet, err = net.ParseCIDR(listenAddr + "/24")
+	} else {
+		_, subnet, err = net.ParseCIDR(listenAddr + "/64")
+	}
```
stgraber, line 122: Can we get the CIDR from the address instead of assuming /24 for v4 and /64 for v6 as that can be very wrong?

I'd probably have gone with `net.InterfaceAddrs()` which gets you CIDR of all addresses on the system.
You'd then need to filter our anything that's not a global address from that by using ip.ParseIP and only going for GlobalUnicast.

masnax: Yeah that's a good point. I was less concerned about invalid subnets since I had added the  `--subnet` flag, but since that's gone it can be problematic.

I've added a `Subnet` field to the `NetworkInfo` struct now. So any selected address will have its associated subnet be used. If the user specifies their own address with `--address`, init will fail if that address doesn't fit in any of the found subnets.

PR #80, opened 2023-03-29
Ovn uplink network setup

### LINKED ISSUE #78: OVN uplink network setup
We should introduce the bare minimum set of questions to setup a functional OVN network, for this we should do:

- [ ] Update `microcloud` init to ask
  - Setup distributed networking? [default=YES] (no keeps things with lxdfan as today)
  - Please select an uplink interface for each server: (show list of possible interfaces on all servers, make sure exactly one is selected per server)
  - IPv4 gateway on the uplink network: (CIDR, empty to skip IPv4)
  - IPv4 address range for use by LXD (required if IPv4 enabled)
  - IPv6 gateway on the uplink network
  - IPv6 address range for use by LXD (required if IPv6 enabled)
- [ ] If OVN is being setup:
  - Define the uplink network as UPLINK in LXD (type physical, set parent for each server, set ipv4.gateway, ipv4.ovn.ranges, ipv6.gateway, ipv6.ovn.ranges on global entry)
  - Define a new network called `default` of type `ovn`
  - Make the default profile use the `default` network
- [ ] If OVN isn't being setup:
  - Create lxdfan0 as an Ubuntu Fan network (same as today)
  - Make the default profile use the `lxdfan0` network

There's a lot more we can do around this to get external addresses routed to it, BGP, ... but the above is the bare minimum and shouldn't be too hard to add.

The list of valid interfaces to use as uplink can be found using:
```go
package main

import (
	"fmt"
	"os"
	"strings"

	"github.com/lxc/lxd/client"
	"github.com/lxc/lxd/shared"
)

func main() {
	d, err := lxd.ConnectLXDUnix("/var/snap/lxd/common/lxd/unix.socket", nil)
	if err != nil {
		os.Exit(1)
	}

	networks, err := d.GetNetworks()
	if err != nil {
		os.Exit(1)
	}

	uplinks := []string{}
	for _, network := range networks {
		// Skip managed networks.
		if network.Managed {
			continue
		}

		// OpenVswitch only supports physical ethernet or VLAN interfaces, LXD also supports plugging in bridges.
		if !shared.StringInSlice(network.Type, []string{"physical", "bridge", "vlan"}) {
			continue
		}

		state, err := d.GetNetworkState(network.Name)
		if err != nil {
			continue
		}

		// OpenVswitch only works with full L2 devices.
		if state.Type != "broadcast" {
			continue
		}

		// Can't use interfaces that aren't up.
		if state.State != "up" {
			continue
		}

		// Make sure the interface isn't in use by ensuring there's no global addresses on it.
		addresses := []string{}
		for _, address := range state.Addresses {
			if address.Scope != "global" {
				continue
			}

			addresses = append(addresses, address.Address)
		}

		if len(addresses) > 0 {
			continue
		}

		uplinks = append(uplinks, network.Name)
	}

	fmt.Printf("Candidate: %s\n", strings.Join(uplinks, ", "))
}
```

The manual version of what's described above is:
```
lxc network create --type physical --target micro01 UPLINK parent=eth0.200
lxc network create --type physical --target micro02 UPLINK parent=eth0.200
lxc network create --type physical --target micro03 UPLINK parent=eth0.200
lxc network create --type physical UPLINK ipv4.gateway=172.30.200.1/24 ipv4.ovn.ranges=172.30.200.100-172.30.200.200 ipv6.gateway=2602:fc62:d:200::1/64 ipv6.ovn.ranges=2602:fc62:d:200::100-2602:fc62:d:200::200
lxc network create default network=UPLINK --type ovn
lxc profile device add default eth0 nic network=default name=eth0
```

masnax:
Closes #78

### DIFF
--- a/microcloud/cmd/microcloud/ask.go
+++ b/microcloud/cmd/microcloud/ask.go
@@ -8,6 +8,7 @@ import (
 
 	cephTypes "github.com/canonical/microceph/microceph/api/types"
 	"github.com/lxc/lxd/lxc/utils"
+	"github.com/lxc/lxd/shared/api"
 	lxdAPI "github.com/lxc/lxd/shared/api"
 	cli "github.com/lxc/lxd/shared/cmd"
 	"github.com/lxc/lxd/shared/logger"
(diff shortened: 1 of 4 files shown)


microcloud/service/lxd.go
```diff
@@ -478,9 +449,93 @@ func (s LXDService) GetUplinkInterfaces(bootstrap bool, peers map[string]mdns.Se
...
+				network.Config["ipv4.ovn.ranges"] = ipRange
+			} else {
+				network.Config["ipv6.gateway"] = gateway
+				network.Config["ipv6.ovn.ranges"] = ipRange
+			}
+		}
+
+		err = client.CreateNetwork(api.NetworksPost{
+			NetworkPut: network,
+			Name:       "UPLINK",
```
stgraber, line 516: I'd just add a line with `Comment: "Uplink for OVN networks"` just so we have a description for it.

microcloud/cmd/microcloud/ask.go
```diff
@@ -477,3 +478,152 @@ func askRemotePool(peerDisks map[string][]lxdAPI.ResourcesStorageDisk, autoSetup
...
+				}
+
+				if addr.To4() != nil && ip == "IPv6" {
+					return fmt.Errorf("Not a valid IPv6")
+				}
+
+				return nil
+			}
+
+			gateway, err := cli.AskString(fmt.Sprintf("Select the %s gateway on the uplink network [CIDR]: ", ip), "", validator)
```
stgraber, line 567: Should probably mention `(empty to skip)` or something along those lines.

PR #93, opened 2023-04-21
doc: add more information to the docs

ru-fu:
Also split up the explanation section into separate pages.

### DIFF
--- a/doc/explanation/initialisation.rst
+++ b/doc/explanation/initialisation.rst
@@ -1,0 +1,76 @@
+.. _explanation-initialisation:
+
+================================
+About the initialisation process
+================================
+
+See :ref:`howto-initialise` for instructions on how to set up MicroCloud.
+
+.. _automatic-server-detection:
+
+Automatic server detection
+--------------------------
+
+MicroCloud uses :abbr:`mDNS (multicast DNS)` to automatically detect other servers on the network.
+This method works in physical networks, but it is usually not supported in a cloud environment.
+
+By default, the scan is limited to the local subnet.
+
+MicroCloud will display all servers that it detects, and you can then select the ones you want to add to the MicroCloud cluster.
+
+.. _bootstrapping-process:
+
+Bootstrapping process
+---------------------
+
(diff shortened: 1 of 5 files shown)


doc/explanation/initialisation.rst
```diff
@@ -0,0 +1,76 @@
...
+   +--------+--------+-----------------------------+---------+---------+
+   | remote | ceph   | Distributed storage on Ceph | 7       | CREATED |
+   +--------+--------+-----------------------------+---------+---------+
+   :input: lxc network list
+   +----------+----------+---------+----------------+--------------------------+-------------+---------+---------+
+   |   NAME   |   TYPE   | MANAGED |      IPV4      |           IPV6           | DESCRIPTION | USED BY |  STATE  |
+   +----------+----------+---------+----------------+--------------------------+-------------+---------+---------+
+   | UPLINK   | physical | YES     |                |                          |             | 2       | CREATED |
+   +----------+----------+---------+----------------+--------------------------+-------------+---------+---------+
+   | default  | ovn      | YES     | 192.0.2.1/24   | 2001:db8:39dc:898c::1/64 |             | 5       | CREATED |
```
ru-fu, line 71: Do these IP addresses make sense here?
I replaced them with addresses from the documentation range, but that might be confusing since those are external addresses if I understand it correctly?

stgraber: Yeah, probably best to use internal addresses as OVN is likely to give rather than use the example subnets which are global.

```
+----------+----------+---------+----------------+---------------------------+-------------+---------+---------+
| default  | ovn      | YES     | 10.45.196.1/24 | fd42:67ac:35af:257a::1/64 |             | 6       | CREATED |
+----------+----------+---------+----------------+---------------------------+-------------+---------+---------+
```

stgraber: Maybe use `10.123.123.1/24` and `fd42:1234:1234:1234::1/64`?

ru-fu: Thanks! Updated.

PR #96, opened 2023-04-25
doc: language fixes for initialization prompts and help texts

ru-fu:
I haven't done any tests or Go validation or anything, but it should really only be changes to the strings. ;)

### DIFF
--- a/microcloud/cmd/microcloud/shutdown.go
+++ b/microcloud/cmd/microcloud/shutdown.go
@@ -14,7 +14,7 @@ type cmdShutdown struct {
 func (c *cmdShutdown) Command() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "shutdown",
-		Short: "Shutdown MicroCluster daemon",
+		Short: "Shut down the MicroCluster daemon",
 		RunE:  c.Run,
 	}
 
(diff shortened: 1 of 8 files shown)


microcloud/cmd/microcloud/shutdown.go
```diff
@@ -14,7 +14,7 @@ type cmdShutdown struct {
 func (c *cmdShutdown) Command() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "shutdown",
-		Short: "Shutdown MicroCluster daemon",
+		Short: "Shut down the MicroCluster daemon",
```
ru-fu, line 17: We're talking about MicroCluster instead of MicroCloud here - is that correct?
Also in the context of join tokens?

Would it be clearer to talk about "the MicroCloud cluster"?
Or if not, do we need to add something to the docs about what MicroCluster is?

stgraber: Nope, that's a typo :)

`MicroCluster` is the name of the boilerplate/shared logic we use for all the Micro projects, it should never make it to something user visible.

ru-fu: OK, thanks, I'll make a follow-up PR.

PR #105, opened 2023-05-08
doc: fix version error

ru-fu:
Read the docs uses an older Python version that cannot deal with urllib3 v2.0.


ru-fu:
It would be cleaner to require a newer Python version, but that requires adding a read-the-docs config file.
We have an issue for that for the starter pack, so we'll get it for free once that is implemented.

PR #124, opened 2023-06-19
ovn: Use SSL as required by current edge snap

stgraber:


tomponline (approved):
That turned out to be simple :) 
Is there any cert management to do?

stgraber:
@tomponline there's a matching change to lxd-pkg-snap which sets up /etc/ovn with the certs coming from microovn, basically using the same certs that it itself uses for microovn.ovn-nbctl

PR #125, opened 2023-06-19
Allow skipping systems for ceph

masnax:
Changes the requirement for ceph to at least 1 disk per machine up to 3 machines, rather than at least 1 disk for every machine.

Additionally, adds prompts to the interactive init to confirm if users wish to continue setting up ceph or ovn if some systems don't meet the requirements and must be skipped.

### DIFF
--- a/microcloud/cmd/microcloud/ask.go
+++ b/microcloud/cmd/microcloud/ask.go
@@ -197,10 +197,9 @@ func askDisks(sh *service.ServiceHandler, peers map[string]mdns.ServerInfo, boot
 
 	var cephDisks map[string][]cephTypes.DisksPost
 	if sh.Services[types.MicroCeph] != nil {
-		skipCeph := false
 		availableDisks := map[string][]lxdAPI.ResourcesStorageDisk{}
 		for peer, disks := range validDisks {
-			availableDisks[peer] = []lxdAPI.ResourcesStorageDisk{}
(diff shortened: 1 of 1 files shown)


microcloud/cmd/microcloud/ask.go
```diff
@@ -520,6 +523,17 @@ func askNetwork(sh *service.ServiceHandler, peers map[string]mdns.ServerInfo, lx
 		return nil, nil, nil
 	}
 
+	if len(peers) != len(networks) {
+		wantsSkip, err := cli.AskBool("Some systems are ineligible for distributed networking. Continue anyway? (yes/no) [default=yes]: ", "yes")
```
ru-fu, line 527: I now get this message when I have 4 VMs selected for MicroCloud, and when I confirm all 4 of them are displayed, so apparently they are all eligible?

ru-fu (commented):
In the table where I select the disks for distributed storage, I now get a "v" character underneath the "boxes" for selection - is that on purpose?

image

ru-fu:
> If for some reason that showed up for you without 16 items in the table, it might be some weird formatting edge case. If it happens again, could you copy the table for me please?

I only have 7 disks set up, and the table had only the 3 that were left for distributed storage.
It happened on all my attempts - I can copy it the next time I try my luck. ;)

PR #128, opened 2023-07-05
doc: update links to linuxcontainers

ru-fu:


doc/conf.py
```diff
@@ -38,7 +38,7 @@
 
 # Used for related links - no need to change
 if 'discourse' in html_context:
-    html_context['discourse_prefix'] = html_context['discourse'] + "/t/"
+    html_context['discourse_prefix'] = "https://discuss.linuxcontainers.org/t/"
```
tomponline, line 41: Should we link to the old forum here?

ru-fu: Yes, the `discourse_prefix` is used to generate links to specific Discourse topics, and those ones are still on the old forum.
I have a task to update the extension to support different Discourse instances, but for now, all links go to the old one. :)

PR #133, opened 2023-07-07
mdns: Replace CutSuffix with HasSuffix and TrimSuffix.

### LINKED ISSUE #132: `strings.CutSuffix` was added in go1.20 but go mod specifies go1.18
We are using `strings.CutSuffix` here

https://github.com/canonical/microcloud/blob/ea0b7a26090f7ae40ff3f27fb9bdea64bf8bb438/microcloud/mdns/lookup.go#L81-L84

This was added in go 1.20 (see https://pkg.go.dev/strings#CutSuffix). We should probably either set the go mod to use go 1.20 or change this to use a combination of `strings.HasSuffix` and `strings.TrimSuffix`.

markylaing:
CutSuffix was added in go1.20. The go module for microcloud specifies go1.18 so we should only use what is availabile in that go version.

Closes #132

### DIFF
--- a/microcloud/mdns/lookup.go
+++ b/microcloud/mdns/lookup.go
@@ -78,11 +78,11 @@ func LookupPeers(ctx context.Context, version string, localPeer string) (map[str
 			return nil, fmt.Errorf("Received empty record")
 		}
 
-		serviceStr, ok := strings.CutSuffix(entry.Name, ".local.")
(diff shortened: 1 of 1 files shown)


tomponline:
@markylaing sounds like we could do with some github workflows here to test the build against go 1.18.

Issue #107, opened 2023-05-11
Race getting OVN southbound connection during microcloud init


tomponline:
Sometimes when running `microcloud init` inside VMs I get this error:

```
Error: failed to notify peer 10.21.203.2:8443: Failed to get OVN client: Failed to get OVN southbound connection string: Failed to run: ovs-vsctl get open_vswitch . external_ids:ovn-remote: exit status 1 (ovs-vsctl: unix:/var/run/openvswitch/db.sock: database connection failed (No such file or directory))
```

But it doesn't happen every time, suggesting it is a race whilst trying to access OVS before its started.

Full steps:

```
+ lxc project create microcloud
Project microcloud created
+ lxc project switch microcloud
+ lxc profile device add default root disk pool=default path=/
Device root added to default
+ lxc profile device add default eth0 nic network=lxdbr0 name=eth0
Device eth0 added to default
+ lxc network create microbr0 ipv4.address=10.123.123.1/24 ipv4.dhcp=false ipv4.nat=true ipv6.address=fd42:1234:1234:1234::1/64 ipv6.nat=true
Network microbr0 created
+ lxc profile device add default eth1 nic network=microbr0 name=eth1
Device eth1 added to default
+ lxc storage volume create default micro01-disk1 size=30GiB --type=block
Storage volume micro01-disk1 created
+ lxc storage volume create default micro01-disk2 size=50GiB --type=block
Storage volume micro01-disk2 created
+ lxc storage volume create default micro02-disk1 size=30GiB --type=block
Storage volume micro02-disk1 created
+ lxc storage volume create default micro02-disk2 size=50GiB --type=block
Storage volume micro02-disk2 created
+ lxc storage volume create default micro03-disk1 size=30GiB --type=block
Storage volume micro03-disk1 created
+ lxc storage volume create default micro03-disk2 size=50GiB --type=block
Storage volume micro03-disk2 created
+ lxc init ubuntu:22.04 micro01 --vm -c limits.cpu=2 -c limits.memory=4GiB
Creating micro01
+ lxc config device add micro01 disk1 disk pool=default source=micro01-disk1
Device disk1 added to micro01
+ lxc config device add micro01 disk2 disk pool=default source=micro01-disk2
Device disk2 added to micro01
+ lxc start micro01
+ lxc init ubuntu:22.04 micro02 --vm -c limits.cpu=2 -c limits.memory=4GiB
Creating micro02
[...]

PR #136, opened 2023-07-12
Restart LXD on MicroCloud start to refresh symlinks

masnax:
Adds a hook that restarts LXD on uninitialized MicroCloud systems when MicroCloud starts.

I'm not sure if immediately calling `GetServer` like I'm doing here will cause issues on slower machines, it seems to work fine on my machine.

This still has the caveat of requiring a `snap restart microcloud` to force LXD to restart. I'm open to also putting this restart on run on an execution of `microcloud init` although I'm concerned that have some odd effects on slower machines again.

### DIFF
--- a/microcloud/service/lxd.go
+++ b/microcloud/service/lxd.go
@@ -9,11 +9,13 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"time"
 
 	lxd "github.com/canonical/lxd/client"
 	"github.com/canonical/lxd/lxd/util"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/microcluster/microcluster"
 
 	"github.com/canonical/microcloud/microcloud/api/types"
@@ -678,6 +680,82 @@ func (s *LXDService) SetConfig(target string, secret string, config map[string]s
 	}
 
 	return c.UpdateServer(newServer, "")
+}
+
+// Restart requests LXD to shutdown, then waits until it is ready.
+func (s *LXDService) Restart(timeoutSeconds int) error {
+	c, err := s.client("")
+	if err != nil {
+		return err
+	}
+
+	_, _, err = c.RawQuery("PUT", "/internal/shutdown", nil, "")
+	if err != nil {
+		return fmt.Errorf("Failed to send shutdown request to LXD: %w", err)
+	}
+
+	err = s.waitReady(c, timeoutSeconds)
+	if err != nil {
+		return err
+	}
+
+	// A sleep might be necessary here on slower machines?
+	_, _, err = c.GetServer()
+	if err != nil {
+		return fmt.Errorf("Failed to initialize LXD server: %w", err)
+	}
+
+	return nil
+}
+
+// waitReady repeatedly (500ms intervals) asks LXD if it is ready, up to the given timeout.
+func (s *LXDService) waitReady(c lxd.InstanceServer, timeoutSeconds int) error {
+	finger := make(chan error, 1)
+	var errLast error
+	go func() {
+		for i := 0; ; i++ {
+			// Start logging only after the 10'th attempt (about 5
+			// seconds). Then after the 30'th attempt (about 15
+			// seconds), log only only one attempt every 10
+			// attempts (about 5 seconds), to avoid being too
+			// verbose.
+			doLog := false
+			if i > 10 {
+				doLog = i < 30 || ((i % 10) == 0)
+			}
+
+			if doLog {
+				logger.Debugf("Checking if LXD daemon is ready (attempt %d)", i)
+			}
+
+			_, _, err := c.RawQuery("GET", "/internal/ready", nil, "")
+			if err != nil {
+				errLast = err
+				if doLog {
+					logger.Warnf("Failed to check if LXD daemon is ready (attempt %d): %v", i, err)
+				}
+
+				time.Sleep(500 * time.Millisecond)
+				continue
+			}
+
+			finger <- nil
+			return
+		}
+	}()
+
+	if timeoutSeconds > 0 {
+		select {
+		case <-finger:
+			break
+		case <-time.After(time.Second * time.Duration(timeoutSeconds)):
+			return fmt.Errorf("LXD is still not running after %ds timeout (%v)", timeoutSeconds, errLast)
+		}
+	} else {
+		<-finger
+	}
+
+	return nil
 }
 
 // defaultGatewaySubnetV4 returns subnet of default gateway interface.
--- a/microcloud/service/lxd.go
+++ b/microcloud/service/lxd.go
@@ -687,11 +687,31 @@ func (s *LXDService) SetConfig(target string, secret string, config map[string]s
 	return c.UpdateServer(newServer, "")
 }
 
+// isInitialized checks if LXD is initialized by fetching the storage pools.
+// If none exist, that means LXD has not yet been set up.
+func (s *LXDService) isInitialized(c lxd.InstanceServer) (bool, error) {
+	pools, err := c.GetStoragePoolNames()
+	if err != nil {
+		return false, err
+	}
+
+	return len(pools) != 0, nil
+}
+
 // Restart requests LXD to shutdown, then waits until it is ready.
 func (s *LXDService) Restart(timeoutSeconds int) error {
 	c, err := s.client("")
 	if err != nil {
 		return err
+	}
+
+	isInit, err := s.isInitialized(c)
+	if err != nil {
+		return fmt.Errorf("Failed to check LXD initialization: %w", err)
+	}
+
+	if isInit {
+		return fmt.Errorf("LXD has already been initialized")
 	}
 
 	_, _, err = c.RawQuery("PUT", "/internal/shutdown", nil, "")
--- a/microcloud/service/lxd.go
+++ b/microcloud/service/lxd.go
@@ -152,6 +152,11 @@ func (s LXDService) Bootstrap() error {
 
 // Join joins a cluster with the given token.
 func (s LXDService) Join(joinConfig JoinConfig) error {
+	err := s.Restart(30)
+	if err != nil {
+		return err
+	}
+
 	config, err := s.configFromToken(joinConfig.Token)
 	if err != nil {
 		return err
--- a/microcloud/service/service_handler.go
+++ b/microcloud/service/service_handler.go
@@ -8,8 +8,8 @@ import (
 	"path/filepath"
 	"sync"
 
-	lxd "github.com/canonical/lxd/client"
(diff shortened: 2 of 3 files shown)


tomponline:
Please can you update the PR description to explain what prompted this change, what does it fix? Thanks

masnax:
@tomponline

The main issue is that when MicroCloud directs LXD to set up an OVN network with MicroOVN, LXD uses some symlinks in its snap environment to execute the commands. These symlinks are generated by the LXD snap's daemon start hook, so if LXD happens to start before MicroOVN is installed, LXD won't have the right symlinks, and setting up the network will fail (Resulting in the error in the main post of #107).

To fix this, LXD needs to be restarted when MicroCloud runs, so it has the right symlinks. 

The problems here are:
* MicroCloud *actually* needs LXD to restart before a call to `microcloud init`, restarting on daemon start might not be enough.
* LXD restarting may not solve the problem in the `microcloud add` case because if the local system needs to restart its LXD, then a node in the cluster has to go down. Additionally, the node being added might need to be signalled to restart its LXD as well.

The other solution (as we briefly discussed on IRC) would be to use the proper paths dynamically in LXD. This would solve the problems above, but would add split the central location where we track the paths of services that LXD works with, and integrate it into the code.

markylaing:
> @tomponline
> 
> The main issue is that when MicroCloud directs LXD to set up an OVN network with MicroOVN, LXD uses some symlinks in its snap environment to execute the commands. These symlinks are generated by the LXD snap's daemon start hook, so if LXD happens to start before MicroOVN is installed, LXD won't have the right symlinks, and setting up the network will fail (Resulting in the error in the main post of #107).
> 
> To fix this, LXD needs to be restarted when MicroCloud runs, so it has the right symlinks.
> 
> The problems here are:
> 
> * MicroCloud _actually_ needs LXD to restart before a call to `microcloud init`, restarting on daemon start might not be enough if MicroOVN is installed after MicroCloud.
> * LXD restarting may not solve the problem in the `microcloud add` case because if the local system needs to restart its LXD, then a node in the cluster has to go down. Additionally, the node being added might need to be signalled to restart its LXD as well.
> 
> The other solution (as we briefly discussed on IRC) would be to use the proper paths dynamically in LXD. This would solve the problems above, but would add split the central location where we track the paths of services that LXD works with, and integrate it into the code.

I just want to double check I'm understanding this correctly:
* If `microOVN` is installed after LXD, the LXD snap (but presumably not LXD itself?) can detect that and will generate symlinks (to what?) that are required to configure OVN?
* We can't restart LXD during microcloud init. Why might this not be enough?
* We can't restart LXD during microcloud add. Why do we need to restart the local LXD when adding a new member? Can we not check if we have an OVN service and if so restart LXD in the new member?

Also where are these hooks defined, and what happens if microOVN is installed after the daemon is started?

markylaing:
Also a side note - can we add an internal endpoint to LXD to ask it to refresh it's symlinks without requiring a restart?

masnax:
> I just want to double check I'm understanding this correctly:
> 
>   * If `microOVN` is installed after LXD, the LXD snap (but presumably not LXD itself?) can detect that and will generate symlinks (to what?) that are required to configure OVN?

The LXD snap creates the symlinks only when the LXD daemon starts. If microovn is installed after LXD, these symlinks will only be set the next time the LXD snap starts.

>   * We can't restart LXD during microcloud init. Why might this not be enough?

This would solve the problem for the init case, but it would cause a consistent delay when a user initially runs `microcloud init`. Not sure how bad this would be on slower systems.

> 
>   * We can't restart LXD during microcloud add. Why do we need to restart the local LXD when adding a new member? Can we not check if we have an OVN service and if so restart LXD in the new member?

So when joining, we perform some actions both on the joining system that sends the join request and the clustered system that eventually handles it (dqlite leader, I believe). I believe currently the symlinks we care about are just for setting up local services for ovn, but technically both the cluster and the joiner might potentially want to use some of these paths. 

> 
> Also where are these hooks defined, and what happens if microOVN is installed after the daemon is started?

Where it pertains to microovn, these symlinks are mainly for some certs and state information passed to ovn commands. They're set up [here](https://github.com/canonical/lxd-pkg-snap/blob/817b6435e50f9c51ab0a82c3d454ff896bdae21b/snapcraft/commands/daemon.start#L324C38-L324C38) and [here](https://github.com/canonical/lxd-pkg-snap/blob/817b6435e50f9c51ab0a82c3d454ff896bdae21b/snapcraft/commands/daemon.start#L485C19-L485C19)

If microovn is installed after LXD is started, absolutely nothing happens. LXD would not have set up the symlinks in a way that would work with microovn.

> Also a side note - can we add an internal endpoint to LXD to ask it to refresh it's symlinks without requiring a restart?

That endpoint would have to have some way of executing either the `daemon.start` snap hook or some other snap action that can also be run when the daemon starts. I'm not familiar enough with how snaps work to know if this is possible.

markylaing:
> > * We can't restart LXD during microcloud init. Why might this not be enough?
> 
> This would solve the problem for the init case, but it would cause a consistent delay when a user initially runs `microcloud init`. Not sure how bad this would be on slower systems.

I don't think it's too much of a concern to have a delay during microcloud init. The init command is interactive and takes a while anyway. We can indicate to the user what we are doing.

> 
> > * We can't restart LXD during microcloud add. Why do we need to restart the local LXD when adding a new member? Can we not check if we have an OVN service and if so restart LXD in the new member?
> 
> So when joining, we perform some actions both on the joining system that sends the join request and the clustered system that eventually handles it (dqlite leader, I believe). I believe currently the symlinks we care about are just for setting up local services for ovn, but technically both the cluster and the joiner might potentially want to use some of these paths.

I'm still confused about this, why would the existing member use symlinks on another machine?

> > Also a side note - can we add an internal endpoint to LXD to ask it to refresh it's symlinks without requiring a restart?
> 
> That endpoint would have to have some way of executing either the `daemon.start` snap hook or some other snap action that can also be run when the daemon starts. I'm not familiar enough with how snaps work to know if this is possible.

Presumably the `daemon.start` snap hook is executed in the same environment as LXD itself, otherwise it would be a security concern. Is that the case @simondeziel @tomponline? 

If this is true then the symlinks created in the daemon.start hook could be created on the fly by LXD itself. We could add a `/internal/detect-integrations` (or similar) endpoint to LXD and call it from microcloud on all nodes during `init`, and on the joining node during `add`. Then there would be no reason to restart LXD.

masnax:
> I'm still confused about this, why would the existing member use symlinks on another machine?

Not from another machine. It's a hypothetical edge case: A node with a symlink to something that's utilized once the node is already clustered, and is handling a join request. This node would've joined the cluster with incorrectly set up symlinks (because LXD started before the relevant service), and never happened to restart. It wouldn't have errored out when it joined because the hypothetical command is run on existing cluster members, so the invalid state on the joiner is irrelevant. Once the node is clustered, its state would be silently invalid until this node becomes dqlite leader and is forwarded a request to add a new cluster member. 

> If this is true then the symlinks created in the daemon.start hook could be created on the fly by LXD itself. We could add a `/internal/detect-integrations` (or similar) endpoint to LXD and call it from microcloud on all nodes during `init`, and on the joining node during `add`. Then there would be no reason to restart LXD.

Right, personally I think this is the most foolproof solution, but involves migrating all of this state logic over to `go`, unless there's some way to leave it in the snap configuration and have the endpoint execute a snap action which runs exactly what's currently in `daemon.start`.

tomponline:
I think this gets to the heart of the issue. Should LXD (beyond its own packaging) have knowledge of and interact with (even just passively) MicroCloud, or should we treat it as just another external client.

I would strongly caution against using `/internal` from anything except lxd itself, as otherwise it fundamentally changes what the `/internal` endpoints are for.

I think we should discuss this in our next microcloud meeting so we can all get on the same page, and consider both microovn and microceph at the same time to see if we can come up with a general solution.

simondeziel:
Those snap to snap interactions should probably be best discussed with the snap people (https://chat.canonical.com/canonical/channels/starcraft). It sounds to me that we would need some more interfaces automatically connected to let them interact.

tomponline:
Lets chat about it in our next meeting as it maybe we can avoid needing to do it in the snap at all.

masnax:
As discussed in our meeting, I've updated this PR to restart LXD in more places:
1) When MicroCloud starts
2) When `microcloud init` is called
3) When a joining system receives its join token from the system that called `microcloud init`

This should suffice and avoid the nasty MicroOVN error until we can discuss a more permanent and resilient solution that doesn't rely on restarting LXD completely.

microcloud/service/service_handler.go
```diff
@@ -84,10 +84,9 @@ func (s *Handler) Start(state *state.State) error {
 		return nil
 	}
 
-	// Attempt to wake up LXD so it can generate certificates already.
-	d, err := lxd.ConnectLXDUnix("/var/snap/lxd/common/lxd/unix.socket", nil)
-	if err == nil {
-		_, _, _ = d.GetServer()
+	err := s.Services[types.LXD].(*LXDService).Restart(30)
+	if err != nil {
+		logger.Error("Failed to reload LXD", logger.Ctx{"error": err})
```
tomponline, line 89: Reload is wrong term here its restart

microcloud/service/lxd.go
```diff
@@ -687,13 +687,33 @@ func (s *LXDService) SetConfig(target string, secret string, config map[string]s
...
 		return err
 	}
 
+	isInit, err := s.isInitialized(c)
+	if err != nil {
+		return fmt.Errorf("Failed to check LXD initialization: %w", err)
+	}
+
+	if isInit {
+		return fmt.Errorf("LXD has already been set up")
```
tomponline, line 714: initialized

microcloud/service/lxd.go
```diff
@@ -680,6 +682,82 @@ func (s *LXDService) SetConfig(target string, secret string, config map[string]s
...
 }
 
+// Restart requests LXD to shutdown, then waits until it is ready.
+func (s *LXDService) Restart(timeoutSeconds int) error {
+	c, err := s.client("")
+	if err != nil {
+		return err
+	}
+
+	_, _, err = c.RawQuery("PUT", "/internal/shutdown", nil, "")
```
tomponline, line 717: Does this actually cause the snap part of lxd to restart? Can u point me to which bit is triggered in snap packaging?

masnax: Yeah, the `daemon.start` hook does actually run again when this happens.

If I'm understanding this correctly, the snap sets up a systemd service based on the the config [here](https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft.yaml#L71-L77). 

Seems that a restart is triggered `on-failure`. I assume this is triggered because LXD is getting restarted directly and not through systemd.

tomponline: lxd stops itself and systemd starts it again

masnax: Ah so it's also enabled by default

tomponline (approved):

PR #137, opened 2023-07-12
Pass table data to template

masnax:
Fixes an issue where artifacts would be shown in the table, and in rare occasions the whole table would be drawn incorrectly.

The issue was that the `go-survey` package we use for the selectable table does not re-compile the table template, instead pulling them from a map keyed by the string value of the template. 

As @markylaing found, there is an `interface{}` type on `MultiSelect` that we can abuse to pass in our own data into the template. What this means is we no longer need a method to keep track of the data being passed into the template and template functions, so we can assign the template as soon as possible.

In order to still preserve the ability to pass in pre-checked values into the table, I've added a field on to the new `templateData` struct, which alongside the `apply_defaults` template function, allows setting pre-checked rows like before.

### DIFF
--- a/microcloud/cmd/microcloud/selection_table.go
+++ b/microcloud/cmd/microcloud/selection_table.go
@@ -107,14 +107,14 @@ const multiSelectQuestionTemplate = `
 {{- define "option"}}
 	  {{- $line := "%s" }}
 	  {{- if (eq .CurrentOpt.Value (index .PageEntries 0).Value) }}
-	         {{- print (scroll_hint_top $line .FilterMessage .PageEntries) "\n" }}
+	         {{- print (scroll_hint_top $line .FilterMessage .PageEntries .Options) "\n" }}
 	  {{- end}}
     {{- if eq .SelectedIndex .CurrentIndex }}{{color .Config.Icons.SelectFocus.Format }}{{ .Config.Icons.SelectFocus.Text }}{{color "reset"}}{{else}} {{end}}
     {{- if index .Checked .CurrentOpt.Index }}{{color .Config.Icons.MarkedOption.Format }} {{ .Config.Icons.MarkedOption.Text }} {{else}}{{color .Config.Icons.UnmarkedOption.Format }} {{ .Config.Icons.UnmarkedOption.Text }} {{end}}
     {{- color "reset"}}
     {{- " "}}{{- .CurrentOpt.Value}}
 	  {{- if eq (add 1 .CurrentIndex) (len .PageEntries) }}
-	        {{- print "\n" (scroll_hint_bot $line .FilterMessage .PageEntries) }}
+	        {{- print "\n" (scroll_hint_bot $line .FilterMessage .PageEntries .Options) }}
 	  {{- end }}
 {{end}}
 {{- if gt (len .PageEntries) 0 }}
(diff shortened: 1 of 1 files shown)


tomponline:
I've used the request review feature so its pinned to my todo list and doesnt get lost.

markylaing (dismissed):
I've noticed that the current `multiSelectQuestionTemplate` contains `%s` formatting directives. Then when we update the table, the `survey.MultiSelectQuestionTemplate` is set via `fmt.Sprintf(multiSelectQuestionTemplate)`. This means that the multiSelectQuestionTemplate isn't actually a valid template. 

Is it possible to pass extra data where the template is rendered? Then rather than using `%s` you can use `.border` and `.header` and we won't need to keep resetting the template.

masnax:
> I've noticed that the current `multiSelectQuestionTemplate` contains `%s` formatting directives. Then when we update the table, the `survey.MultiSelectQuestionTemplate` is set via `fmt.Sprintf(multiSelectQuestionTemplate)`. This means that the multiSelectQuestionTemplate isn't actually a valid template.

Yes, I suppose this would technically make more sense as a function than a global variable. 

> Is it possible to pass extra data where the template is rendered? Then rather than using `%s` you can use `.border` and `.header` and we won't need to keep resetting the template.

Unfortunately the template data is set internally, (here)[https://github.com/go-survey/survey/blob/fa37277e6394c29db7bcc94062cb30cd7785a126/core/template.go#L43] and its return value is used internally where the table is rendered.

markylaing:
> > I've noticed that the current `multiSelectQuestionTemplate` contains `%s` formatting directives. Then when we update the table, the `survey.MultiSelectQuestionTemplate` is set via `fmt.Sprintf(multiSelectQuestionTemplate)`. This means that the multiSelectQuestionTemplate isn't actually a valid template.
> 
> Yes, I suppose this would technically make more sense as a function than a global variable.
> 
> > Is it possible to pass extra data where the template is rendered? Then rather than using `%s` you can use `.border` and `.header` and we won't need to keep resetting the template.
> 
> Unfortunately the template data is set internally, [here](https://github.com/go-survey/survey/blob/fa37277e6394c29db7bcc94062cb30cd7785a126/core/template.go#L43) and its return value is used internally where the table is rendered.

I find it a little bizarre that it is allowed to set the template but there is no easy/obvious way to set additional template data. I have managed to pass some extra data in via the default selection field though https://github.com/canonical/microcloud/compare/main...markylaing:microcloud:table-template

masnax:
> > I find it a little bizarre that it is allowed to set the template but there is no easy/obvious way to set additional template data. I have managed to pass some extra data in via the default selection field though [main...markylaing:microcloud:table-template](https://github.com/canonical/microcloud/compare/main...markylaing:microcloud:table-template)
> 
> Of course this won't work if we actually want to set any defaults. Maybe we can submit an issue upstream for this.

Actually, by making the map a `map[string]any`, we can maintain the current functionality of the `Default` field by just adding a new template function.

I ended up using this approach, but with a struct to clearly represent the 3 fields that are passed into the template. So this fully eliminates any need to reassign the template, and we don't need a global table anymore! Yay!

I'll open an issue asking for a cleaner way to pass in data to the template though.

markylaing:
I think perhaps the second and third commits could be combined into one? Only because the border and header values are passed into the template one way, then it is immediately changed in the following commit. 

Apart from that, I think it's good to go.

masnax:
@markylaing As you said you were happy with this, I'll dismiss your earlier review and merge.

masnax:
Ah, still can't merge without a review :)

markylaing (approved):
Looks good. Can you post a link to the issue so we can follow it? Thanks.

masnax:
https://github.com/go-survey/survey/issues/491

PR #138, opened 2023-07-13
Reorganize init configuration

masnax:
In an effort to make it easier to debug the built-up configuration of `microcloud init` and `microcloud add`, and also easier to validate, this PR reorganizes the init configuration around a map of structs which will hold all the configuration obtained by the init prompts.

The main change here is the introduction of the `InitSystem` type which maintains a record of all configuration that is built up over the user prompts. The type holds lists of API structs to be included in the join config or run after cluster setup. The API structs are mainly boilerplate, so some template functions have been introduced to reduce duplication around that as well.

This will also help with the upcoming preseed PR which will just need to validate the `yaml` payload into the same `map[string]InitSystem`.


masnax:
In order to fix a disk adding bug (which is also fixed in this refactor), I've created a new PR #139 and rebased this branch off of that one, so that one should now be merged first.

markylaing (changes requested):
Overall this refactor looks much clearer and simpler. That being said, as you are already refactoring, I think now is a good time to add some tests to make sure the changes you have made are working as intended.

tomponline:
@markylaing yeah @masnax discussed a similar concern in our 1:1 this week.

tomponline:
It would be useful to get a test script, similar to https://github.com/canonical/lxd-ci/pull/3 added for microcloud setup

microcloud/service/lxd.go
```diff
@@ -383,6 +383,10 @@ func (s LXDService) GetUplinkInterfaces(bootstrap bool, peers map[string]mdns.Se
 	}
 
 	for _, info := range peers {
+		if info.Name == s.Name() {
```
tomponline, line 298: This should have a comment explaining why as its not obvious.

microcloud/service/lxd_config.go
```diff
@@ -0,0 +1,180 @@
...
+func (s LXDService) DefaultPendingFanNetwork() api.NetworksPost {
+	return api.NetworksPost{Name: "lxdfan0", Type: "bridge"}
+}
+
+// DefaultFanNetwork returns the default Ubuntu Fan network configuration when
+// creating the finalized network.
+func (s LXDService) DefaultFanNetwork() (api.NetworksPost, error) {
+	underlay, _, err := defaultGatewaySubnetV4()
+	if err != nil {
+		return api.NetworksPost{}, fmt.Errorf("Couldn't determine Fan overlay subnet: %w", err)
```
tomponline, line 22: I tend to avoid using contractions in user messages, i.e "Couldn't" I would tend to use use "Failed to" or "Could not".

@ru-fu is this something you avoid in the docs too out of interest?

microcloud/service/lxd_config.go
```diff
@@ -0,0 +1,180 @@
...
+// DefaultCephStoragePool returns the default remote storage configuration when
+// creating the finalized pool.
+func (s LXDService) DefaultCephStoragePool() api.StoragePoolsPost {
+	return api.StoragePoolsPost{
+		Name:   "remote",
+		Driver: "ceph",
+		StoragePoolPut: api.StoragePoolPut{
+			Config: map[string]string{
+				"ceph.rbd.du":       "false",
+				"ceph.rbd.features": "layering,striping,exclusive-lock,object-map,fast-diff,deep-flatten",
```
tomponline, line 164: @masnax I don't know if you know why we use these here, but it would be good to get some advice from the ceph team as to what they do and then document here as a comment why we use them.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -61,14 +81,21 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
...
-		// 	return err
-		// }
+	systems[name] = InitSystem{
+		ServerInfo: mdns.ServerInfo{
+			Name:    name,
+			Address: addr,
+		},
 	}
 
+	//if !c.flagAutoSetup {
```
tomponline, line 91: We shouldn't leave this in.

microcloud/cmd/microcloud/ask.go
```diff
@@ -483,21 +483,44 @@ func askRemotePool(systems map[string]InitSystem, autoSetup bool, wipeAllDisks b
 	return nil
 }
 
-func askNetwork(sh *service.Handler, peers map[string]mdns.ServerInfo, lxdConfig map[string][]api.ClusterMemberConfigKey, bootstrap bool, autoSetup bool) (map[string]string, map[string]string, error) {
```
tomponline, line 484: This function signature is much nicer :)

tomponline (changes requested):
A few small changes needed, but overall this looks much clearer and more maintainable, thanks!

One thing I did notice throughout, and something I've wondered about before, is where does microcloud log to? I didn't see much logging (if any at all), and I think it would be useful to log more so we can see where things are failing, what do you think?

masnax:
Already working on tests :)

masnax:
> A few small changes needed, but overall this looks much clearer and more maintainable, thanks!
> 
> One thing I did notice throughout, and something I've wondered about before, is where does microcloud log to? I didn't see much logging (if any at all), and I think it would be useful to log more so we can see where things are failing, what do you think?

Just to syslog. `microcluster` automatically sets up a syslog logger when the daemon starts. It's used mainly around mdns at the moment, because that's the part that's most "magic". I'm not logging all that much in the CLI because its outputs are sort of are like logs printed directly to stdout anyway, and outside of that `microcloud` just calls the APIs to its services which have their own logs. 

There is a bit of a dearth of logs though, so I'll try to think about how I can incorporate them some more.

microcloud/service/lxd_config.go
ru-fu: Most style guides advise to avoid contractions, but the Ubuntu style guide actually encourages them: https://docs.ubuntu.com/styleguide/en#contractions
I use them occasionally, but I'm way too used to NOT using them, so it's rather rare. ;)

microcloud/service/lxd_config.go
masnax: These are defined rather well in the ceph docs. I've linked them and provided a brief summary in the comments.

microcloud/cmd/microcloud/ask.go
masnax: I very much agree :)

masnax:
@tomponline If you're OK with it, I think we can go ahead and merge this PR now, since the system tests have been running against the changes introduced here on my system.

tomponline:
> @tomponline If you're OK with it, I think we can go ahead and merge this PR now, since the system tests have been running against the changes introduced here on my system.

I cannot merge it with the tests failing on compile errors though:

```
golangci-lint run --timeout 5m
Error: cmd/microcloud/ask.go:29:22: undefined: cli.AskBool (typecheck)
				retry, err = cli.AskBool(fmt.Sprintf("%s (yes/no) [default=yes]: ", question), "yes")
				                 ^
Error: cmd/microcloud/ask.go:56:24: undefined: cli.AskBool (typecheck)
			confirm, err := cli.AskBool(fmt.Sprintf("%s not found. Continue anyway? (yes/no) [default=yes]: ", serviceStr), "yes")
			                    ^
Error: cmd/microcloud/ask.go:131:22: undefined: cli.AskBool (typecheck)
		filter, err := cli.AskBool(fmt.Sprintf("Limit search for other MicroCloud servers to %s? (yes/no) [default=yes]: ", subnet.String()), "yes")
		                   ^
Error: cmd/microcloud/ask.go:647:24: undefined: cli.AskString (typecheck)
			gateway, err := cli.AskString(msg, "", validator)
			                    ^
Error: cmd/microcloud/ask.go:671:29: undefined: cli.AskString (typecheck)
					rangeStart, err := cli.AskString(fmt.Sprintf("Specify the first %s address in the range to use with LXD: ", ip), "", validator)
					                       ^
Error: cmd/microcloud/ask.go:676:27: undefined: cli.AskString (typecheck)
					rangeEnd, err := cli.AskString(fmt.Sprintf("Specify the last %s address in the range to use with LXD: ", ip), "", validator)
					                     ^
make: *** [Makefile:27: check-static] Error 1
```

https://github.com/canonical/microcloud/actions/runs/6164591577/job/16730614381?pr=138

tomponline (changes requested):
the code doesn't current compile

tomponline:
Once its green, can you re-request a review by click the button next to my name as a reviewer so it appears in my todo list here https://github.com/pulls/review-requested

Thanks!

masnax:
Ah, looks like the LXD cli change is the culprit because the workflow runs `make update-gomod` before building. 

I'll need to send out some PRs for LXD Cloud, MicroOVN, and MicroCeph too.

microcloud/service/lxd.go
```diff
@@ -400,7 +400,7 @@ func (s *LXDService) Restart(timeoutSeconds int) error {
 	}
 
 	_, _, err = c.RawQuery("PUT", "/internal/shutdown", nil, "")
-	if err != nil {
+	if err != nil && err.Error() != "Shutdown already in progress" {
```
tomponline, line 417: What http code do you get with this error? We should perhaps change this to something you can detect programmatically that wont break if the message is changed. (This is the sort of thing juju used to do and it didn't go well).

microcloud/service/lxd.go
```diff
@@ -148,7 +148,21 @@ func (s LXDService) Bootstrap() error {
 		return fmt.Errorf("Failed to initialize cluster: %w", err)
 	}
 
-	return nil
+	for {
+		select {
+		case <-time.After(30 * time.Second):
+			return fmt.Errorf("Timed out waiting for LXD cluster to initialize")
+		default:
+			names, err := client.GetClusterMemberNames()
```
tomponline, line 156: Is this a busy wait that will dos the server until its ready? Shall we put a short skeeo between esxh iteration?

tomponline: Actually I don't see how this loop can work as the default case will always immediately be chosen and then block on GetClusterMemberNames. I don't see how the timeout can work. 

Did I miss something?

tomponline: Ideally we need GetClusterMemberNames to take a context, is this in the lxd client ?

tomponline (changes requested):
Something doesn't look right with those cluster member timeouts to me.

microcloud/service/lxd.go
masnax: This isn't trying to block fetching cluster members. It's repeatedly fetching the cluster members over the course of 30 seconds until the list is no longer empty. 

The reason for this is that there seems to be an bug (which I will file an issue for once I've investigated it a bit more) where LXD in particular will occasionally return from bootstrapping while the cluster is still non-functional.

masnax: I ended up adding the loop to all the services just to keep things consistent

masnax: It could use short sleep, though.

masnax: I don't think adding a context timeout to `ClusterMemberNames` all that necessary in this case because the client should already timeout before 30s is over.

masnax: If you've set up MicroCloud and received the error `failed to join the cluster with the given token` or something like that, then it's because of this error. When issuing a token after bootstrapping, the list of cluster members is still empty so the token gets generated without any addresses.

masnax-test:
Testing automated detection of PR: "138--2023-09-14T23:01:36Z--e6d9074e61766b1e501b013cf480fcf84770d2fd"

masnax-test:
Testing automated detection of PR: "138--2023-09-15T00:33:04Z--e6d9074e61766b1e501b013cf480fcf84770d2fd"

microcloud/service/lxd.go
masnax: It's just the default `500 Internal Server Error`. I can make it `429 Too Many Requests` and check for that instead.

masnax:
> Testing automated detection of PR: "138--2023-09-15T00:33:04Z--e6d9074e61766b1e501b013cf480fcf84770d2fd"

My strategy with this is to have a `PS5` machine poll for PRs every 5? minutes. If we encounter a new combination of `<pr num>--<timestamp>--<last commit hash>` that we haven't ever seen before, then that commit becomes queued on the machine. In order to qualify for the queue, a PR needs to pass all github workflows, be mergeable, and not be a draft. That's why this PR is the only one that got the comment.

The machine then comments on the PR that it has seen it, and waits to checkout the PR, build the snap, and run the test suite. Then finally make another comment on the test result.

This also has the added benefit of not having to tear down the VMs very often, since it takes about an hour to get them running. Instead, they can just hang out on the PS5 machine and we can just manually clear the snap states between tests.

tomponline:
> > Testing automated detection of PR: "138--2023-09-15T00:33:04Z--e6d9074e61766b1e501b013cf480fcf84770d2fd"
> 
> My strategy with this is to have a `PS5` machine poll for PRs every 5? minutes. If we encounter a new combination of `<pr num>--<timestamp>--<last commit hash>` that we haven't ever seen before, then that commit becomes queued on the machine. In order to qualify for the queue, a PR needs to pass all github workflows, be mergeable, and not be a draft. That's why this PR is the only one that got the comment.
> 
> The machine then comments on the PR that it has seen it, and waits to checkout the PR, build the snap, and run the test suite. Then finally make another comment on the test result.
> 
> This also has the added benefit of not having to tear down the VMs very often, since it takes about an hour to get them running. Instead, they can just hang out on the PS5 machine and we can just clear the snap states between tests.

Sounds like a good workaround for now whilst we wait for a permanent solution. Thanks.

microcloud/service/lxd.go
tomponline: Ah yes I see, makes sense now. I thought you were trying to timeout the call to `ClusterMemberNames` - this is why I shouldnt do reviews so late :)

Anyway, you're right we need a sleep in there otherwise we are going to be sending a DoS to the target server.

But still something isn't quite right with this loop. The time.After call is repeated on each loop iteration, meaning a new time.After channel and go routine is created for each iteration. This is a problem for 2 reason; 1 its resetting the timer on each check, and 2 its creating a new timer every time.

From the Go manual:

> The underlying Timer is not recovered by the garbage collector until the timer fires. If efficiency is a concern, use NewTimer instead and call Timer.Stop if the timer is no longer needed. 

https://pkg.go.dev/time#After

So I think it would be more correct and efficient to do this:

```go
timeout := time.NewTimer(30 * time.Second)
defer timeout.Stop()

for {
		select {
		case <-timeout.C:
			return fmt.Errorf("Timed out waiting for LXD cluster to initialize")
		default:
			names, err := client.GetClusterMemberNames()
           ....
          time.Sleep(time.Second)
}
```

An alternative is to use a context here:

```go
ctx, cancel := context.WithTimeout(context.TODO, 30*time.Second)
defer cancel()

for {
		select {
		case <-ctx.Done():
			return fmt.Errorf("Timed out waiting for LXD cluster to initialize: %w", ctx.Error())
		default:
			names, err := client.GetClusterMemberNames()
           ....
         time.Sleep(time.Second)
}
```

What do you think?

microcloud/service/lxd.go
tomponline: Yeah thats a good idea, thanks!

tomponline (changes requested):
looks like we just need to comments in https://github.com/canonical/microcloud/pull/138#discussion_r1326907887 addressed and we can merge this.

tomponline (approved):

microcloud/service/microovn.go
```diff
@@ -54,7 +55,30 @@ func (s OVNService) Client() (*client.Client, error) {
...
+			names, err := s.ClusterMembers()
+			if err != nil {
+				return err
+			}
+
+			if len(names) > 0 {
+				return nil
+			}
+
+			time.Sleep(300 * time.Millisecond)
```
tomponline, line 79: Why 300ms BTW?

tomponline:
Static analysis failures

microcloud/service/microovn.go
masnax: Not too fast, not too slow, do you prefer something else usually?

microcloud/service/microovn.go
```diff
@@ -54,7 +55,30 @@ func (s OVNService) Client() (*client.Client, error) {
 
 // Bootstrap bootstraps the MicroOVN daemon on the default port.
 func (s OVNService) Bootstrap() error {
-	return s.m.NewCluster(s.name, util.CanonicalNetworkAddress(s.address, s.port), 2*time.Minute)
+	err := s.m.NewCluster(s.name, util.CanonicalNetworkAddress(s.address, s.port), 2*time.Minute)
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 1 * time.Minute)
```
tomponline, line 63: Is there a global CTX that can be used in case init process is canceled BTW?

masnax: Ah good point, these aren't in the `main` package so they shouldn't use `Background()`

microcloud/service/microovn.go
tomponline: It depends for me. If I'm waiting for up to minute or more I tend to do 1s wait. If I'm waiting say up to 5s I would do between 100ms and 500ms.

masnax: Fair enough, bumped it to a second.

masnax:
@tomponline Looks to be passing now, so you can merge when ready.

tomponline (approved):
Thanks!

masnax:
And just like that, MicroCloud `edge` got way less buggy!

tomponline:
Yay! Thanks

I turned off requiring signed commits for now btw

tomponline:
As that was preventing merging

PR #139, opened 2023-07-13
microcloud/cmd/microcloud/main/init: Apply remote pool to all peers

masnax:
Fixes a bug where only peers that added disks to `microceph` would create a pending remote storage pool. Now all peers will create the pending remote storage pool if any disks across all peers are added to `microceph`.

It's not too pretty but this will be refactored out by #138 soon anyway.


markylaing (approved):
LGTM. As you say this will be refactored out in the next PR.

PR #156, opened 2023-09-11
Sequentially join the cluster

masnax:
Relies on #138 

his modifies the cluster join logic of MicroCloud to issue a token and instruct a system to join the cluster sequentially, one peer at a time.

This slightly slows down the join process but makes it more resilient as one node will not be overwhelmed with too many token requests, and then each join token won't have only the issuer's address as the target of the join request.

When there are 7 or more nodes left, MicroCloud will fall back to the old concurrent style of joining which issues all tokens for all systems at once, then concurrently instructs them all to join. After 7 peers, we should begin to see spare roles so there is less of a chance of role turnover blocking the cluster completely.

### DIFF
--- a/microcloud/cmd/microcloud/main_init.go
+++ b/microcloud/cmd/microcloud/main_init.go
@@ -498,7 +498,7 @@ func setupCluster(s *service.Handler, systems map[string]InitSystem) error {
 
 	config := map[string]string{"network.ovn.northbound_connection": ovnConfig}
 	lxd := s.Services[types.LXD].(*service.LXDService)
-	lxdClient, err := lxd.Client("")
+	lxdClient, err := lxd.Client(context.Background(), "")
 	if err != nil {
 		return err
 	}
@@ -528,7 +528,7 @@ func setupCluster(s *service.Handler, systems map[string]InitSystem) error {
 
 	// Create preliminary networks & storage pools on each target.
 	for name, system := range systems {
-		lxdClient, err := lxd.Client(system.ServerInfo.AuthSecret)
+		lxdClient, err := lxd.Client(context.Background(), system.ServerInfo.AuthSecret)
 		if err != nil {
 			return err
 		}
@@ -553,7 +553,7 @@ func setupCluster(s *service.Handler, systems map[string]InitSystem) error {
 	system, bootstrap := systems[s.Name]
 	if bootstrap {
 		lxd := s.Services[types.LXD].(*service.LXDService)
-		lxdClient, err := lxd.Client(system.ServerInfo.AuthSecret)
+		lxdClient, err := lxd.Client(context.Background(), system.ServerInfo.AuthSecret)
 		if err != nil {
 			return err
 		}
@@ -600,7 +600,7 @@ func setupCluster(s *service.Handler, systems map[string]InitSystem) error {
 
 	// With storage pools set up, add some volumes for images & backups.
 	for name, system := range systems {
-		lxdClient, err := lxd.Client(system.ServerInfo.AuthSecret)
+		lxdClient, err := lxd.Client(context.Background(), system.ServerInfo.AuthSecret)
 		if err != nil {
 			return err
 		}
--- a/microcloud/cmd/microcloud/main_init.go
+++ b/microcloud/cmd/microcloud/main_init.go
@@ -356,8 +356,6 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
 		time.Sleep(3 * time.Second)
 	}
 
-	fmt.Println("Cluster initialization is complete")
-
 	return nil
 }
 
@@ -403,6 +401,8 @@ func setupCluster(s *service.Handler, systems map[string]InitSystem) error {
 	if err != nil {
 		return err
 	}
+
+	fmt.Println("Configuring cluster-wide devices...")
 
 	var ovnConfig string
 	if s.Services[types.MicroOVN] != nil {
--- a/microcloud/cmd/microcloud/main_init.go
+++ b/microcloud/cmd/microcloud/main_init.go
@@ -262,6 +262,61 @@ func lookupPeers(s *service.Handler, autoSetup bool, subnet *net.IPNet, systems 
 	return nil
 }
 
+// waitForJoin issues a token and instructs a system to request a join,
+// and then waits for the request to either complete or time out.
+// If the request was successful, it additionally waits until the cluster appears in the database.
+func waitForJoin(sh *service.Handler, clusterSize int, secret string, peer string, cfg types.ServicesPut) error {
+	mut := sync.Mutex{}
+	err := sh.RunConcurrent(false, false, func(s service.Service) error {
+		token, err := s.IssueToken(context.Background(), peer)
+		if err != nil {
+			return fmt.Errorf("Failed to issue %s token for peer %q: %w", s.Type(), peer, err)
+		}
+
+		mut.Lock()
+		cfg.Tokens = append(cfg.Tokens, types.ServiceToken{Service: s.Type(), JoinToken: token})
+		mut.Unlock()
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	cloud := sh.Services[types.MicroCloud].(*service.CloudService)
+	err = cloud.RequestJoin(context.Background(), secret, peer, cfg)
+	if err != nil {
+		return fmt.Errorf("System %q failed to join the cluster: %w", peer, err)
+	}
+
+	clustered := make(map[types.ServiceType]bool, len(sh.Services))
+	for service := range sh.Services {
+		clustered[service] = false
+	}
+
+	now := time.Now()
+	for len(clustered) != 0 {
+		if time.Since(now) >= time.Second*30 {
+			return fmt.Errorf("Timed out waiting for cluster member %q to appear", peer)
+		}
+
+		for service := range clustered {
+			systems, err := sh.Services[service].ClusterMembers(context.Background())
+			if err != nil {
+				return err
+			}
+
+			if len(systems) == clusterSize+1 {
+				delete(clustered, service)
+			}
+		}
+	}
+
+	fmt.Printf(" Peer %q has joined the cluster\n", peer)
+
+	return nil
+}
+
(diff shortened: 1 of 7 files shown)


tomponline:
> 1. Then we concurrently tell all of those systems to request a join at once. All of these requests go to that one address, which causes a bunch of errors (timeouts, dqlite errors, etc) when there are either too many machines requesting to join, or the system is too slow to handle that many requests.

Are these errors in microcloud/microcluster or LXD?

Did you make any attempt to identify the specific issues and fix those? Or lock the contended paths internally before this workaround was added?

tomponline:
> 2\. Dqlite will very occasionally lock up on LXD on a joining system (reporting dqlite leader not found). The only fix I've found here is to restart that LXD.

Was this reported to to dqlite team? It sounds like something we should investigate and fix.

tomponline:
> I picked >7 half-arbitrarily to use the more speedy configuration, because at that point each token will have 7 addresses, avoiding the issue with overwhelming a system. At the same time, my assumption is that with more systems available in the cluster, it should be less likely to run into whatever is causing issue 2.

Do the joining clients pick an endpoint from the candidate list at random then?

I personally think we should leave this sort of workaround out of the code, especially as its based on assumptions.
Also, because of the way all reads and writes in dqlite go to the leader, they are all going to go back to a single member eventually anyway, even if they proxy through a random host first.

tomponline:
> This new logic mainly addresses point 1 by using up each join token to add a system, and then waiting until it's successfully added before issuing another token for another system. This is definitely slower, but avoids overwhelming the systems.

This part sounds like an acceptable workaround temporarily, as long as the issues themselves are logged and investigated, because ultimately it sounds like you've discovered some issue(s) in MicroCloud/LXD/dqlite that need fixing.

masnax:
> Are these errors in microcloud/microcluster or LXD?
> 
> Did you make any attempt to identify the specific issues and fix those? Or lock the contended paths internally before this workaround was added?

I would argue this was mainly a misuse of the clustering API of LXD and MicroCluster in the first place. Especially on a low-powered machine, sending 10-50 concurrent requests to accept 4 dqlite cluster members (LXD, MicroCloud, MicroOVN, MicroCeph) is sure to cause a bunch of problems.

> Was this reported to to dqlite team? It sounds like something we should investigate and fix.

This is the same long-standing dqlite problem that we've been having for years, and still affects the LXD CI occasionally. FWIW I've only noticed it happening with LXD's implementation rather than the one in MicroCluster. That said `go-dqlite` (which is used in MicroCluster) seems to be much more resource-intensive than LXD's implementation so I'm not sure if it would be feasible to transition LXD to that right now. I'll talk to the dqlite team again about looking into these.

> I personally think we should leave this sort of workaround out of the code, especially as its based on assumptions.

We can do this, though sequentially clustering 10+ machines would start taking minutes. Each sequential join takes about 6 seconds in the best-case scenario on my system.

> Do the joining clients pick an endpoint from the candidate list at random then?

Good point, I think they attempt to connect to addresses in alphabetical order by hostname. 

> Also, because of the way all reads and writes in dqlite go to the leader, they are all going to go back to a single member eventually anyway, even if they proxy through a random host first.
My hope was that this roundabout process would actually give the leader some time to process the join requests.

masnax:
I'll split the concurrent part off into a branch PR because I do want to investigate this further when we actually have space to test 10+ cluster members.

masnax:
> Are these errors in microcloud/microcluster or LXD?
> 
> Did you make any attempt to identify the specific issues and fix those? 

For the most part, the issues are just timeouts in handling requests because the cluster is busy. That part affects both LXD and all microcluster projects.

> Or lock the contended paths internally before this workaround was added?

Not sure what you mean by this?

tomponline:
> I'll split the concurrent part off into a draft PR because I do want to investigate this further when we actually have space to test 10+ cluster members.

Good idea. Thanks

tomponline:
> This is the same long-standing dqlite problem that we've been having for years, and still affects the LXD CI occasionally. FWIW I've only noticed it happening with LXD's implementation rather than the one in MicroCluster. That said `go-dqlite` (which is used in MicroCluster) seems to be much more resource-intensive than LXD's implementation so I'm not sure if it would be feasible to transition LXD to that right now. I'll talk to the dqlite team again about looking into these.

Thanks. I'd like to understand more about your observations of go-dqlite's App implementation (I assume you are referring to that) compared to LXD's direct integration WRT to resource usage.

tomponline:
> Good point, I think they attempt to connect to addresses in alphabetical order by hostname.

Can you confirm that please?

tomponline:
> > I personally think we should leave this sort of workaround out of the code, especially as its based on assumptions.
> 
> We can do this, though sequentially clustering 10+ machines would start taking minutes. Each sequential join takes about 6 seconds in the best-case scenario on my system.
> 
> > Do the joining clients pick an endpoint from the candidate list at random then?
> 
> Good point, I think they attempt to connect to addresses in alphabetical order by hostname.
> 
> > Also, because of the way all reads and writes in dqlite go to the leader, they are all going to go back to a single member eventually anyway, even if they proxy through a random host first.
> 
> My hope was that this roundabout process would actually give the leader some time to process the join requests.

I accept your point about wait times increasing. Does the microcloud init process give progress feedback to the user indicating its still doing something as each member is joined successfully? That would alleviate some concerns around users getting concerned that nothing is happening.

> My hope was that this roundabout process would actually give the leader some time to process the join requests.

If that is your ultimate aim then it would be far cheaper all round to just put a sleep before starting to join each member.
This isn't something I like the sound of, but its far preferable to bouncing requests around the cluster with the only aim to increase request latency. As that is just going to increase queries on the leader DB (for lookups) and overall increase utilization of the cluster.

tomponline:
> I would argue this was mainly a misuse of the clustering API of LXD and MicroCluster in the first place. Especially on a low-powered machine, handling 10-50 concurrent requests to accept 4 dqlite cluster members (LXD, MicroCloud, MicroOVN, MicroCeph) is sure to cause a bunch of problems.

OK thats a fair point. Where are you seeing the join errors, is it in all off the components (microcloud, lxd, microceph and microovn) or just a subset?

My main concern with all of these sorts of approaches we are discussing here (limiting to arbitrary concurrent numbers, introducing arbitrary delays by sleeps or by routing the request via an intermediate) is that they are arbitrary and likely to still fail on some slower setups and be overly limited on faster ones. At least with sequential join it is as fast as it can be whilst still being reliably correct. I think this should be our primary aim; correctness and reliability over speed. Especially given this is just for cluster initialisation and only happens once per cluster.

>>  Or lock the contended paths internally before this workaround was added?

> Not sure what you mean by this?

My meaning with this was that if each leader system of the respective packages can only reliably handle one joining member at a time, perhaps we should catch new join requests coming in whilst there is one ongoing and either lock/block the request until the ongoing one has started, or we could reject with a known HTTP error (503 maybe https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) that the clients could detect and back off for a random period of time and retry again.

I'm not yet clear *exactly* which component in these packages is causing the failure(s), but if there are inherent limitations in concurrency they should be handled correctly in the server side so the client gets a useful error it can do something with.

masnax:
...What, it rebased successfully once and now it's conflicting again. That's weird, sorry about the ping.

tomponline:
Thanks this can be debased now

masnax:
@tomponline rebased

microcloud/service/microcloud.go
```diff
@@ -106,36 +100,34 @@ func (s CloudService) Join(ctx context.Context, joinConfig JoinConfig) error {
...
 
-			joinedChan <- joinResponse{Name: peer, Error: client.JoinServices(ctx, c, cfg)}
-		}(peer, cfg)
+		return client.JoinServices(ctx, c, joinConfig)
 	}
 
+	joinedChan := make(chan error)
+
+	go func() {
+		joinedChan <- joinFunc()
```
tomponline, line 128: Can we accept a context rather than starting a go routine which could block forever? That way the caller can specify timeout and this function would return if context was canceled. Usually this sort of uncancellable go routine is a code smell to me.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -262,154 +262,112 @@ func lookupPeers(s *service.Handler, autoSetup bool, subnet *net.IPNet, systems
...
-	if len(joinConfig) > initialSize {
-		initialCfg := map[string]types.ServicesPut{}
-		for peer, info := range joinConfig {
-			initialCfg[peer] = info
-
-			if len(initialCfg) == initialSize {
-				break
-			}
+	cloud := sh.Services[types.MicroCloud].(*service.CloudService)
+	errChan := cloud.RequestJoin(context.Background(), secret, peer, cfg)
```
tomponline, line 287: Rather than implementing this timeout, can we pass a context with timeout to RequestJoin?

microcloud/cmd/microcloud/main_init.go
```diff
@@ -262,154 +262,112 @@ func lookupPeers(s *service.Handler, autoSetup bool, subnet *net.IPNet, systems
...
-				return nil
-			}
+	clusterSize := len(cluster)
 
-			if entry.Error != nil {
-				return fmt.Errorf("Peer %q failed to join the cluster: %w", entry.Name, entry.Error)
-			}
+	fmt.Println("Awaiting cluster formation ...")
+	for peer, cfg := range joinConfig {
+		// On a join, roles might still be shuffling, causing dqlite to occasionally block.
```
tomponline, line 353: This comment needs t9 be updated

microcloud/service/microcloud.go
masnax: I had tried this initially with a context timeout, but it didn't work for some reason. If dqlite locked up on cluster join, the whole request would still block.

Looking at it again, it seems it's because I'm using `ConnectLXDUnix` rather than `ConnectLXDUnixWithContext` and `operation.Wait` instead of `operation.WaitContext` so the context wasn't actually propagating to the underlying request. 

That makes this a lot simpler after all, thanks!

tomponline: Excellent! You're welcome 😀

tomponline (approved):

PR #162, opened 2023-09-18
Adds a format flag when listing tokens and cluster members.

### LINKED ISSUE #161: Allow microcloud to output tables in the CLI as json and csv
Just like `lxc` can output information in different formats with the `-f` flag, this would be needed for the microcloud CLI as well.
The motivation behind this originates from the microcloud charm development. Indeed some actions like a node removal (when a charm is scaled down) require to check for the node ROLE before it actually calls `microcloud cluster remove <node_name>` .. A removal when the role is still in `PENDING` state leads to error during the removal (e.g, `Failed to run "['microcloud', 'cluster', 'remove', 'metal-3']": Error: remove /var/snap/microcloud/common/state/truststore/metal-1.yaml: file does not exist` . the certificate was probably not generated yet) 

Having a `json` output format would allow the charm developer to easily track these ROLE states and eventually enqueue an action for later if such a state like `PENDING` is detected.

markylaing:
Closes #161

### DIFF
--- a/microcloud/cmd/microcloud/cluster_members.go
+++ b/microcloud/cmd/microcloud/cluster_members.go
@@ -36,7 +36,8 @@ func (c *cmdClusterMembers) Run(cmd *cobra.Command, args []string) error {
 }
 
 type cmdClusterMembersList struct {
-	common *CmdControl
+	common     *CmdControl
+	flagFormat string
 }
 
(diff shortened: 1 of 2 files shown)


markylaing:
Blocked by #163. Will rebase once that is merged.

tomponline (approved):

PR #164, opened 2023-09-20
Preseed yaml

masnax:
Adds support for preseed init

### DIFF
--- a/microcloud/cmd/microcloud/main_init.go
+++ b/microcloud/cmd/microcloud/main_init.go
@@ -116,7 +116,7 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	err = lookupPeers(s, c.flagAutoSetup, subnet, systems)
+	err = lookupPeers(s, c.flagAutoSetup, subnet, nil, systems)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


masnax:
I've just dropped a somewhat descriptive yaml under the `doc` directory, but we probably want some more comprehensive documentation of this feature.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -139,11 +139,15 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func lookupPeers(s *service.Handler, autoSetup bool, subnet *net.IPNet, systems map[string]InitSystem) error {
+func lookupPeers(s *service.Handler, autoSetup bool, subnet *net.IPNet, expectedSystems []string, systems map[string]InitSystem) error {
```
tomponline, line 152: This function would benefit from a doc block defining what the arguments it accepts are. Especially the new argument, what format is the string slice in and does it do?

Ideally one should be able to use a function by reading its doc block without having to inspect its implementation. It also helps reviewers to check description matches implementation ;)

masnax: Sure, I'll add a comment. In this case it's just a list of hostnames we expect to find. Basically a subset of the keys from the next argument `systems`. 

That said, should one really be using a function without inspecting its implementation, though?

masnax: Just to clarify, I'm not trying to say comments are bad or anything, just being pedantic :)

masnax: Done!

tomponline:
@masnax did you want a review of this now?

microcloud/cmd/microcloud/main_init.go
```diff
@@ -139,11 +139,21 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
...
 }
 
-func lookupPeers(s *service.Handler, autoSetup bool, subnet *net.IPNet, systems map[string]InitSystem) error {
+// lookupPeers attempts to find eligible systems over mDNS, optionally limiting lookup to the given subnet if not nil.
+// Found systems will be progressively added to a table, and the user selection is added to the `systems` map.
+//
+// - If `autoSetup` is true, all systems found in the first 5s will be recorded, and no other input is required.
+// - `expectedSystems` is a list of expected hostnames. If given, the behaviour is similar to `autoSetup`,
+// except it will wait up to a minute for exclusively these systems to be recorded.
+func lookupPeers(s *service.Handler, autoSetup bool, subnet *net.IPNet, expectedSystems []string, systems map[string]InitSystem) error {
```
tomponline, line 158: Thanks for the comment, this explains it well.

This isn't a blocker, just a nit, but the coupling of different timeouts with different modes feels quite peculiar for a utility function like this.

For future reference pushing the timeout decision to the caller and just having a ctx or timeout argument here would make the function description cleaner to understand and separate the concerns of what the function should do by when.

tomponline (approved):

PR #167, opened 2023-09-27
doc: update to the latest version of the starter pack

ru-fu:


tomponline (approved):

doc/custom_conf.py
```diff
@@ -0,0 +1,162 @@
...
+disable_feedback_button = False
+
+############################################################
+### Additional configuration
+############################################################
+
+## Add any configuration that is not covered by the common conf.py file.
+
+intersphinx_mapping = {
+    'lxd': ('https://documentation.ubuntu.com/lxd/en/latest/', None)
```
tomponline, line 161: is this right here?

ru-fu: Yes, this is what we use to link to the LXD documentation. Intersphinx allows to use the Sphinx anchors directly - so instead of linking to `https://documentation.ubuntu.com/lxd/en/latest/`, we can link to `lxd:index`, for example. And we'd only need to update one location if we decided to move servers again. ;)

tomponline: excellent thanks

tomponline (approved):

PR #176, opened 2023-10-04
Preseed support fix for FAN networks

masnax:
If no part of the OVN configuration is specified, MicroCloud should try to set up a FAN network instead. 

If even one system has a specified interface, or if uplink addresses are given, then MicroCloud will try to find defaults for the other systems and set up OVN.


masnax:
For the record, this passes the test suite.

tomponline (approved):

PR #185, opened 2023-10-11
microcloud/version: Bump version to 1.1

masnax:


tomponline:
Where is the api version specified?

tomponline:
Please can you change commit message to:

```
Release MicroCloud 1.1"
```

masnax:
> Where is the api version specified?

MicroCloud doesn't have much of an API, and what little there's is isn't user-accessible, so there's no explicit versioning yet. There's schema versioning built into microcluster but microcloud doesn't extend it yet.

tomponline:
> > Where is the api version specified?
> 
> MicroCloud doesn't have much of an API, and what little there is isn't user-accessible, so there's no explicit versioning yet. There's schema versioning built into microcluster but microcloud doesn't extend it yet.

Hrm are you saying the endpoints don't start with /1.0?

masnax:
> > > Where is the api version specified?
> > 
> > 
> > MicroCloud doesn't have much of an API, and what little there is isn't user-accessible, so there's no explicit versioning yet. There's schema versioning built into microcluster but microcloud doesn't extend it yet.
> 
> Hrm are you saying the endpoints don't start with /1.0?

Sorry, I misunderstood you a bit. I thought you meant a record of API updates like we have with LXD. 

The actual API version is defined by MicroCluster here, it's not managed by MicroCloud:
`https://github.com/canonical/microcluster/blob/e3b7b60a4d71577f5d0f5460f71dbb33dfcb5d9c/internal/rest/client/client.go#L35`

PR #198, opened 2023-10-23
Default to `latest/edge` snap if `MICROCLOUD_SNAP_PATH` unset

masnax:


microcloud/test/includes/microcloud.sh
```diff
@@ -858,8 +866,12 @@ setup_system() {
       done
     "
 
-    lxc file push "${MICROCLOUD_SNAP_PATH}" "${name}"/root/microcloud.snap
-    lxc exec "${name}" -- sh -c "PATH=\$PATH:/snap/bin snap install --devmode /root/microcloud.snap"
+    if [ -n "${MICROCLOUD_SNAP_PATH}" ]; then
+      lxc file push "${MICROCLOUD_SNAP_PATH}" "${name}"/root/microcloud.snap
+      lxc exec "${name}" -- sh -c "PATH=\$PATH:/snap/bin snap install --devmode /root/microcloud.snap"
```
tomponline, line 871: @masnax did you figure out if we can use `--dangerous` here if we install the latest/edge snap first (so the interfaces are setup first)? As using `--devmode` isn't really viable for testing. As we've seen recently.

tomponline: If not then I suspect we should remove this bit and use the new sideloading mechanism instead.

tomponline (approved):

PR #199, opened 2023-10-23
CephFS support

masnax:
Adds a step in the interactive setup after selecting disks for MicroCeph that asks if the user also wants CephFS.

This is also included in the preseed support with a boolean flag in the `storage` section.

### DIFF
--- a/microcloud/service/lxd_config.go
+++ b/microcloud/service/lxd_config.go
@@ -188,3 +188,45 @@ func (s LXDService) DefaultCephStoragePoolJoinConfig() api.ClusterMemberConfigKe
 		Value:  "lxd_remote",
 	}
 }
+
+// DefaultPendingCephFSStoragePool returns the default cephfs storage configuration when
+// creating a pending pool on a specific cluster member target.
+func (s LXDService) DefaultPendingCephFSStoragePool() api.StoragePoolsPost {
+	return api.StoragePoolsPost{
+		Name:   "remote-fs",
+		Driver: "cephfs",
+		StoragePoolPut: api.StoragePoolPut{
+			Config: map[string]string{
+				"source": "lxd_cephfs",
+			},
+		},
+	}
+}
+
+// DefaultCephFSStoragePool returns the default cephfs storage configuration when
+// creating the finalized pool.
+func (s LXDService) DefaultCephFSStoragePool() api.StoragePoolsPost {
+	return api.StoragePoolsPost{
+		Name:   "remote-fs",
+		Driver: "cephfs",
+		StoragePoolPut: api.StoragePoolPut{
+			Config: map[string]string{
+				"cephfs.create_missing": "true",
+				"cephfs.meta_pool":      "lxd_cephfs_meta",
+				"cephfs.data_pool":      "lxd_cephfs_data",
+			},
+			Description: "Distributed file-system storage using CephFS",
+		},
+	}
+}
+
+// DefaultCephFSStoragePoolJoinConfig returns the default cephfs storage configuration when
+// joining an existing cluster.
+func (s LXDService) DefaultCephFSStoragePoolJoinConfig() api.ClusterMemberConfigKey {
+	return api.ClusterMemberConfigKey{
+		Entity: "storage-pool",
+		Name:   "remote-fs",
+		Key:    "source",
+		Value:  "lxd_cephfs",
+	}
+}
--- a/microcloud/test/includes/microcloud.sh
+++ b/microcloud/test/includes/microcloud.sh
@@ -97,6 +97,29 @@ fi
 echo "${setup}" | sed -e '/^\s*#/d' -e '/^\s*$/d'
 }
 
+# set_debug_binaries: Adds {app}.debug binaries if the corresponding {APP}_DEBUG_PATH environment variable is set.
+set_debug_binaries() {
+  name="${1}"
+
+    # Add debug binaries for MicroCloud.
+    if [ -n "${MICROCLOUD_DEBUG_PATH}" ] && [ -n "${MICROCLOUDD_DEBUG_PATH}" ]; then
+      lxc exec "${name}" -- rm -rf /var/snap/microcloud/common/microcloudd.debug
+      lxc exec "${name}" -- rm -rf /var/snap/microcloud/common/microcloud.debug
+
+      lxc file push "${MICROCLOUDD_DEBUG_PATH}" "${name}"/var/snap/microcloud/common/microcloudd.debug
+      lxc file push "${MICROCLOUD_DEBUG_PATH}" "${name}"/var/snap/microcloud/common/microcloud.debug
+
+      lxc exec "${name}" -- systemctl restart snap.microcloud.daemon || true
+    fi
+
+    # Add a debug binary for LXD.
+    if [ -n "${LXD_DEBUG_PATH}" ]; then
+      lxc exec "${name}" -- rm -rf /var/snap/lxd/common/lxd.debug
+      lxc file push "${LXD_DEBUG_PATH}" "${name}"/var/snap/lxd/common/lxd.debug
+      lxc exec "${name}" -- systemctl reload snap.lxd.daemon || true
+      lxc exec "${name}" -- lxd waitready
+    fi
+}
 
 # set_remote: Adds and switches to the remote for the MicroCloud node with the given name.
 set_remote() {
@@ -381,16 +404,17 @@ reset_snaps() {
     # These are set to always pass in case the snaps are already disabled.
     echo "Disabling LXD and MicroCloud for ${name}"
     lxc exec "${name}" -- sh -c "
-      if pidof -q lxd ; then
-        kill -9 \$(pidof lxd)
-      fi
+      rm -rf /var/snap/lxd/common/lxd.debug
+      rm -rf /var/snap/microcloud/common/microcloudd.debug
+      rm -rf /var/snap/microcloud/common/microcloud.debug
+
+      for app in lxd lxd.debug microcloud microcloud.debug microcloudd microcloudd.debug ; do
+        if pidof -q \${app} > /dev/null; then
+          kill -9 \$(pidof \${app})
+        fi
+      done
 
       snap disable lxd > /dev/null 2>&1 || true
-
-      if pidof -q microcloud ; then
-        kill -9 \$(pidof microcloud)
-      fi
-
       snap disable microcloud > /dev/null 2>&1 || true
 
       systemctl stop snap.lxd.daemon snap.lxd.daemon.unix.socket > /dev/null 2>&1 || true
@@ -455,6 +479,8 @@ reset_snaps() {
 
       lxd waitready
     "
+
+    set_debug_binaries "${name}"
   )
 }
 
@@ -924,6 +950,8 @@ setup_system() {
     else
       lxc exec "${name}" -- sh -c "PATH=\$PATH:/snap/bin snap install microcloud --channel latest/edge"
     fi
+
+    set_debug_binaries "${name}"
   )
 
   # Sleep some time so the snaps are fully set up.
--- a/microcloud/test/includes/microcloud.sh
+++ b/microcloud/test/includes/microcloud.sh
@@ -539,6 +539,10 @@ cluster_reset() {
           microceph.ceph tell mon.\* injectargs '--mon-allow-pool-delete=true'
           lxc storage rm remote || true
           microceph.rados purge lxd_remote --yes-i-really-really-mean-it --force
+          microceph.ceph fs fail lxd_cephfs || true
+          microceph.ceph fs rm lxd_cephfs  --yes-i-really-mean-it || true
+          microceph.rados purge lxd_cephfs_meta --yes-i-really-really-mean-it --force || true
+          microceph.rados purge lxd_cephfs_data --yes-i-really-really-mean-it --force || true
           microceph.rados purge .mgr --yes-i-really-really-mean-it --force
 
           for pool in \$(microceph.ceph osd pool ls) ; do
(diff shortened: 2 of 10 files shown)


microcloud/test/suites/preseed.sh
```diff
@@ -94,6 +97,7 @@ EOF
   lxc exec micro01 -- sh -c "
   cat << EOF > /root/preseed.yaml
 lookup_subnet: ${lookup_addr}/24
+lookup_interface: enp5s0
```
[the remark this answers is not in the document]
tomponline: Also are these really related to cephfs, shouldnt they be in a separate commit?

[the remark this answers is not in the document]
tomponline: Ah looks like you meant to include this change later after c45a983429b063a6070203606251ae1f1186f1ab

microcloud/test/includes/microcloud.sh
```diff
@@ -567,11 +567,14 @@ cluster_reset() {
             microceph.ceph osd pool rm \${pool} \${pool} --yes-i-really-really-mean-it
           done
 
-          for pool in \$(microceph.ceph osd ls) ; do
-            microceph.ceph osd out \${pool}
-            microceph.ceph osd down \${pool} --definitely-dead
-            microceph.ceph osd purge \${pool} --yes-i-really-mean-it --force
-            microceph.ceph osd destroy \${pool} --yes-i-really-mean-it --force
+          # OSDs may re-up themselves before we call destroy, so retry until they are all actually gone.
```
tomponline, line 589: @masnax please can you ask the microceph team about this?

microcloud/test/includes/microcloud.sh
```diff
@@ -442,6 +442,25 @@ reset_snaps() {
...
+
+    # Add debug binaries for MicroCloud.
+    if [ -n "${MICROCLOUD_DEBUG_PATH}" ] && [ -n "${MICROCLOUDD_DEBUG_PATH}" ]; then
+      lxc exec "${name}" -- rm -rf /var/snap/microcloud/common/microcloudd.debug
+      lxc exec "${name}" -- rm -rf /var/snap/microcloud/common/microcloud.debug
+
+      lxc file push "${MICROCLOUDD_DEBUG_PATH}" "${name}"/var/snap/microcloud/common/microcloudd.debug
+      lxc file push "${MICROCLOUD_DEBUG_PATH}" "${name}"/var/snap/microcloud/common/microcloud.debug
+
+      lxc exec "${name}" -- sh -c "systemctl restart snap.microcloud.daemon"
```
tomponline, line 454: out of interest, why do you use `sh -c` here rather than just calling the command like:

```
lxc exec "${name}" -- systemctl restart snap.microcloud.daemon
```

tomponline (commented):
@masnax @ru-fu will this require any documentation updates to cover the new functionality?

@masnax Also I was thinking, has any consideration been given to how to detect whether the LXD version being used supports setting up cephfs in this way? Are we just going to give an error about unsupported cephfs config keys if an older version of LXD is used? Would it be possible to check the API extensions and indicate more clearly what the issue is?

@masnax Finally, this PR included some other changes not related to cephfs support, it would be useful to reviewers to include a brief description of additional changes in the PR description so they have the context for the change when reviewing. Thanks!

ru-fu:
> @masnax @ru-fu will this require any documentation updates to cover the new functionality?

Yes, it will need updates to https://canonical-microcloud.readthedocs-hosted.com/en/latest/tutorial/get_started/ and https://canonical-microcloud.readthedocs-hosted.com/en/latest/how-to/initialise/ (both have the initialisation steps).
We should probably also add some information to https://canonical-microcloud.readthedocs-hosted.com/en/latest/explanation/microcloud/#storage to explain why/when you want to enable it.

microcloud/test/suites/preseed.sh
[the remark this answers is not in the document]
masnax: Ah yeah oops that one belongs to a different commit. I included these here because the `preseed` tests were failing. 

I suppose we don't technically _need_ to require the interface name as well, but it does capture the edge case of 2 different interfaces sharing the same subnet or IP.

Since `microcoud init` will just present you with an entry for every address and every interface, it means if we don't include the interface name in `preseed` that it will actually be less precise than the interactive CLI at this step.

microcloud/test/includes/microcloud.sh
masnax: Sure I'll discuss with them. I don't think this should be a blocker because this whole section is a bit hacky and needs an environment flag to enable. In this section, rather than restoring from snapshot and waiting for the VM agent, we manually disable the snaps, wipe their state directories, clear out any stray modules and processes, and then re-enable the snaps.

It's convenient as it speeds up local testing, but far less clean than just restoring from snapshot, which is the default behaviour of the test suite.

microcloud/test/includes/microcloud.sh
masnax: Habit, really. the shorter way is, well, shorter so I'll switch it :)

masnax:
> @masnax Also I was thinking, has any consideration been given to how to detect whether the LXD version being used supports setting up cephfs in this way? Are we just going to give an error about unsupported cephfs config keys if an older version of LXD is used? Would it be possible to check the API extensions and indicate more clearly what the issue is?

Yes I was hoping this discussion would come up. Since MicroCloud is released now, we'll need to start thinking about which versions of LXD/MicroOVN/MicroCeph should be supported. If MicroCloud depends on a more recent change in LXD, we can check for an API extension in MicroCloud, but if LXD support starts at `latest/stable`, then eventually that API extension check will become redundant.

This is also an issue with the MicroOVN content interface. That doesn't even have an API extension because it's a snap-exclusive change so we have no way of judging in MicroCloud about whether or not we need to restart LXD to get it to pick up MicroOVN and MicroCeph, or if LXD is capable of finding them via the content interface. 


> @masnax Finally, this PR included some other changes not related to cephfs support, it would be useful to reviewers to include a brief description of additional changes in the PR description so they have the context for the change when reviewing. Thanks!

Will do, thanks. In this case the changes were because the tests were broken due to the changes that were made in Riga to address MicroCloud using the wrong interface for mDNS lookup/broadcast.

tomponline:
> If MicroCloud depends on a more recent change in LXD, we can check for an API extension in MicroCloud, but if LXD support starts at `latest/stable`, then eventually that API extension check will become redundant if the version of LXD in `latest/stable` started with that API extension.


Can we at least detect if LXD supports the features being requested (as cephfs is optional right?) and give a clear error message as to the issue? Rather than just passing on the LXD unrecognised config option error.

masnax:
> > If MicroCloud depends on a more recent change in LXD, we can check for an API extension in MicroCloud, but if LXD support starts at `latest/stable`, then eventually that API extension check will become redundant if the version of LXD in `latest/stable` started with that API extension.
> 
> Can we at least detect if LXD supports the features being requested (as cephfs is optional right?) and give a clear error message as to the issue? Rather than just passing on the LXD unrecognised config option error.

So we don't allow bypassing CephFS, but just error out with a more descriptive error. I'm happy with that.

I am still a bit concerned that to arbitrarily do this for some API extensions and not others is going to cause confusion down the line. Eventually we'll lose the context of why we included it for cephfs, but not for some other API extension that existed prior to MicroCloud, because the minimum version that MicroCloud supports would already include `storage_cephfs_create_missing`, just like the minimum version of LXD today already supports all the pre-existing API extensions.

The other option is to check for a API version/count right when starting MicroCloud, and expect LXD to be at or above that. That way we'll error out before hitting any particular API extension that we require.

tomponline:
I think we should check for compatible versions of lxd, microceph and microovn indeed, but where possible if we can support using a lesser version albeit with degraded functionality (and/or clear error messages as to why something isnt available) then all the better.

masnax:
> I think we should check for compatible versions of lxd, microceph and microovn indeed, but where possible if we can support using a lesser version albeit with degraded functionality (and/or clear error messages as to why something isnt available) then all the better.

So that sounds to me like we should eliminate the "arbitrary" part and check for API extensions wherever it makes sense. I'll do it for cephfs here and add an issue to track where we can do this in other places as well.

microcloud/service/lxd_config.go
```diff
@@ -184,3 +184,45 @@ func (s LXDService) DefaultCephStoragePoolJoinConfig() api.ClusterMemberConfigKe
...
+	return api.StoragePoolsPost{
+		Name:   "remote-fs",
+		Driver: "cephfs",
+		StoragePoolPut: api.StoragePoolPut{
+			Config: map[string]string{
+				"cephfs.create_missing": "true",
+				"cephfs.meta_pool":      "lxd_cephfs_meta",
+				"cephfs.data_pool":      "lxd_cephfs_data",
+			},
+			Description: "Remote CephFS file-system",
```
roosterfish, line 214: To be in line with the description in `DefaultCephStoragePool()` maybe this can be changed to something like "Distributed file-system storage using CephFS"?

microcloud/test/includes/microcloud.sh
```diff
@@ -910,6 +929,26 @@ setup_system() {
     else
       lxc exec "${name}" -- sh -c "PATH=\$PATH:/snap/bin snap install microcloud --channel latest/edge"
     fi
+
+
+    # Add debug binaries for MicroCloud.
```
roosterfish, line 934: Can you please put these blocks for adding the debug binaries (MicroCloud and LXD) into its own functions so they can be called at both places?

tomponline:
@masnax also, as we add new features to microcloud, do we also need to add the concept of API extensions, so that if clients to MicroCloud (such as the charm?) consume the API they can detect which features are available?

masnax:
> @masnax also, as we add new features to microcloud, do we also need to add the concept of API extensions, so that if clients to MicroCloud (such as the charm?) consume the API they can detect which features are available?

That might be something useful to bake into microcluster, but in this case there's no API change. The only thing that the microcloud API does at this time is proxy requests to the other services.

tomponline:
> > @masnax also, as we add new features to microcloud, do we also need to add the concept of API extensions, so that if clients to MicroCloud (such as the charm?) consume the API they can detect which features are available?
> 
> That might be something useful to bake into microcluster, but in this case there's no API change. The only thing that the microcloud API does at this time is proxy requests to the other services.

Doesnt this add support for new API interactions to indicate cephfs support?

microcloud/test/includes/microcloud.sh
masnax: This loop's been removed now in favour of following the exact process as MicroCeph when removing disks (but without the slow validation steps).

masnax:
@tomponline This will now skip the `cephfs` blocks if LXD doesn't support the cephfs create_missing API extension.

microcloud/test/includes/microcloud.sh
```diff
@@ -498,6 +498,10 @@ reset_system() {
 
     lxc exec "${name}" -- ip link del lxdfan0 || true
 
+    # Resync the time in case it got out of sync with the other VMs.
```
tomponline, line 501: This seem suspect. Any ideas @simondeziel ?

microcloud/test/includes/microcloud.sh
```diff
@@ -2,7 +2,7 @@
 unset_interactive_vars() {
   unset LOOKUP_IFACE LIMIT_SUBNET SKIP_SERVICE EXPECT_PEERS \
     SETUP_ZFS ZFS_FILTER ZFS_WIPE \
-    SETUP_CEPH CEPH_WARNING CEPH_FILTER CEPH_WIPE \
+    SETUP_CEPH CEPH_WARNING CEPH_FILTER CEPH_WIPE SETUP_CEPHFS \
```
tomponline, line 5: In this commit, is the cephfs pool actually checked for valid functionality, i.e by creating a cephfs volume and attaching it to a running VM and container? I'd like to avoid the issues we saw prior to the 1.1 release where we found LXD VMs couldnt be launched on ceph. Apologies if I am not reading this correctly and its already there.

microcloud/test/includes/microcloud.sh
simondeziel: @masnax is this due to VM snapshots being restored or something similar?

masnax: Actually no. It seems they drift while staying up too. I occasionally see a 5 minute drift on the certs that are generated when the state directory doesn't exist.

microcloud/test/includes/microcloud.sh
masnax: Added now.

microcloud/test/includes/microcloud.sh
masnax: It's just hovering at 0.0-0.1, occasionally hitting 0.3. But you may have a point, it was particularly noticeable when I was debugging dqlite by having 10 nodes concurrently join the cluster over and over and over.

masnax:
@roosterfish Could you take a look this one last time so we can merge it? Thanks

microcloud/test/includes/microcloud.sh
```diff
@@ -350,16 +404,17 @@ reset_snaps() {
...
+      rm -rf /var/snap/lxd/common/lxd.debug
+      rm -rf /var/snap/microcloud/common/microcloudd.debug
+      rm -rf /var/snap/microcloud/common/microcloud.debug
 
-      snap disable lxd > /dev/null 2>&1 || true
-
-      if pidof -q microcloud ; then
-        kill -9 \$(pidof microcloud)
-      fi
+      for app in lxd lxd.debug microcloud microcloud.debug ; do
```
roosterfish, line 411: Sorry if I am missing something but shouldn't this line include `microcloudd microcloudd.debug`? I guess the daemon should be killed here.

roosterfish (commented):
Still looking good, just a single remark on the test.

microcloud/test/includes/microcloud.sh
masnax: Yes, good catch. `microcloud` is just the client.

roosterfish (approved):
LGTM!

tomponline (approved):

PR #200, opened 2023-10-24
doc: add a cheat sheet for common commands

ru-fu:
Include a command list, linking to LXD documentation for more information.

### DIFF
--- a/doc/.custom_wordlist.txt
+++ b/doc/.custom_wordlist.txt
@@ -1,3 +1,5 @@
+ACL
+ACLs
 AGPL
 backend
 Ceph
--- a/doc/.sphinx/spellingcheck.yaml
+++ b/doc/.sphinx/spellingcheck.yaml
@@ -20,6 +20,7 @@ matrix:
       - code
       - pre
       - spellexception
+      - .spellexception
       - link
       - title
       - div.relatedlinks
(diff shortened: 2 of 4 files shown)


ru-fu:
Preview here: https://canonical-microcloud--200.com.readthedocs.build/en/200/how-to/commands/

camille-rodriguez:
I'd suggest calling it what it is - a cheatsheet. People commonly use that keyword to search for this kind of information. The name of the page and the tab could be "MicroCloud Cheatsheet". The tab name "work with MicroCloud" doesn't really ring the same bell in my mind. It might also be easier for SOE online. Otherwise, nicely done!

ru-fu:
> I'd suggest calling it what it is - a cheatsheet. People commonly use that keyword to search for this kind of information. The name of the page and the tab could be "MicroCloud Cheatsheet". The tab name "work with MicroCloud" doesn't really ring the same bell in my mind. It might also be easier for SOE online. Otherwise, nicely done!

Hmm, it isn't really a word though, and it doesn't work well with Diataxis either, so in the navigation I prefer to keep it as is.
But I'm fine with renaming the page from "How to work with MicroCloud (common commands)" to "How to work with MicroCloud (cheat sheet)", if you think that helps?
Otherwise, I can include "cheat sheet" in the text somewhere. You're right that it helps with SEO.

tomponline (approved):

PR #204, opened 2023-10-31
Specify lookup interface to mdns handler

masnax:
Additionally adds a `lookup_interface` key to the preseed so that can have the same behaviour.


masnax:
@tomponline here's the PR, I'm running the testsuite on it at the moment, but you can also try to build it and use `microcloudd.debug` to see if it solves the problem.

masnax:
Tests are now passing.

tomponline:
Cheers, this seems to have fixed the issue, and i can get further into the setup process now. Nice one!

tomponline (approved):
Thanks for fast turn around!

Issue #212, opened 2023-11-10
unable to cluster over vlan sub-interface


baycarbone:
I'm trying to setup a microcloud cluster by picking a vlan sub-interface for the cluster internal traffic, but although network connectivity is fine (nodes can ping each other for example), the microcloud snap is not able to discover the other nodes.

 When taking traces on the initiating node and the other nodes, I can see the MDNS requests arriving on the other nodes, but no reply is sent back.

Here is the `microcloud init` output:
```
ubuntu@micro-node10:~$ sudo microcloud init
Waiting for LXD to start...
Select an address for MicroCloud's internal traffic:
Space to select; enter to confirm; type to filter results.
Up/down to move; right to select all; left to select none.
       +--------------+---------+
       |   ADDRESS    |  IFACE  |
       +--------------+---------+
  [ ]  | 172.27.41.34 | enp0s25 |
  [ ]  | 10.102.0.4   | ext.102 |
  [ ]  | 10.104.0.4   | ext.104 |
> [x]  | 10.103.0.4   | ext.103 |
  [ ]  | 10.101.0.4   | ext.101 |
       +--------------+---------+
```

Here is the network config on 2 nodes:
```
ubuntu@micro-node10:~$ ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: enp0s25: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether b8:ae:ed:7b:f3:42 brd ff:ff:ff:ff:ff:ff
    inet 172.27.41.34/23 brd 172.27.41.255 scope global enp0s25
       valid_lft forever preferred_lft forever
    inet6 fe80::baae:edff:fe7b:f342/64 scope link
       valid_lft forever preferred_lft forever
3: ext: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 8c:ae:4c:df:53:48 brd ff:ff:ff:ff:ff:ff
    inet6 fe80::8eae:4cff:fedf:5348/64 scope link
[...]

PR #213, opened 2023-11-10
microcloud/service: Properly broadcast available interfaces over those interfaces

masnax:
Lookup was recently updated to lookup mDNS records via a particular interface, but the mDNS broadcasts were always coming from the default interface, even when advertising a different interface, so the lookup would ignore these records as they aren't coming from an interface that it expected.


tomponline (approved):
Thanks!

tomponline:
Should fix https://github.com/canonical/microcloud/issues/212

PR #220, opened 2023-11-17
doc: add information about fan network

ru-fu:
When not using MicroOVN, the Ubuntu fan is used instead - document the limitations.

### DIFF
--- a/doc/explanation/microcloud.rst
+++ b/doc/explanation/microcloud.rst
@@ -24,15 +24,20 @@ See :ref:`lxd:clustering` in the LXD documentation for more information.
 Networking
 ----------
 
-.. include:: ../../README.md
-   :parser: myst_parser.sphinx_
-   :start-after: ## **What about networking?**
-   :end-before: ## **What's next?**
+By default, MicroCloud uses MicroOVN for networking, which is a minimal wrapper around OVN (Open Virtual Network).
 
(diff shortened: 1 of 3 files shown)


doc/explanation/microcloud.rst
```diff
@@ -24,16 +24,21 @@ See :ref:`lxd:clustering` in the LXD documentation for more information.
 Networking
 ----------
 
-.. include:: ../../README.md
-   :parser: myst_parser.sphinx_
-   :start-after: ## **What about networking?**
-   :end-before: ## **What's next?**
+By default, MicroCloud uses MicroOVN for networking, which is a minimal implementation of OVN (Open Virtual Network).
```
tomponline, line 27: I don't think we should say its a minimal implementation of ovn. How about a "minimal alternative to OVN"?

ru-fu: Changed to "minimal wrapper around OVN".

tomponline: No that isn't correct I'm afraid, as Ubuntu FAN isn't anything to do with OVN.

ru-fu: Hmm, not sure I'm following - we're talking about MicroOVN here and not about Ubuntu Fan?

tomponline (approved):
Ah OK I got confused as the PR description said "When not using MicroOVN, the Ubuntu fan is used instead - document the limitations" so I had read the minimal implementation as referring to the fan rather than microovn vs ovn.

PR #227, opened 2023-11-20
Fix ceph disk validation

### LINKED ISSUE #226: Preseed: Direct disk selection expects a filter to exist
When directly selecting disks for ceph storage, if not all systems have a disk, then MicroCloud will expect the user to have included a filter in the preseed configuration. This makes sense for zfs storage as every system should have a disk, but doesn't make sense for ceph, where we just need a minimum of 3 spread over 3 peers.

masnax:
Closes #226 

Currently when directly specifying disks for ceph, MicroCloud will require all systems to specify a disk unless a filter is also specified. 

Instead, we should only require 3 systems to specify a disk, and only during bootstrap.

### DIFF
--- a/microcloud/cmd/microcloud/main_init_preseed.go
+++ b/microcloud/cmd/microcloud/main_init_preseed.go
@@ -214,8 +214,8 @@ func (p *Preseed) validate(name string, bootstrap bool) error {
 	}
 
 	containsCephStorage = directCephCount > 0
-	if containsCephStorage && directCephCount < len(p.Systems) && len(p.Storage.Ceph) == 0 {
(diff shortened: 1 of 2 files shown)


tomponline:
Should we have a test for this?

tomponline (approved):
Should we have a test for this though?

masnax:
> Should we have a test for this though?

The exact configuration is already tested with the interactive CLI, it was just explicitly blocked for preseeds. I've now added a unit test that covers the validation component that was blocking it.

tomponline (approved):

tomponline:
@masnax please can you sign your commits.

PR #228, opened 2023-11-20
feat: Give option to specify custom DNS addresses for OVN uplink

gabrielmougard:
Introduces a new feature that allows users to specify the DNS servers for the UPLINK network during the MicroCloud initialization process. Previously, MicroCloud configured the instances to use the gateway IP as the DNS server by default. Now you can set these values interactively like so:

```bash
Specify the default DNS addresses (comma-separated IPv4 addresses) for this OVN network instead of using the default gateway address? (empty to skip): <COMMA_SEPARATED_IPV4_ADDR>
```

You can also add this behaviour in the preseed file like so:

```yaml
...
ovn:
  ipv4_gateway: 10.1.123.1/24
  ipv4_range: 10.1.123.100-10.1.123.254
  ipv6_gateway: fd42:1:1234:1234::1/64
  dns_servers: 10.1.123.1,8.8.8.8 # now we can add custom DNS addresses to the OVN network like so
```

### DIFF
--- a/microcloud/cmd/microcloud/ask.go
+++ b/microcloud/cmd/microcloud/ask.go
@@ -11,6 +11,7 @@ import (
 	cli "github.com/canonical/lxd/shared/cmd"
 	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/lxd/shared/units"
+	"github.com/canonical/lxd/shared/validate"
 	cephTypes "github.com/canonical/microceph/microceph/api/types"
 
 	"github.com/canonical/microcloud/microcloud/api/types"
@@ -643,7 +644,8 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 	}
 
 	// Prepare the configuration.
-	config := map[string]string{}
+	ipConfig := map[string]string{}
+	dnsConfig := map[string]string{}
 	if bootstrap {
 		for _, ip := range []string{"IPv4", "IPv6"} {
 			validator := func(s string) error {
@@ -674,37 +676,34 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 			}
 
 			if gateway != "" {
-				validator := func(s string) error {
-					addr := net.ParseIP(s)
-					if addr == nil {
-						return fmt.Errorf("Invalid IP address %q", s)
-					}
-
-					if addr.To4() == nil && ip == "IPv4" {
-						return fmt.Errorf("Not a valid IPv4")
-					}
-
-					if addr.To4() != nil && ip == "IPv6" {
-						return fmt.Errorf("Not a valid IPv6")
-					}
-
-					return nil
-				}
-
 				if ip == "IPv4" {
-					rangeStart, err := c.asker.AskString(fmt.Sprintf("Specify the first %s address in the range to use with LXD: ", ip), "", validator)
+					rangeStart, err := c.asker.AskString(fmt.Sprintf("Specify the first %s address in the range to use with LXD: ", ip), "", validate.Required(validate.IsNetworkAddressV4))
 					if err != nil {
 						return err
 					}
 
-					rangeEnd, err := c.asker.AskString(fmt.Sprintf("Specify the last %s address in the range to use with LXD: ", ip), "", validator)
+					rangeEnd, err := c.asker.AskString(fmt.Sprintf("Specify the last %s address in the range to use with LXD: ", ip), "", validate.Required(validate.IsNetworkAddressV4))
 					if err != nil {
 						return err
 					}
 
-					config[gateway] = fmt.Sprintf("%s-%s", rangeStart, rangeEnd)
+					ipConfig[gateway] = fmt.Sprintf("%s-%s", rangeStart, rangeEnd)
+
+					gatewayAddr, _, err := net.ParseCIDR(gateway)
+					if err != nil {
+						return err
+					}
+
+					dnsAddresses, err := c.asker.AskString(fmt.Sprintf("Specify the DNS addresses (comma-separated IPv4 / IPv6 addresses) for the distributed network (default: %s): ", gatewayAddr.String()), gatewayAddr.String(), validate.Optional(validate.IsListOf(validate.IsNetworkAddress)))
+					if err != nil {
+						return err
+					}
+
+					if dnsAddresses != "" {
+						dnsConfig[gateway] = dnsAddresses
+					}
 				} else {
-					config[gateway] = ""
+					ipConfig[gateway] = ""
 				}
 			}
 		}
@@ -742,7 +741,7 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 		var ipv4Gateway string
 		var ipv4Ranges string
 		var ipv6Gateway string
-		for gateway, ipRange := range config {
+		for gateway, ipRange := range ipConfig {
 			ip, _, err := net.ParseCIDR(gateway)
 			if err != nil {
 				return err
@@ -756,7 +755,12 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 			}
 		}
 
-		uplink, ovn := lxd.DefaultOVNNetwork(ipv4Gateway, ipv4Ranges, ipv6Gateway)
+		var allDNSServers string
+		for _, dnsAddr := range dnsConfig {
+			allDNSServers = dnsAddr
+		}
+
+		uplink, ovn := lxd.DefaultOVNNetwork(ipv4Gateway, ipv4Ranges, ipv6Gateway, allDNSServers)
 		bootstrapSystem.Networks = []api.NetworksPost{uplink, ovn}
 		systems[sh.Name] = bootstrapSystem
 	}
--- a/microcloud/test/suites/preseed.sh
+++ b/microcloud/test/suites/preseed.sh
@@ -29,6 +29,7 @@ ovn:
   ipv4_gateway: 10.1.123.1/24
   ipv4_range: 10.1.123.100-10.1.123.254
   ipv6_gateway: fd42:1:1234:1234::1/64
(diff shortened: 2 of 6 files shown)


masnax:
First thought, you can make it one question without having the yes/no, and just passing empty to skip, like https://github.com/canonical/microcloud/blob/0c3009cf15e68b1f58726489f7c699c14e35e6b9/microcloud/cmd/microcloud/ask.go#L670

tomponline:
Please can you update the PR description with an example of how this would look in the overall initialization process.

microcloud/test/suites/preseed.sh
```diff
@@ -29,6 +29,7 @@ ovn:
   ipv4_gateway: 10.1.123.1/24
   ipv4_range: 10.1.123.100-10.1.123.254
   ipv6_gateway: fd42:1:1234:1234::1/64
+  dns_servers: 10.1.123.1,8.8.8.8
```
tomponline, line 32: please help me understand where the tests check that these settings are applied to LXD?

gabrielmougard: For now, there's no check. I'll add that, thanks!

gabrielmougard: should be added now

gabrielmougard:
@simondeziel how is it possible to run the integration test suite in the CI?

microcloud/cmd/microcloud/ask.go
```diff
@@ -691,6 +692,22 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 					return nil
 				}
 
+				commaValidator := func(s string) error {
```
tomponline, line 695: @gabrielmougard can you use `validate.Optional(validate.IsListOf(validate.IsNetworkAddress)),` the same as LXD does here?

tomponline:
> @simondeziel how is it possible to run the integration test suite in the CI?

Its not, it has to be run manually for now. We have no CI that supports nested virt.

microcloud/cmd/microcloud/ask.go
```diff
@@ -702,9 +719,18 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 						return err
 					}
 
-					config[gateway] = fmt.Sprintf("%s-%s", rangeStart, rangeEnd)
+					ipConfig[gateway] = fmt.Sprintf("%s-%s", rangeStart, rangeEnd)
+
+					dnsAddresses, err := c.asker.AskString("Specify the default DNS addresses (comma-separated IPv4 addresses) for this OVN network instead of using the default gateway address? (empty to skip): ", "", commaValidator)
```
tomponline, line 724: Why only IPv4?

tomponline: This option is specifying the IPs of DNS servers reachable on the uplink network that will be used for external DNS resolution by OVN networks using that uplink.

So how about:

> Specify the DNS servers (comma-separated IP addresses) for OVN networks to use instead of using the default gateway address? (empty to skip):

Also, have we mentioned OVN in `microcloud init` up until this point before?

gabrielmougard: Good catch thanks. I'll add IPv6.

simondeziel:
> > @simondeziel how is it possible to run the integration test suite in the CI?
> 
> Its not, it has to be run manually for now. We have no CI that supports nested virt.

That's apparently about to change as we've seen a few GitHub-hosted Action runners with nested-virt enabled. That said, this will probably not be enough for microcloud due to the small CPU/RAM allocation on the runner. Could be interesting for LXD though :)

microcloud/cmd/microcloud/ask.go
```diff
@@ -756,7 +759,35 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
...
+
+		if ipv6DNSNameServers != "" {
+			if allDNSServers != "" {
+				allDNSServers += "," + ipv6DNSNameServers
+			} else {
+				allDNSServers = ipv6DNSNameServers
+			}
+		}
+
+		uplink, ovn := lxd.DefaultOVNNetwork(ipv4Gateway, ipv4Ranges, ipv6Gateway, allDNSServers)
```
masnax, line 790: So here, it looks to me like you're fetching the results from each of the ipv4 and ipv6 dns questions, and then just appending them to the same list, which gets stored in the OVN config as `dns.nameservers`. Which says to me that you probably don't need 2 separate questions for the nameservers.

How about having just 1 question where the user can enter both ipv4 and ipv6 addresses? Then we won't need this extra IP parsing and version checking.

microcloud/test/suites/preseed.sh
```diff
@@ -56,6 +57,13 @@ EOF
     validate_system_microovn ${m}
   done
 
+  # Check that the created UPLINK network has the right DNS servers.
+  dns_addresses=$(lxc exec micro01 -- sh -c "lxc network get UPLINK dns.nameservers")
+  if [ "${dns_addresses}" != "10.1.123.1,8.8.8.8" ] ; then
+    echo "ERROR: UPLINK network has wrong DNS server addresses: ${dns_addresses}"
+    return 1
+  fi
+
```
masnax, line 66: Since this config key affects the input/output of every interactive CLI test, can you please move this logic over here:
https://github.com/canonical/microcloud/blob/0c3009cf15e68b1f58726489f7c699c14e35e6b9/microcloud/test/includes/microcloud.sh#L198

That validator is called from https://github.com/canonical/microcloud/blob/0c3009cf15e68b1f58726489f7c699c14e35e6b9/microcloud/test/includes/microcloud.sh#L255

So you can add an argument to `validate_system_lxd` for the comma-delimited list of nameservers, and have it send the args down to `validate_system_lxd_ovn` which will check the result in LXD if the argument is set. This will just make it easier to write future tests for the CLI and be able to consistently validate the resulting cluster.

masnax (changes requested):
Looks good, just a couple suggestions :)

microcloud/cmd/microcloud/ask.go
tomponline: @masnax @gabrielmougard yes agreed, we shouldn't ask the DNS question twice, just once and accept IPv4 and IPv6 addresses, this is what I mentioned a few days ago here:

https://github.com/canonical/microcloud/pull/228#discussion_r1403270125

tomponline (changes requested):
@gabrielmougard once you've worked on @masnax please could you post here an example of the `microcloud init` output so we can see what the end result looks like. Thanks

gabrielmougard:
Example of `microcloud init` :

```
root@micro01:~# microcloud init
Waiting for LXD to start...
Select an address for MicroCloud's internal traffic:

 Using address "10.47.247.229" for MicroCloud

Limit search for other MicroCloud servers to 10.47.247.229/24? (yes/no) [default=yes]: 
Scanning for eligible servers ...

 Selected "micro01" at "10.47.247.229"
 Selected "micro03" at "10.47.247.25"
 Selected "micro02" at "10.47.247.233"

Would you like to set up local storage? (yes/no) [default=yes]: 
Select exactly one disk from each cluster member:

Select which disks to wipe:

 Using "/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_disk1" on "micro01" for local storage pool
 Using "/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_disk1" on "micro02" for local storage pool
 Using "/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_disk1" on "micro03" for local storage pool

Would you like to set up distributed storage? (yes/no) [default=yes]: 
Select from the available unpartitioned disks:

Select which disks to wipe:

 Using 2 disk(s) on "micro01" for remote storage pool
 Using 2 disk(s) on "micro03" for remote storage pool
 Using 2 disk(s) on "micro02" for remote storage pool

Configure distributed networking? (yes/no) [default=yes]: 
Select exactly one network interface from each cluster member:

 Using "enp6s0" on "micro01" for OVN uplink
 Using "enp6s0" on "micro03" for OVN uplink
 Using "enp6s0" on "micro02" for OVN uplink

Specify the IPv4 gateway (CIDR) on the uplink network (empty to skip IPv4): 10.47.247.1/24
Specify the first IPv4 address in the range to use with LXD: 10.47.247.100
Specify the last IPv4 address in the range to use with LXD: 10.47.247.120
Specify the default DNS addresses (comma-separated IPv4 / IPv6 addresses) for this OVN network instead of using the default gateway address? (empty to skip): 10.47.247.1,8.8.8.8,::ffff:a2f:f701
Specify the IPv6 gateway (CIDR) on the uplink network (empty to skip IPv6): 
Initializing a new cluster
 Local MicroCloud is ready
 Local LXD is ready
 Local MicroOVN is ready
 Local MicroCeph is ready
Awaiting cluster formation ...
 Peer "micro03" has joined the cluster
 Peer "micro02" has joined the cluster
Configuring cluster-wide devices ...
MicroCloud is ready
root@micro01:~# lxc network get UPLINK dns.nameservers
If this is your first time running LXD on this machine, you should also run: lxd init
To start your first container, try: lxc launch ubuntu:22.04
Or for a virtual machine: lxc launch ubuntu:22.04 --vm

10.47.247.1,8.8.8.8,::ffff:a2f:f701
root@micro01:~# 
```

masnax:
@gabrielmougard So the line is
```
Specify the default DNS addresses (comma-separated IPv4 / IPv6 addresses) for this OVN network instead of using the default gateway address? (empty to skip): 10.47.247.1,8.8.8.8,::ffff:a2f:f701
```

Which is a bit long to me, how about something like
```
Specify the DNS addresses (comma-separated IPv4 / IPv6 addresses) for the distributed network (default: <the actual default gateway address>):
```

The changes here are to bring the question in line with the other network questions by omitting OVN and just calling it a "distributed" network, and showing the default gateway as the default rather than just saying it's the default. Personally I'd prefer seeing what the default is to just being told to fallback to a default configuration.

tomponline (approved):

tomponline (approved):

PR #234, opened 2023-11-28
Clear up network questions

### LINKED ISSUE #230: MicroCloud init fails
I have been playing with microcloud setup for a while now.
I have three arm64 boards all with ubuntu core 22 images
2 Xilinx KV260 and a Mediatek board.

While the setup works fine .  But quite a few times it fails during `microcloud init` giving the following error

**Error: System "cobuntu" failed to join the cluster: Failed to update cluster status of services: Failed to join "LXD" cluster: LXD has already been initialized**

On the respective node(in this case **cobuntu**) if you check logs of microcloud daemon it gives following error.
`
Sep 30 12:48:50 cobuntu microcloud.daemon[1811]: time="2023-09-30T12:48:50Z" level=error msg="Failed to start server" err="accept tcp [::]:9443: use of closed network connection"
`
 @tomponline explained during product roadmap sprint 
The reason for this can be 
two different SOCs being used as nodes for cluster formation and xilinx boards being short on memory and processing power.
I was just rasing it here since I never observed CPU usage or memory usage exceeding boards limit .

`
lxd 5.19-31ff7b6 26096 latest/stable canonical** -
microceph 0+git.7b5672b 710 quincy/stable canonical** -
microcloud 1.1-04a1c49 737 latest/stable canonical** -
microovn 22.03.3+snap2d1a04de44 301 22.03/stable canonical** -

`

masnax:
Closes #230  

It's a bit hard to be descriptive, leave out OVN jargon, and keep the questions concise, but here's something a bit nicer.

### DIFF
--- a/doc/tutorial/get_started.rst
+++ b/doc/tutorial/get_started.rst
@@ -384,7 +384,7 @@ See the full initialisation process here:
     Using 1 disk(s) on "micro3" for remote storage pool
 
    Configure distributed networking? (yes/no) [default=yes]:  yes
-   Select exactly one network interface from each cluster member:
+   Select an available interface per system to provide external connectivity for distributed network(s):
    Space to select; enter to confirm; type to filter results.
    Up/down to move; right to select all; left to select none.
           +----------+--------+----------+
@@ -402,8 +402,8 @@ See the full initialisation process here:
     Using "enp6s0" on "micro4" for OVN uplink
 
    Specify the IPv4 gateway (CIDR) on the uplink network (empty to skip IPv4): 192.0.2.1/24
-   Specify the first IPv4 address in the range to use with LXD: 192.0.2.100
-   Specify the last IPv4 address in the range to use with LXD: 192.0.2.254
+   Specify the first IPv4 address in the range to use on the uplink network: 192.0.2.100
+   Specify the last IPv4 address in the range to use on the uplink network: 192.0.2.254
    Specify the IPv6 gateway (CIDR) on the uplink network (empty to skip IPv6): 2001:db8:d:200::1/64
 
    Initializing a new cluster
--- a/microcloud/cmd/microcloud/ask.go
+++ b/microcloud/cmd/microcloud/ask.go
@@ -582,7 +582,7 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 	}
 
 	if missingSystems {
-		wantsSkip, err := c.asker.AskBool("Some systems are ineligible for distributed networking. Continue anyway? (yes/no) [default=yes]: ", "yes")
+		wantsSkip, err := c.asker.AskBool("Some systems are ineligible for distributed networking, which requires either an interface with no IPs assigned or a bridge. Continue anyway? (yes/no) [default=yes]: ", "yes")
 		if err != nil {
 			return err
 		}
@@ -594,7 +594,7 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 
 	// Uplink selection table.
 	header := []string{"LOCATION", "IFACE", "TYPE"}
-	fmt.Println("Select exactly one network interface from each cluster member:")
+	fmt.Println("Select an available interface per system to provide external connectivity for distributed network(s):")
 	data := [][]string{}
 	for peer, nets := range networks {
 		for _, net := range nets {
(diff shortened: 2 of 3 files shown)


masnax:
This is assuming we do want to avoid referencing OVN concepts directly in the setup. If that's not a concern then we can have slightly longer descriptions that specifically inform about what the interfaces are going to be used for.

microcloud/cmd/microcloud/ask.go
```diff
@@ -593,7 +593,7 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 
 	// Uplink selection table.
 	header := []string{"LOCATION", "IFACE", "TYPE"}
-	fmt.Println("Select exactly one network interface from each cluster member:")
+	fmt.Println("Select exactly one interface per system to dedicate for the distributed network:")
```
roosterfish, line 596: Maybe if we say "distributed overlay networks" it's even more clear that the actual virtual network traffic is meant? Just an idea :)

tomponline: We need to mention that this interface is for traffic going between the overlay network and the ddyetbal uplink network. Otherwise it'll be understood its for intra cluster overlay traffic.

tomponline: How about:

```
Select an available interface per system to provide external connectivity for distributed network(s)
```

roosterfish: The egress traffic is also a valid point. +1 for the last suggestion.

tomponline: Yeah, I used "external" as it could also be ingress too.

tomponline: And I used "connectivity" to imply Internet access, but not specifically mandate it.

microcloud/cmd/microcloud/ask.go
```diff
@@ -581,7 +581,7 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 	}
 
 	if missingSystems {
-		wantsSkip, err := c.asker.AskBool("Some systems are ineligible for distributed networking. Continue anyway? (yes/no) [default=yes]: ", "yes")
+		wantsSkip, err := c.asker.AskBool("Some systems are missing an interface with no IPs assigned, or a bridge. Continue anyway? (yes/no) [default=yes]: ", "yes")
```
ru-fu, line 584: I feel this is more ambiguous. It could be understood as "some systems are missing ..., or they are a bridge".
How about:
`Some systems are ineligible for distributed networking (they have neither an interface with no IPs assigned nor a bridge).`
It would be nicer to turn it around ("neither a bridge nor an interface with no IPs assigned"), but we probably want to put more emphasis on the interface with no IPs assigned?

microcloud/cmd/microcloud/ask.go
ru-fu: Should it be "network(s)"? The next questions mention "the distributed network".

ru-fu (commented):
Can you do a search and replace on the docs as well, to fix the places where we mention the init questions?

microcloud/cmd/microcloud/ask.go
tomponline: The user is selecting an interface for use by the distributed network(s) for external connectivity.
The init process is setting up one initial distributed network, but the user can add more in the future (thats one of the big benefits of OVN).

microcloud/cmd/microcloud/ask.go
masnax: The double negative might be confusing as well. 
How is: 
```
Some systems are ineligible for distributed networking: Requires (1) an interface with no IPs assigned or (2) a bridge. Continue anyway? (yes/no) [default=yes]
```

masnax:
Also added a commit that should clear up the error from #230

microcloud/cmd/microcloud/ask.go
ru-fu:
```
Some systems are ineligible for distributed networking, which requires either an interface with no IPs assigned or a bridge. Continue anyway? (yes/no) [default=yes]
```
I see why you're using the numbers to make the alternatives clear, but they would prompt me to type either 1 or 2 instead of "yes".

microcloud/cmd/microcloud/ask.go
```diff
@@ -581,7 +581,7 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 	}
 
 	if missingSystems {
-		wantsSkip, err := c.asker.AskBool("Some systems are ineligible for distributed networking. Continue anyway? (yes/no) [default=yes]: ", "yes")
+		wantsSkip, err := c.asker.AskBool("Some systems are ineligible for distributed networking: Requires (1) an interface with no IPs assigned or (2) a bridge. Continue anyway? (yes/no) [default=yes]: ", "yes")
```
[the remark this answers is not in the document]
masnax: Thanks!

masnax:
@ru-fu docs should now be updated as well.

ru-fu (dismissed):
Thank you!

doc/tutorial/get_started.rst
```diff
@@ -400,8 +400,8 @@ See the full initialisation process here:
     Using "enp6s0" on "micro4" for OVN uplink
 
    Specify the IPv4 gateway (CIDR) on the uplink network (empty to skip IPv4): 192.0.2.1/24
-   Specify the first IPv4 address in the range to use with LXD: 192.0.2.100
-   Specify the last IPv4 address in the range to use with LXD: 192.0.2.254
+   Specify the first IPv4 address in the range to use for the distributed network(s): 192.0.2.100
```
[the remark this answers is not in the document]
ru-fu: Only if we also change it in the code.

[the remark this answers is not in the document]
masnax: The previous question just says "specify the gateway (CIDR) on the uplink network" and this whole block starts with "Do you want to set up distributed networks?" so I think it might be less of a mouthful if we just say

`Specify the {first/last} {ipv4/ipv6} address in the range to use on the uplink network`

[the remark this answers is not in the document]
ru-fu: I actually find Tom's version clearer - yes, it's long and a bit repetitive, but for me who has no clue about networks, this makes it clear that okay, I'm now creating the distributed network, and I need to specify and address on the uplink network.

tomponline (approved):

PR #237, opened 2023-11-29
doc: clarify RAM requirements in the tutorial

ru-fu:
Make it clear that the tutorial uses less RAM than recommended, and that this should not be done in a production environment.


roosterfish (approved):
Thanks, LGTM!

tomponline (approved):

PR #242, opened 2023-12-06
doc/networking: clarify how networking works

ru-fu:
Make it clear that you need an uplink network and two dedicated network interfaces for MicroCloud.

### DIFF
--- a/doc/explanation/microcloud.rst
+++ b/doc/explanation/microcloud.rst
@@ -26,8 +26,33 @@ Networking
 
 By default, MicroCloud uses MicroOVN for networking, which is a minimal wrapper around OVN (Open Virtual Network).
 
-MicroOVN requires an uplink network that is an actual L2 subnet (which is usually not the case in a virtual cloud environment).
-In addition, MicroOVN requires its own dedicated network interface, for example, a dedicated physical network interface, a VLAN, or a virtual function on an :abbr:`SR-IOV (Single root I/O virtualisation)`-capable network interface.
+For external connectivity, MicroOVN requires an uplink network.
+This uplink network must support broadcast and multicast traffic (so that IP adverts and packets can flow between the OVN virtual router and the uplink network).
+Proper Ethernet networks generally fulfil these requirements, but virtual cloud environments often don't.
+
+Each machine in the MicroCloud cluster should have at least two available network interfaces (which can be connected to the same network or to different networks):
+
+.. _microcloud-networking-intracluster:
+
+Network interface for intra-cluster traffic
+  MicroCloud requires one network interface that is pre-configured with an IP address that is within the same subnet as the IPs of the other cluster members.
+  The network that it is connected to must support multicast.
+
+  This network interface can be, for example, a dedicated physical network interface, a VLAN, or a virtual function on an :abbr:`SR-IOV (Single root I/O virtualisation)`-capable network interface.
+  It serves as the dedicated network interface for MicroOVN and is used for multicast discovery (during setup) and all internal traffic between the MicroCloud, OVN, and Ceph members.
+
+.. _microcloud-networking-uplink:
+
+Network interface to connect to the uplink network
+  MicroCloud requires one network interface for connecting OVN to the uplink network.
+  This network interface must either be an unused interface that does not have an IP address configured, or a bridge.
+
+  MicroCloud configures this interface as an uplink interface that provides external connectivity to the MicroCloud cluster.
+
+  You can specify a different interface to be used as the uplink interface for each cluster member.
+  MicroCloud requires that all uplink interfaces are connected to the uplink network, using the gateway and IP address range information that you provide during the MicroCloud initialisation process.
+
+If you have a network interface that is configured as a Linux bridge, you can use it for both network interfaces.
 
 See :ref:`lxd:network-ovn` in the LXD documentation for more information.
 
--- a/doc/reference/index.rst
+++ b/doc/reference/index.rst
@@ -33,7 +33,8 @@ Also see Ceph's :ref:`ceph:hardware-recommendations`.
 Networking requirements
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-For networking, MicroCloud requires a dedicated network interface and an uplink network that is an actual L2 subnet.
+For networking, MicroCloud requires two dedicated network interfaces: one for intra-cluster communication and one for external connectivity.
+To allow for external connectivity, MicroCloud requires an uplink network that supports broadcast and multicast.
 See :ref:`explanation-networking` for more information.
 
 The IP addresses of the machines must not change after installation, so DHCP is not supported.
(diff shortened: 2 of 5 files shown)


doc/explanation/microcloud.rst
```diff
@@ -26,8 +26,30 @@ Networking
 
 By default, MicroCloud uses MicroOVN for networking, which is a minimal wrapper around OVN (Open Virtual Network).
 
-MicroOVN requires an uplink network that is an actual L2 subnet (which is usually not the case in a virtual cloud environment).
-In addition, MicroOVN requires its own dedicated network interface, for example, a dedicated physical network interface, a VLAN, or a virtual function on an :abbr:`SR-IOV (Single root I/O virtualisation)`-capable network interface.
+For outside connectivity, MicroOVN requires an uplink network.
+This uplink network must be an actual L2 subnet (which is usually not the case in a virtual cloud environment).
```
tomponline, line 30: "L2 subnet" is a contradiction in networking terms.

This is because "subnet" is an IP (internet protocol) concept, and IP is layer 3 (network layer).
L2 (or layer 2) indicates the layer below the network layer, which is the data link layer, most commonly Ethernet.

This link is useful to explain the different concepts:
https://www.cloudflare.com/en-gb/learning/network-layer/what-is-the-network-layer/

Although any mention of layers can become confusing, as that link describes, there are different definitions of layers.

I wonder if we shouldn't avoid mentioning layers at all, and instead be explicit as to what we require from the uplink network.

Namely we expect the uplink network interface to support broadcast and multicast traffic so that IP adverts from the OVN virtual router(s) are understood by the uplink network and can then ensure packets for the router make their way to the active chassis member.

tomponline: The reason we've said about layer 2 networking in the past is because a proper Ethernet network generally allows for ARP and NDP packets to be broadcast to all nodes on the network, including the uplink router, and this allows for the OVN virtual router(s) external IP on the uplink network to be migrated between LXD cluster members in the case that one LXD cluster member goes down and the others need to take over the role of providing external connectivity to the OVN network(s) by taking control of the virtual IP(s).

doc/tutorial/get_started.rst
```diff
@@ -131,7 +131,7 @@ Complete the following steps to create the required disks in a LXD storage pool:
 3. Create a network
 -------------------
 
-MicroCloud requires an uplink network that the cluster members can use to communicate.
+MicroCloud requires an uplink network that the cluster members can use for outside connectivity.
```
tomponline, line 134: We use the term "external" in the microcloud init process, shall we use the same term here too?

doc/explanation/microcloud.rst
ru-fu, line 30: Does this look better?

```suggestion
This uplink network must support broadcast and multicast traffic (so that IP adverts and packets from the OVN virtual router are transmitted and understood).
Proper Ethernet networks generally fulfil these requirements, but virtual cloud environments often don't.
```

doc/explanation/microcloud.rst
```diff
@@ -26,8 +26,30 @@ Networking
...
+
+Each machine in the MicroCloud cluster must have two available network interfaces:
+
+.. _microcloud-networking-intracluster:
+
+Network interface for intra-cluster traffic
+  MicroCloud requires one network interface that is configured with an IP address of a real layer 2 network.
+  All cluster members must be connected to the same network.
```
ru-fu, line 38: Is that correct?

```suggestion
  MicroCloud requires one network interface that is configured with an IP address.
  All cluster members must be connected to the same layer 2 network.
```

doc/explanation/microcloud.rst
tomponline: Yep sounds good thanks

tomponline: Maybe change "from" to "flow between the OVN virtual router and the uplink network"

doc/explanation/microcloud.rst
tomponline: How about this:

> MicroCloud requires one network interface be pre-configured with an IP within the same subnet as the other cluster members and that the network supports multicast. This is so that cluster member discovery can take place.

doc/explanation/microcloud.rst
```diff
@@ -26,8 +26,30 @@ Networking
 
 By default, MicroCloud uses MicroOVN for networking, which is a minimal wrapper around OVN (Open Virtual Network).
 
-MicroOVN requires an uplink network that is an actual L2 subnet (which is usually not the case in a virtual cloud environment).
-In addition, MicroOVN requires its own dedicated network interface, for example, a dedicated physical network interface, a VLAN, or a virtual function on an :abbr:`SR-IOV (Single root I/O virtualisation)`-capable network interface.
+For outside connectivity, MicroOVN requires an uplink network.
+This uplink network must be an actual L2 subnet (which is usually not the case in a virtual cloud environment).
+
+Each machine in the MicroCloud cluster must have two available network interfaces:
```
tomponline, line 32: How about this:

> Each machine in the MicroCloud cluster should have at least two network interfaces; one for intra-cluster traffic and one to provide external connectivity for the distributed network(s). These network interfaces can be connected to the same network. Alternatively a single network interface can be used if it is configured as a Linux bridge.

doc/explanation/microcloud.rst
```diff
@@ -26,8 +26,30 @@ Networking
...
+.. _microcloud-networking-uplink:
+
+Network interface to connect to the uplink network
+  MicroCloud requires one network interface for connecting OVN to the uplink network.
+  This network interface must either be an unused interface that does not have an IP address configured, or a bridge.
+
+  MicroCloud configures this interface as an uplink interface that provides outside connectivity to the MicroCloud cluster.
+
+  You can use different interfaces on each cluster member.
+  MicroCloud will connect all interfaces to the same uplink network, using the gateway and IP address range information that you provide during the MicroCloud initialisation process.
```
tomponline, line 52: I don't quite understand this bit:

> MicroCloud will connect all interfaces to the same uplink network...

What is "interfaces" here?

I think this should say something along the lines of:

> MicroCloud will configure the distributed OVN network(s) to use the same uplink network for external connectivity...

doc/reference/index.rst
```diff
@@ -33,7 +33,8 @@ Also see Ceph's :ref:`ceph:hardware-recommendations`.
 Networking requirements
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-For networking, MicroCloud requires a dedicated network interface and an uplink network that is an actual L2 subnet.
+For networking, MicroCloud requires two dedicated network interfaces: one for intra-cluster communication and one for outside connectivity.
```
tomponline, line 36: external rather than outside for consistency with microcloud init

doc/explanation/microcloud.rst
tomponline: OK how about this:

> You can specify a different interface to be used for the uplink for each cluster member.
> MicroCloud will requires that all interfaces are connected to the uplink network, using the gateway and IP address range information that you provide during the MicroCloud initialisation process.

ru-fu: Thanks, updated. :)

tomponline (approved):

roosterfish (approved):
Sorry for the late response, looking good!

Issue #239, opened 2023-12-04
feat: add option for specifying the network to use for OVN east-west (underlay) traffic


gabrielmougard:
JIRA ticket liink: https://warthogs.atlassian.net/jira/software/c/projects/LXD/boards/54?assignee=63d9eb6128cddcc70770aff1&selectedIssue=LXD-635

Add the option to specify the underlay network to be used. Right now, we only have the option to provide the details for the uplink, but the network used for the OVN clustering and thus the network over which the geneve tunnels are created is the same network used for the clustering of LXD and Ceph.

Allowing a user to specify interfaces to be used for an underlay network could serve the purpose of having dedicated VLANs for each purpose i.e. one VLAN for LXD clustering, one for Ceph access, one for Ceph replication and one for OVN.

## Interactive mode

After the OVN setup occurs, the user will see a prompt like:
```
Configure dedicated underlay networking? (yes/no) [default=no]:
```

If `yes`, the users will be asked to select one interface per detected machines (among the remaining interfaces not being used by the previous OVN setup) and will be prompted to enter a ipv4 subnet for the gateway with a ip range. If `no`, the setup is skipped.

## Preseed mode

We'll introduce the following YAML keys:

* `ovn_underlay_interface: <INTERFACE_NAME>` to characterise a node's network interface for the east-west traffic (in the same fashion as `ovn_uplink_interface` for the north-south traffic) 
* And the `ovn_underlay` field to describe the subnet used with the ip range (just like the existing `ovn` field does for the uplink connection)
```yaml
# `ovn_underlay` is optional and represents the private OVN east-west network configuration for LXD.
ovn_underlay:
  ipv4_gateway: 10.0.0.1/24
  ipv4_range: 10.0.0.100-10.0.0.254
  ipv6_gateway: cafe::1/64
```

PR #245, opened 2023-12-07
feat: specify an ovn underlay network 

gabrielmougard:
closes https://github.com/canonical/microcloud/issues/239
related to: https://github.com/canonical/microovn/pull/113
require: https://github.com/canonical/microcluster/pull/168

## WIP

* Add interactive + preseed tests (validation that a custom Geneve tunnel has been established can be done with `microovn.ovn-sbctl --columns=ip,type find Encap type=geneve`)

### DIFF
--- a/api/services.go
+++ b/api/services.go
@@ -125,7 +125,7 @@ func servicesPut(state *state.State, r *http.Request) response.Response {
 	services := make([]types.ServiceType, len(req.Tokens))
 	for i, cfg := range req.Tokens {
 		services[i] = types.ServiceType(cfg.Service)
-		joinConfigs[cfg.Service] = service.JoinConfig{Token: cfg.JoinToken, LXDConfig: req.LXDConfig, CephConfig: req.CephConfig}
+		joinConfigs[cfg.Service] = service.JoinConfig{Token: cfg.JoinToken, LXDConfig: req.LXDConfig, CephConfig: req.CephConfig, OVNConfig: req.OVNConfig}
 	}
 
 	// Default to the first iface if none specified.
--- a/cmd/microcloud/main_init.go
+++ b/cmd/microcloud/main_init.go
@@ -95,7 +95,7 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
 func (c *cmdInit) RunInteractive(cmd *cobra.Command, args []string) error {
 	// Initially restart LXD so that the correct MicroCloud service related state is set by the LXD snap.
 	fmt.Println("Waiting for LXD to start...")
-	lxdService, err := service.NewLXDService("", "", c.common.FlagMicroCloudDir)
+	lxdService, err := service.NewLXDService("", "", c.common.FlagMicroCloudDir, nil)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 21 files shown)


markylaing:
Blocked by https://github.com/canonical/microcluster/pull/94 and https://github.com/canonical/microcluster/pull/86

masnax:
@gabrielmougard Looks like there's some doc rebases to do before the tests will run.

api/services.go
```diff
@@ -98,7 +98,7 @@ func serviceTokensPost(s *state.State, r *http.Request) response.Response {
 		return response.BadRequest(err)
 	}
 
-	sh, err := service.NewHandler(s.Name(), req.ClusterAddress, s.OS.StateDir, false, false, types.ServiceType(serviceType))
+	sh, err := service.NewHandler(s.Name(), req.ClusterAddress, s.OS.StateDir, false, false, nil, types.ServiceType(serviceType))
```
masnax, line 101: You've added this config argument to `NewHandler` but it's always set to `nil` in every case. Is there a reason for this or can it be removed?

cmd/microcloud/main_init.go
```diff
@@ -491,6 +503,22 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
...
+	// If they don't have the `custom_encapsulation_ip` extension, output that the supplied custom encapsulation IPs
+	// were ignored.
+	if len(membersWithCustomEncapsulationIP) > 0 {
+		hasFeature, err := sh.Services[types.MicroOVN].(*service.OVNService).HasCustomEncapsulationIpFeature(context.Background())
+		if err != nil {
+			return err
+		}
+
+		if !hasFeature {
+			fmt.Println("Underlay network setup for MicroOVN was not applied because your MicroOVN memmbers do not support custom encapsulation IP for the Geneve tunnel, skipping")
```
masnax, line 517: Why are we checking this after joining nodes together? The list of API extensions is stored in memory during on first star, so why can't we check this earlier?

masnax (changes requested):
Looks good, just a couple questions

cmd/microcloud/main_init.go
gabrielmougard: Ah yes, true. I faced the issue of the DB not initialized at first (for the joining members), that's why I put the check at the end. But I didn't remember that it was also in memory.... let's do that then. Thanks!

api/services.go
gabrielmougard: Yes, good catch. I'm pretty sure I can remove this.

cmd/microcloud/main_init.go
gabrielmougard: @masnax Then, I'm wondering, do we need this function for checking the database if we can check it in memory before?

masnax: Maybe what would work best here is extending the `Service` interface with a new `SupportsFeature` function that checks the in-memory API extensions of the other snaps.


Since we're going to introduce enforced versioning of the snaps, I think for the purposes of this PR, it's safe to assume that if your local MicroOVN supports the feature, and you were actually able to find systems that are eligible to cluster with you, then you don't have to check those systems since they must already be at the same version as you.

Right now that validation isn't in place but it will be before the LTS so I think it's fine here to just check the local snaps.

api/services.go
gabrielmougard: fixed

cmd/microcloud/main_init.go
gabrielmougard: We will need this then https://github.com/canonical/microcluster/pull/168

gabrielmougard: @masnax updated :)

PR #248, opened 2023-12-12
fix: don't panic when IPv6 is not supported

### LINKED ISSUE #159: Add option to disable ipv6 protocol in microcloud init command.
**Problem description:**

**microcloud init** require IPv4 and IPv6 stack.
microcloud use calls _[mdns.DefaultParams(service)](https://github.com/canonical/microcloud/blob/da6d8a4962423d0d5516b3c329c584df7761c937/microcloud/mdns/lookup.go#L147)_ which instructs to use ipv4 and ipv6 sockets in msdn Query function call.
**microcloud init** fails with message **Error: Failed lookup: write udp6 [::]:<SRC_PORT>->[ff02::fb]:5353: sendto: cannot assign requested address** in case if ipv6 is disabled.

**Prerequisites**

- The single instance with the single network interface is enough for reproduction
- install microcloud via snap: _snap install microcloud_  (neverminded which channel is used, all are impacted)
- disable ipv6 stack (for example using shell command: _sudo sysctl -w net.ipv6.conf.all.disable_ipv6=1; sudo sysctl -w net.ipv6.conf.default.disable_ipv6=1_)

**Steps to reproduce:**

- run _sudo microcloud init --auto_
execution stops after **Scanning for eligible servers ...** step with the message **Error: Failed lookup: write udp6 [::]:<SRC_PORT>->[ff02::fb]:5353: sendto: cannot assign requested address**

**Recommendations to fix**
Add option to command disable ipv6 protocol in **microcloud init** command.

gabrielmougard:
closes #159 

- This fix enable the detection for a cluster member for IPv6 support to set the appropriate flag for `hashicorp/mdns` to avoid panicking if IPv6 is not supported,
- This also contains a fix for avoiding an infinite timeout issue in the `lookupPeers` function.
- However, the mDNS query (without support for IPv6) being received by the other cluster member mDNS servers, whose the responses are well written back to the UDPv4 connection opened on the server, are not received by the client and the reason still remains a mystery...

### DIFF
--- a/mdns/lookup.go
+++ b/mdns/lookup.go
@@ -148,7 +148,26 @@ func Lookup(ctx context.Context, iface *net.Interface, service string, size int)
 	params.Interface = iface
 	params.Entries = entriesCh
 	params.Timeout = 100 * time.Millisecond
-	err := mdns.Query(params)
+	ipv4Supported, ipv6Supported, err := checkIPStatus(iface.Name)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to check IP status: %w", err)
+	}
+
+	if !ipv4Supported {
+		logger.Info("IPv4 is not supported on this system network interface: disabling IPv4 mDNS", logger.Ctx{"iface": iface.Name})
+		params.DisableIPv4 = true
+	}
+
+	if !ipv6Supported {
+		logger.Info("IPv6 is not supported on this system network interface: disabling IPv6 mDNS", logger.Ctx{"iface": iface.Name})
+		params.DisableIPv6 = true
+	}
+
+	if params.DisableIPv4 && params.DisableIPv6 {
+		return nil, fmt.Errorf("No supported IP versions on the network interface %q", iface.Name)
+	}
+
+	err = mdns.Query(params)
 	if err != nil {
 		return nil, fmt.Errorf("Failed lookup: %w", err)
 	}
--- a/test/suites/basic.sh
+++ b/test/suites/basic.sh
@@ -47,6 +47,46 @@ test_interactive() {
     validate_system_lxd "${m}" 3 disk1
   done
 
+  # Reset the systems with just LXD and no IPv6 support.
+  reset_systems 3 3 1
+
+  for m in micro01 micro02 micro03 ; do
+    lxc exec "${m}" -- echo 1 > /proc/sys/net/ipv6/conf/all/disable_ipv6
+    lxc exec "${m}" -- snap disable microceph || true
+    lxc exec "${m}" -- snap disable microovn || true
+    lxc exec "${m}" -- snap restart microcloud
+  done
+
+  echo "Creating a MicroCloud with ZFS storage and no IPv6 support"
+  microcloud_interactive | lxc exec micro01 -- sh -c "microcloud init > out"
+
+  lxc exec micro01 -- tail -1 out | grep "MicroCloud is ready" -q
+  for m in micro01 micro02 micro03 ; do
+    validate_system_lxd "${m}" 3 disk1
+  done
+
+  # Reset the systems with just LXD and no IPv4 support.
+  gw_net_addr=$(lxc network get lxdbr0 ipv4.address)
+  lxc network set lxdbr0 ipv4.address none
+  reset_systems 3 3 1
+
+  for m in micro01 micro02 micro03 ; do
+    lxc exec "${m}" -- snap disable microceph || true
+    lxc exec "${m}" -- snap disable microovn || true
+    lxc exec "${m}" -- snap restart microcloud
+  done
+
+  export PROCEED_WITH_NO_OVERLAY_NETWORKING="no" # This will avoid to setup the cluster if no overlay networking is available.
+  echo "Creating a MicroCloud with ZFS storage and no IPv4 support"
+  ! microcloud_interactive | lxc exec micro01 -- sh -c "microcloud init 2> err" || false
+
+  # Ensure we error out due to a lack of usable overlay networking.
+  lxc exec micro01 -- cat err | grep "cluster bootstrapping aborted due to lack of usable networking" -q
+
+  # Set the IPv4 address back to the original value.
+  lxc network set lxdbr0 ipv4.address "${gw_net_addr}"
+  unset PROCEED_WITH_NO_OVERLAY_NETWORKING
+
   # Reset the systems and install microceph.
   reset_systems 3 3 1
 
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -625,6 +625,19 @@ func (c *CmdControl) askRemotePool(systems map[string]InitSystem, autoSetup bool
 	return nil
 }
 
+func (c *CmdControl) askProceedIfNoOverlayNetwork() error {
+	proceedWithNoOverlayNetworking, err := c.asker.AskBool("FAN networking is not usable. Do you want to proceed with setting up an inoperable cluster? (yes/no) [default=no]: ", "no")
+	if err != nil {
+		return err
+	}
+
+	if proceedWithNoOverlayNetworking {
+		return nil
+	}
+
+	return fmt.Errorf("cluster bootstrapping aborted due to lack of usable networking")
+}
+
 func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSystem, microCloudInternalSubnet *net.IPNet, autoSetup bool) error {
 	_, bootstrap := systems[sh.Name]
 	lxd := sh.Services[types.LXD].(*service.LXDService)
@@ -656,7 +669,7 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 	// Automatic setup gets a basic fan setup.
 	if autoSetup {
 		if !fanUsable {
-			fmt.Println("FAN networking is not usable, skipping")
+			return c.askProceedIfNoOverlayNetwork()
 		}
 
 		return nil
@@ -665,7 +678,7 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 	// Environments without OVN get a basic fan setup.
 	if sh.Services[types.MicroOVN] == nil {
 		if !fanUsable {
-			fmt.Println("FAN networking is not usable, skipping")
+			return c.askProceedIfNoOverlayNetwork()
 		}
 
 		return nil
@@ -787,7 +800,7 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 	if !canOVN {
 		fmt.Println("No dedicated uplink interfaces detected, skipping distributed networking")
 		if !fanUsable {
-			fmt.Println("FAN networking is not usable, skipping")
+			return c.askProceedIfNoOverlayNetwork()
 		}
 
 		return nil
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -628,29 +628,46 @@ func (c *CmdControl) askRemotePool(systems map[string]InitSystem, autoSetup bool
 func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSystem, microCloudInternalSubnet *net.IPNet, autoSetup bool) error {
 	_, bootstrap := systems[sh.Name]
 	lxd := sh.Services[types.LXD].(*service.LXDService)
-	for peer, system := range systems {
-		if bootstrap {
-			system.TargetNetworks = []api.NetworksPost{lxd.DefaultPendingFanNetwork()}
-			if peer == sh.Name {
-				network, err := lxd.DefaultFanNetwork()
-				if err != nil {
-					return err
+
+	// Check if FAN networking is usable.
+	fanUsable, _, err := lxd.FanNetworkUsable()
+	if err != nil {
+		return err
+	}
+
+	if fanUsable {
+		for peer, system := range systems {
+			if bootstrap {
+				system.TargetNetworks = []api.NetworksPost{lxd.DefaultPendingFanNetwork()}
+				if peer == sh.Name {
+					network, err := lxd.DefaultFanNetwork()
+					if err != nil {
+						return err
+					}
+
+					system.Networks = []api.NetworksPost{network}
 				}
-
-				system.Networks = []api.NetworksPost{network}
 			}
-		}
 
-		systems[peer] = system
+			systems[peer] = system
+		}
 	}
 
 	// Automatic setup gets a basic fan setup.
 	if autoSetup {
+		if !fanUsable {
+			fmt.Println("FAN networking is not usable, skipping")
+		}
+
 		return nil
 	}
 
 	// Environments without OVN get a basic fan setup.
 	if sh.Services[types.MicroOVN] == nil {
+		if !fanUsable {
+			fmt.Println("FAN networking is not usable, skipping")
+		}
+
 		return nil
 	}
 
@@ -769,6 +786,10 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 
 	if !canOVN {
 		fmt.Println("No dedicated uplink interfaces detected, skipping distributed networking")
+		if !fanUsable {
+			fmt.Println("FAN networking is not usable, skipping")
+		}
+
 		return nil
 	}
 
--- a/cmd/microcloud/main_init.go
+++ b/cmd/microcloud/main_init.go
@@ -200,14 +200,13 @@ func lookupPeers(s *service.Handler, autoSetup bool, iface *net.Interface, subne
 		}()
 	}
 
-	var timeAfter <-chan time.Time
+	timeoutDuration := time.Minute
 	if autoSetup {
-		timeAfter = time.After(5 * time.Second)
+		timeoutDuration = 5 * time.Second
 	}
 
-	if len(expectedSystems) > 0 {
-		timeAfter = time.After(1 * time.Minute)
-	}
+	ctx, cancel := context.WithTimeout(context.Background(), timeoutDuration)
+	defer cancel()
 
 	expectedSystemsMap := make(map[string]bool, len(expectedSystems))
 	for _, system := range expectedSystems {
@@ -219,7 +218,7 @@ func lookupPeers(s *service.Handler, autoSetup bool, iface *net.Interface, subne
 	done := false
 	for !done {
 		select {
-		case <-timeAfter:
+		case <-ctx.Done():
 			done = true
 		case err := <-selectionCh:
 			if err != nil {
@@ -235,7 +234,7 @@ func lookupPeers(s *service.Handler, autoSetup bool, iface *net.Interface, subne
 				break
 			}
 
-			peers, err := mdns.LookupPeers(context.Background(), iface, mdns.Version, s.Name)
+			peers, err := mdns.LookupPeers(ctx, iface, mdns.Version, s.Name)
 			if err != nil {
 				return err
 			}
--- a/cmd/microcloud/main_init_preseed.go
+++ b/cmd/microcloud/main_init_preseed.go
@@ -498,8 +498,14 @@ func (p *Preseed) Parse(s *service.Handler, bootstrap bool) (map[string]InitSyst
 		systems[peer] = system
 	}
 
+	// Check if FAN networking is usable.
+	fanUsable, _, err := lxd.FanNetworkUsable()
+	if err != nil {
+		return nil, err
+	}
+
 	// Setup FAN network if OVN not available.
-	if len(ifaceByPeer) == 0 {
+	if len(ifaceByPeer) == 0 && fanUsable {
 		for peer, system := range systems {
 			if bootstrap {
 				system.TargetNetworks = append(system.TargetNetworks, lxd.DefaultPendingFanNetwork())
--- a/mdns/mdns.go
+++ b/mdns/mdns.go
@@ -48,3 +48,39 @@ func dnsTXTSlice(list []byte) []string {
 
 	return parts
 }
+
+// checkIPStatus checks if the interface is up, has multicast, and supports IPv4/IPv6.
+func checkIPStatus(iface string) (ipv4OK bool, ipv6OK bool, err error) {
+	netInterface, err := net.InterfaceByName(iface)
+	if err != nil {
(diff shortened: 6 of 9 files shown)


microcloud/mdns/mdns.go
```diff
@@ -48,3 +50,18 @@ func dnsTXTSlice(list []byte) []string {
 
 	return parts
 }
+
+// CheckIPv6Status returns true if IPv6 is enabled, false otherwise.
+func CheckIPv6Status() (bool, error) {
+	allIPv6IfacesDisabled, err := ioutil.ReadFile("/proc/sys/net/ipv6/conf/all/disable_ipv6")
```
markylaing, line 56: If I recall correctly, the `ioutil` package is deprecated. The linter should have caught this. I'll add a ticket to investigate but in the meantime can you use `os.ReadFile` thanks.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -179,14 +179,12 @@ func lookupPeers(s *service.Handler, autoSetup bool, iface *net.Interface, subne
 		}()
 	}
 
-	var timeAfter <-chan time.Time
+	timeoutDuration := 1 * time.Minute
```
markylaing, line 182: I don't think we need to multiply by 1 :wink:

markylaing (changes requested):
Just a couple of small changes. Thanks.

microcloud/mdns/mdns.go
```diff
@@ -48,3 +50,18 @@ func dnsTXTSlice(list []byte) []string {
...
 }
+
+// CheckIPv6Status returns true if IPv6 is enabled, false otherwise.
+func CheckIPv6Status() (bool, error) {
+	allIPv6IfacesDisabled, err := os.ReadFile("/proc/sys/net/ipv6/conf/all/disable_ipv6")
+	if err != nil {
+		return false, err
+	}
+
+	newIPv6IfacesDisabled, err := os.ReadFile("/proc/sys/net/ipv6/conf/default/disable_ipv6")
```
masnax, line 61: MicroCloud may select a particular network interface, so I'm not sure if checking `all` or `default` is sufficient. You may need to also check the path for the particular network interface that will be used for the lookup.

microcloud/mdns/lookup.go
```diff
@@ -148,6 +148,10 @@ func Lookup(ctx context.Context, iface *net.Interface, service string, size int)
 	params.Interface = iface
 	params.Entries = entriesCh
 	params.Timeout = 100 * time.Millisecond
+	if !ipv6Supported {
+		params.DisableIPv6 = true
```
masnax, line 152: Would it be worthwhile to check the other way (ipv4) as well? If a system is lacking IPv4 addresses that are eligible for mDNS, would it still result in a panic?

microcloud/cmd/microcloud/main_init.go
```diff
@@ -179,20 +179,27 @@ func lookupPeers(s *service.Handler, autoSetup bool, iface *net.Interface, subne
...
 		expectedSystemsMap[system] = true
 	}
 
+	ipv6Supported, err := mdns.CheckIPv6Status()
+	if err != nil {
+		return fmt.Errorf("Could not check IPv6 status: %v", err)
+	}
+
+	if !ipv6Supported {
+		fmt.Println("IPv6 is not supported on this system, disabling IPv6 mDNS")
```
masnax, line 200: I'm not sure if it's absolutely necessary to include a CLI message message for this. At minimum it's adding another parameter that we have to keep track of. This might be a bit cleaner if `CheckIPv6Status()` is instead called within `LookupPeers` (or `Lookup`) and maybe set a warning or info log if IPv6 is disabled.

masnax (changes requested):
Looks good, just a couple questions/changes, thanks!

tomponline:
> * However, the mDNS query (without support for IPv6) being received by the other cluster member mDNS servers, whose the responses are well written back to the UDPv4 connection opened on the server, are not received by the client and the reason still remains a mystery...

Could you explain this one for me? Are you saying that if you have IPv6 disabled, this prevents the panic, but auto discovery doesn't work still?

gabrielmougard:
@tomponline As this unexpected behaviour seems to be quite different (discovery issue vs panicking) from the original issue title, should I create a new related ticket ?

tomponline:
@gabrielmougard also could you quote and reply to my questions rather than editing them in the future. Thanks

gabrielmougard:
@masnax do you have an idea on what could be going wrong here? I must admit that I'm running out of ideas...

masnax:
@gabrielmougard I'm not able to replicate the discovery issue with this branch myself.

I've tried setting `disable_ipv6` to 1 for every combination of `all`, `default`, and `enp5s0`, and I've also manually cleared every ipv6 address with `ip a del`. At least on my end, the systems are all still discoverable with ipv4. 

Can you provide some reproducer steps for the discoverability issue? 

Also, I just remembered that we will have to add permissions to the snap to allow accessing those `/proc/sys/net` paths as otherwise it's going to result in an apparmor denial.

masnax:
> @masnax should I open a new issue with a reproducer with this ipv4 discoverability issue I experience as this seems out of scope for this ticket (it is likely that my microcloud cluster is misconfigured somehow) so that we can close this one (avoiding the panicking issue) ?

It would be great if you could file an issue with reproducer steps. I think we ought to take a look at the issue further before we go ahead and merge this. If it ends up being unrelated then we can go ahead but we should investigate first.

microcloud/mdns/mdns.go
gabrielmougard: fixed

microcloud/cmd/microcloud/main_init.go
gabrielmougard: fixed

microcloud/mdns/mdns.go
gabrielmougard: fixed

microcloud/cmd/microcloud/main_init.go
gabrielmougard: fixed

microcloud/mdns/lookup.go
gabrielmougard: fixed

gabrielmougard:
@masnax I made it work! It's been a while since I experienced the last IPv4 issue (so I don't really remember the setup steps I took), so now after following the official microcloud doc and manually disabling IPv6, I was able to get mDNS to work on IPv4 :). I updated the `Lookup()` function though, just to return an error early if IPv6 AND IPv4 are both not supported.

microcloud/mdns/mdns.go
```diff
@@ -48,3 +50,44 @@ func dnsTXTSlice(list []byte) []string {
...
+
+// checkIPStatus checks if the interface is up, has multicast, and supports IPv4/IPv6.
+func checkIPStatus(iface string) (ipv4OK bool, ipv6OK bool, err error) {
+	netInterface, err := net.InterfaceByName(iface)
+	if err != nil {
+		return false, false, err
+	}
+
+	if netInterface.Flags&net.FlagUp != net.FlagUp {
+		fmt.Printf("Interface %q is down\n", iface)
```
tomponline, line 62: Why are you printing output here?

gabrielmougard: exactly. I removed the printfs, but should I keep them as `logger.Warnf` ?

microcloud/mdns/mdns.go
```diff
@@ -48,3 +50,42 @@ func dnsTXTSlice(list []byte) []string {
...
+		ifaceIPv6DisabledBytes, err := os.ReadFile(fmt.Sprintf("/proc/sys/net/ipv6/conf/%s/disable_ipv6", iface))
+		if err != nil {
+			return true, false, err
+		}
+
+		ifaceIPv6Disabled = strings.TrimSpace(string(ifaceIPv6DisabledBytes)) == "0"
+	}
+
+	ipv6OK = strings.TrimSpace(string(allIPv6IfacesDisabled)) == "0" && strings.TrimSpace(string(newIPv6IfacesDisabled)) == "0" && ifaceIPv6Disabled
+	return true, ipv6OK, nil
```
markylaing, line 90: The naming in this function is a bit confusing for me. For each of the checks can we have something like:

```golang
allIPv6InterfacesDisabledBytes, err := os.ReadFile("/proc/sys/net/ipv6/conf/all/disable_ipv6")
if err != nil {
    return true, false, err
}

// Check disabled is false.
allIPv6InterfacesEnabled = strings.TrimSpace(string(allIPv6InterfacesDisabledBytes)) == "0"
```

Then at the bottom we'd have:
```golang
ipv6OK = allIPv6InterfacesEnabled && newIPv6InterfacesEnabled && interfaceIPv6Enabled
```

microcloud/mdns/lookup.go
```diff
@@ -148,7 +148,26 @@ func Lookup(ctx context.Context, iface *net.Interface, service string, size int)
 	params.Interface = iface
 	params.Entries = entriesCh
 	params.Timeout = 100 * time.Millisecond
-	err := mdns.Query(params)
+	ipv4Supported, ipv6Supported, err := checkIPStatus(iface.Name)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to check IPv6 status: %w", err)
```
markylaing, line 153: Should probably have `return nil, fmt.Errorf("Failed to check IP status: %w", err)"` as the caller doesn't no the error cause here.

markylaing (dismissed):
Looks good, just a few minor tweaks :)

microcloud/cmd/microcloud/main_init.go
```diff
@@ -179,14 +179,12 @@ func lookupPeers(s *service.Handler, autoSetup bool, iface *net.Interface, subne
 		}()
 	}
 
-	var timeAfter <-chan time.Time
+	timeoutDuration := time.Minute
```
tomponline, line 182: Why dont you use a context.WithTimeout() here then you can pass it to  mdns.LookupPeers() too and potentially just wait for that to return.

microcloud/mdns/mdns.go
```diff
@@ -48,3 +50,42 @@ func dnsTXTSlice(list []byte) []string {
...
+	}
+
+	newIPv6IfacesDisabled, err := os.ReadFile("/proc/sys/net/ipv6/conf/default/disable_ipv6")
+	if err != nil {
+		return true, false, err
+	}
+
+	ifaceIPv6Disabled := false
+	if iface != "" {
+		ifaceIPv6DisabledBytes, err := os.ReadFile(fmt.Sprintf("/proc/sys/net/ipv6/conf/%s/disable_ipv6", iface))
```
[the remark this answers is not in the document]
gabrielmougard: fixed

microcloud/mdns/mdns.go
```diff
@@ -48,3 +50,42 @@ func dnsTXTSlice(list []byte) []string {
...
+		return false, false, nil
+	}
+
+	if netInterface.Flags&net.FlagMulticast != net.FlagMulticast {
+		return false, false, nil
+	}
+
+	allIPv6IfacesDisabled, err := os.ReadFile("/proc/sys/net/ipv6/conf/all/disable_ipv6")
+	if err != nil {
+		return true, false, err
```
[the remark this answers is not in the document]
gabrielmougard: fixed

microcloud/mdns/mdns.go
gabrielmougard: fixed

microcloud/mdns/lookup.go
gabrielmougard: fixed

microcloud/cmd/microcloud/main_init.go
gabrielmougard: fixed

gabrielmougard:
@markylaing @tomponline updated

masnax:
> snap

Looks like this is going to be an issue, judging from the tests:

```
Error: Failed to check IP status: open /proc/sys/net/ipv6/conf/all/disable_ipv6: permission denied
```

tomponline:
Can you infer this from ipv6 link local addresses on the interfaces instead?

masnax:
> Can you infer this from ipv6 link local addresses on the interfaces instead?

I tried testing it locally and so far checking for the link local address seems sufficient.

Given just 1 interface `enp5s0`, the panic seems to occur in these 3 cases:
* `echo 1 > /proc/sys/net/ipv6/conf/all/disable_ipv6`
* `echo 1 > /proc/sys/net/ipv6/conf/enp5s0/disable_ipv6`
* `ip a del fe80... dev enp5s0`

It doesn't happen when setting `/proc/sys/net/ipv6/conf/default/disable_ipv6` because that seems to only apply to newly created interfaces.

And in all 3 cases, the link local address is indeed missing from `ip a show`, so I think it would be sufficient to check for that.

gabrielmougard:
@markylaing @tomponline do you think we can merge this now or is there some last comments you would like to add?

masnax (changes requested):
@gabrielmougard Can you please add a system test case that verifies that MicroCloud does not panic if IPv4 or IPv6 is disabled?

masnax:
@gabrielmougard just a ping in case you forgot about this one, it still needs tests and a rebase, thanks :)

gabrielmougard:
@masnax I added a test with IPv6 deactivated. However, how can you disable IPv4 on a system (while keeping IPv6) ? I could not find a solution... @simondeziel do you know if it is  possible?

masnax:
> @masnax I added a test with IPv6 deactivated. However, how can you disable IPv4 on a system (while keeping IPv6) ? I could not find a solution... @simondeziel do you know if it is possible?

Well naively you could just `ip a del` for every ipv4 address on the lookup interface.

Or set `ipv4.address=none` in the LXD config before restoring the VM from snapshot.

gabrielmougard:
> > @masnax I added a test with IPv6 deactivated. However, how can you disable IPv4 on a system (while keeping IPv6) ? I could not find a solution... @simondeziel do you know if it is possible?
> 
> Well naively you could just `ip a del` for every ipv4 address on the lookup interface.
> 
> Or set `ipv4.address=none` in the LXD config before restoring the VM from snapshot.

@masnax @simondeziel I chose the `ipv4.address=none` . But in order to make it work, I had to update the `lxdfan0` creation logic (it was only working withh IPv4). However, in order to read IPv6 routes, I need to read `/proc/net/ipv6_route` which is restricted in read mode (unlike `/proc/net/route`). A potential workaround would be to enable the `network-observe` snap plug to give MicroCloud a priviledged read access to the system network configuration. What do you think?

simondeziel:
> in order to make it work, I had to update the `lxdfan0` creation logic (it was only working withh IPv4).

Wait, aren't Fan networks IPv4 only by design?

> However, in order to read IPv6 routes, I need to read `/proc/net/ipv6_route` which is restricted in read mode (unlike `/proc/net/route`).

That sounds like a snapd bug to me as both files are world readable and convey the same information.

gabrielmougard:
> > in order to make it work, I had to update the `lxdfan0` creation logic (it was only working withh IPv4).
> 
> Wait, aren't Fan networks IPv4 only by design?
> 
> > However, in order to read IPv6 routes, I need to read `/proc/net/ipv6_route` which is restricted in read mode (unlike `/proc/net/route`).
> 
> That sounds like a snapd bug to me as both files are world readable and convey the same information.

@simondeziel you're right. I juts checked the validation rule in LXD in regards to `fan.<overlay_subnet | underlay_subnet>`, and only IPv4 is accepted. I don't see much more explanations on why it does not work with IPv6 though..

gabrielmougard:
Then, in case IPv4 is not supported, shall we create a basic bridge network as `lxdfan0` (i.e, `bridge.mode=standard`). If so, we should give it an other name I think.

gabrielmougard:
> > in order to make it work, I had to update the `lxdfan0` creation logic (it was only working withh IPv4).
> 
> Wait, aren't Fan networks IPv4 only by design?
> 
> > However, in order to read IPv6 routes, I need to read `/proc/net/ipv6_route` which is restricted in read mode (unlike `/proc/net/route`).
> 
> That sounds like a snapd bug to me as both files are world readable and convey the same information.

@simondeziel do you know where I can create the issue ticket? I don't see the 'Issues' section https://github.com/snapcore/snapd

bboozzoo:
I commented on the bug report, but I'll comment here as well. Please add a plug for `network-observe`. When connected it should give you access to the relevant files (and more).

gabrielmougard:
> I commented on the bug report, but I'll comment here as well. Please add a plug for `network-observe`. When connected it should give you access to the relevant files (and more).

@bboozzoo thanks a lot!

tomponline:
> > > in order to make it work, I had to update the `lxdfan0` creation logic (it was only working withh IPv4).
> > 
> > 
> > Wait, aren't Fan networks IPv4 only by design?
> > > However, in order to read IPv6 routes, I need to read `/proc/net/ipv6_route` which is restricted in read mode (unlike `/proc/net/route`).
> > 
> > 
> > That sounds like a snapd bug to me as both files are world readable and convey the same information.
> 
> @simondeziel you're right. I juts checked the validation rule in LXD in regards to `fan.<overlay_subnet | underlay_subnet>`, and only IPv4 is accepted. I don't see much more explanations on why it does not work with IPv6 though..

Ubuntu FAN networks only support IPv4 because of the way they use an unallocated IPv4 range for the overlay mapping.

gabrielmougard:
@masnax @simondeziel tests should be good now :)

test/suites/basic.sh
```diff
@@ -41,6 +41,44 @@ test_interactive() {
...
+    lxc exec "${m}" -- snap disable microceph || true
+    lxc exec "${m}" -- snap disable microovn || true
+    lxc exec "${m}" -- snap restart microcloud
+  done
+
+  echo "Creating a MicroCloud with ZFS storage and no IPv4 support"
+  ! microcloud_interactive | lxc exec micro01 -- sh -c "microcloud init 2> out" || false
+
+  # Ensure we exited with the right error message.
+  lxc exec micro01 -- tail -1 err | grep "Error: Could not determine Fan overlay subnet: No default gateway for IPv4" -q
```
simondeziel, line 77:
```suggestion
  lxc exec micro01 -- tail -1 out | grep "Error: Could not determine Fan overlay subnet: No default gateway for IPv4" -q
```

My diagnosis: too much go error handling for you ;)

gabrielmougard: fixed

mdns/lookup.go
```diff
@@ -148,7 +148,26 @@ func Lookup(ctx context.Context, iface *net.Interface, service string, size int)
...
 	params.Entries = entriesCh
 	params.Timeout = 100 * time.Millisecond
-	err := mdns.Query(params)
+	ipv4Supported, ipv6Supported, err := checkIPStatus(iface.Name)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to check IP status: %w", err)
+	}
+
+	if !ipv4Supported {
+		logger.Infof("IPv4 is not supported on this system network interface %q: disabling IPv4 mDNS\n", iface.Name)
```
tomponline, line 157: We dont use logger.*f() in logging for LXD anymore (instead preferring contextual logging), is this still used in MicroCloud or is this incorrect? cc @roosterfish

simondeziel:
@tomponline I'll check with IS what's up with the DCO check. For this PR, I confirmed the 4 existing commits have the signed-off-by tag.

mdns/lookup.go
masnax: I wasn't aware of this paradigm. Is the reason just consistency?

I'll file an issue to bring MicroCloud & microcluster in line with LXD.

tomponline: Yeah we've been moving over to contextual logging for a few years now I think. 

You'll see in LXD mostly `logger.Info("message", logger.Ctx{"key":value})`

test/suites/basic.sh
```diff
@@ -41,6 +41,44 @@ test_interactive() {
...
+    lxc exec "${m}" -- snap disable microceph || true
+    lxc exec "${m}" -- snap disable microovn || true
+    lxc exec "${m}" -- snap restart microcloud
+  done
+
+  echo "Creating a MicroCloud with ZFS storage and no IPv4 support"
+  ! microcloud_interactive | lxc exec micro01 -- sh -c "microcloud init 2> out" || false
+
+  # Ensure we exited with the right error message.
+  lxc exec micro01 -- tail -1 out | grep "Error: Could not determine Fan overlay subnet: No default gateway for IPv4" -q
```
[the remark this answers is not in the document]
gabrielmougard: Ok, I can add a check to skip FAN networking. What about the case where OVN is also not usable? Should we error out here or allow the cluster creation to continue? For example:

```go
if !canOVN {
	fmt.Println("No dedicated uplink interfaces detected, skipping distributed networking")
	if !fanUsable {
		return fmt.Errorf("Neither FAN nor OVN networking is usable")
	}

	return nil
}
```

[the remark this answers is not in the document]
masnax: I don't think we need to error out necessarily. But since OVN takes precedence over the FAN network, we shouldn't log anything about the FAN network if we are attempting to set up OVN.

If the cluster size is >= 3 (supports OVN):
* If OVN is unavailable, log that we will skip it
* Then, attempt to fallback to FAN network
* Then, if FAN is unavailable, log that we will skip it

If the cluster size is < 3 (skips OVN entirely):
* If FAN is unavailable, log that we will skip it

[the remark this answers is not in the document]
masnax: It can still be set up after the fact. The alternative is that microcloud would not be able to support <3 node systems without ipv4.

mdns/lookup.go
gabrielmougard: fixed

gabrielmougard:
tests working now (except the usual timeouts in some interactive tests)

test/suites/basic.sh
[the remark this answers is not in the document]
tomponline: So from our meeting the other day we concluded that we would add a question for the end user to decide whether or not to proceed with setting up an inoperable cluster (from the point of view of no cluster wide overlay networking being available) in such scenarios.

simondeziel (approved):

masnax (approved):
There is 1 change here that needs to be made, particularly checking if FAN would be supported on _all_ systems and not just the local system before proceeding. 

But I can add that in the refactor PR, so this one can be merged as-is.

PR #259, opened 2024-01-26
Reuse existing MicroCeph and MicroOVN clusters

### LINKED ISSUE #145: Handle MicroCeph already installed
I have MicroCeph installed independently, before MicroCloud to use a dedicated network. When asked during the MicroCloud install if I want distributed storage, I say "no", but there seems to be an internal assumption preventing the setup to complete.

```
Would you like to set up distributed storage? (yes/no) [default=yes]: no

Configure distributed networking? (yes/no) [default=yes]: 
Select exactly one network interface from each cluster member:

 Using "enp6s0" on "n2" for OVN uplink
 Using "enp6s0" on "n1" for OVN uplink
 Using "enp6s0" on "n3" for OVN uplink

Specify the IPv4 gateway (CIDR) on the uplink network (empty to skip IPv4): 192.168.10.1/24
Specify the first IPv4 address in the range to use with LXD: 192.168.10.100
Specify the last IPv4 address in the range to use with LXD: 192.168.10.200
Specify the IPv6 gateway (CIDR) on the uplink network (empty to skip IPv6): 
Initializing a new cluster
 Local MicroCloud is ready
 Local LXD is ready
 Local MicroOVN is ready
Error: Failed to bootstrap local MicroCeph: Failed to initialize local remote entry: A remote with name "n1" already exists
```

masnax:
Closes #145

This PR allows initializing a new MicroCloud with some nodes that have already set up MicroCeph or MicroOVN in the past. This can be useful if you already have a MicroCeph cluster for example, that you want to "upgrade" into a MicroCloud.

Just after confirming the list of systems to use for the MicroCloud, we will try to grab a list of cluster members from each system. The underlying microcluster package will either return the list or report that the database is not initialized. We will record the first list that we obtain for each of MicroOVN and MicroCeph, and if the list is either non-existent or identical across any other systems, we can try to reuse those clusters.

The reuse process basically amounts to asking the system that has an existing cluster to generate join tokens and send them to the node orchestrating the initialization. Normally you have to be trusted to make such a request, so instead we can use the auth secret that we got from finding the systems over mDNS to send a request using the MicroCloud proxy, which will initiate the request from the unix socket on the clustered system.

The behaviour is as follows:
* If two systems belong to different clusters for the same service, we will error out as they wouldn't be able to join each other.
* If a cluster contains more nodes than what will end up in the MicroCloud, we will proceed as normal and ignore those nodes. Perhaps a user might want a larger MicroCeph cluster than their LXD cluster, for example.
* If `--auto` is specified, we will strictly require no clustered systems.
* The preseed file will have a new key `reuse_existing_clusters` which can be set to true or false. If true, we will reuse any clusters we find, and if false, we will skip that service entirely and set up as normal.
* In interactive mode, we will prompt the user for each service whether they want to reuse the existing cluster or setup MicroCloud without that service.

### DIFF
--- a/doc/how-to/initialise.rst
+++ b/doc/how-to/initialise.rst
@@ -90,6 +90,35 @@ Complete the following steps to initialise MicroCloud:
 
 See an example of the full initialisation process in the :ref:`Get started with MicroCloud <initialisation-process>` tutorial.
 
+Excluding MicroCeph or MicroOVN from MicroCloud
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If the MicroOVN or MicroCeph snap is not installed on the system that runs :command:`microcloud init`, you will be prompted with the following question::
+
+    MicroCeph not found. Continue anyway? (yes/no) [default=yes]:
+
+    MicroOVN not found. Continue anyway? (yes/no) [default=yes]:
+
+If you choose ``yes``,  only existing services will be configured on all systems.
+If you choose ``no``, the setup will be cancelled.
+
+All other systems must have at least the same set of snaps installed as the system that runs :command:`microcloud init`, otherwise they will not be available to select from the list of systems.
+Any questions associated to these systems will be skipped. For example, if MicroCeph is not installed, you will not be prompted for distributed storage configuration.
+
+Reusing an existing MicroCeph or MicroOVN with MicroCloud
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If some of the systems are already part of a MicroCeph or MicroOVN cluster, you can choose to reuse this cluster when initialising MicroCloud when prompted with the following question::
+
+    "micro01" is already part of a MicroCeph cluster. Do you want to add this cluster to MicroCloud? (add/skip) [default=add]:
+
+    "micro01" is already part of a MicroOVN cluster. Do you want to add this cluster to MicroCloud? (add/skip) [default=add]:
+
+If you choose ``add``, MicroCloud will add the remaining systems selected for initialisation to the pre-existing cluster.
+If you choose ``skip``, the respective service will not be set up at all.
+
+If more than one MicroCeph or MicroOVN cluster exists among the systems, the MicroCloud initialisation will be cancelled.
+
 .. _howto-initialise-preseed:
 
 Non-interactive configuration
--- a/microcloud/api/types/services.go
+++ b/microcloud/api/types/services.go
@@ -36,3 +36,9 @@ type ServiceToken struct {
 	Service   ServiceType `json:"service" yaml:"service"`
 	JoinToken string      `json:"join_token" yaml:"join_token"`
 }
+
+// ServiceTokensPost represents a request to issue a join token for a MicroCloud service.
+type ServiceTokensPost struct {
+	ClusterAddress string `json:"cluster_address" yaml:"cluster_address"`
+	JoinerName     string `json:"joiner_name"     yaml:"joiner_name"`
+}
--- a/microcloud/cmd/microcloud/ask.go
+++ b/microcloud/cmd/microcloud/ask.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 	cli "github.com/canonical/lxd/shared/cmd"
 	"github.com/canonical/lxd/shared/logger"
@@ -818,6 +819,47 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 		uplink, ovn := lxd.DefaultOVNNetwork(ipv4Gateway, ipv4Ranges, ipv6Gateway, dnsAddresses)
 		bootstrapSystem.Networks = []api.NetworksPost{uplink, ovn}
 		systems[sh.Name] = bootstrapSystem
+	}
+
+	return nil
+}
+
+// askClustered checks whether any of the selected systems have already initialized any expected services.
+// If a service is already initialized on some systems, we will offer to add the remaining systems, or skip that service.
+// If multiple systems have separately initialized the same service, we will abort initialization.
+// Preseed yamls will have a flag that sets whether to reuse the cluster.
+// In auto setup, we will expect no initialized services so that we can be opinionated about how we configure the cluster without user input.
+func (c *CmdControl) askClustered(s *service.Handler, autoSetup bool, systems map[string]InitSystem) error {
+	expectedServices := make(map[types.ServiceType]service.Service, len(s.Services))
+	for k, v := range s.Services {
+		expectedServices[k] = v
+	}
+
+	for serviceType := range expectedServices {
+		initializedSystem, _, err := checkClustered(s, autoSetup, serviceType, systems)
+		if err != nil {
+			return err
+		}
+
+		if initializedSystem != "" {
+			question := fmt.Sprintf("%q is already part of a %s cluster. Do you want to add this cluster to Microcloud? (add/skip) [default=add]", initializedSystem, serviceType)
+			validator := func(s string) error {
+				if !shared.ValueInSlice[string](s, []string{"add", "skip"}) {
+					return fmt.Errorf("Invalid input, expected one of (add,skip) but got %q", s)
+				}
+
+				return nil
+			}
+
+			addOrSkip, err := c.asker.AskString(question, "add", validator)
+			if err != nil {
+				return err
+			}
+
+			if addOrSkip != "add" {
+				delete(s.Services, serviceType)
+			}
+		}
 	}
 
 	return nil
--- a/microcloud/cmd/microcloud/main_init.go
+++ b/microcloud/cmd/microcloud/main_init.go
@@ -619,14 +619,40 @@ func checkClustered(s *service.Handler, autoSetup bool, serviceType types.Servic
 // setupCluster Bootstraps the cluster if necessary, adds all peers to the cluster, and completes any post cluster
 // configuration.
 func setupCluster(s *service.Handler, systems map[string]InitSystem) error {
+	initializedServices := map[types.ServiceType]string{}
 	_, bootstrap := systems[s.Name]
 	if bootstrap {
+		for serviceType := range s.Services {
+			for peer, system := range systems {
+				if system.InitializedServices[serviceType] != nil {
+					initializedServices[serviceType] = peer
+					break
+				}
+			}
+		}
+
 		fmt.Println("Initializing a new cluster")
+		mu := sync.Mutex{}
 		err := s.RunConcurrent(true, false, func(s service.Service) error {
+			// If there's already an initialized system for this service, we don't need to bootstrap it.
+			if initializedServices[s.Type()] != "" {
+				return nil
+			}
+
 			err := s.Bootstrap(context.Background())
 			if err != nil {
 				return fmt.Errorf("Failed to bootstrap local %s: %w", s.Type(), err)
 			}
+
+			mu.Lock()
+			clustered := systems[s.Name()]
+			if clustered.InitializedServices == nil {
+				clustered.InitializedServices = map[types.ServiceType]map[string]string{}
+			}
+
+			clustered.InitializedServices[s.Type()] = map[string]string{s.Name(): s.Address()}
+			systems[s.Name()] = clustered
+			mu.Unlock()
 
 			fmt.Printf(" Local %s is ready\n", s.Type())
 
@@ -635,28 +661,43 @@ func setupCluster(s *service.Handler, systems map[string]InitSystem) error {
 		if err != nil {
 			return err
 		}
-
-		// Only add disks for the local MicroCeph as other systems will add their disks upon joining.
-		var c *client.Client
-		for _, disk := range systems[s.Name].MicroCephDisks {
-			if c == nil {
-				c, err = s.Services[types.MicroCeph].(*service.CephService).Client("", "")
-				if err != nil {
-					return err
-				}
-			}
-
-			logger.Debug("Adding disk to MicroCeph", logger.Ctx{"peer": s.Name, "disk": disk.Path})
-			_, err = cephClient.AddDisk(context.Background(), c, &disk)
-			if err != nil {
-				return err
-			}
-		}
 	}
 
 	err := AddPeers(s, systems)
 	if err != nil {
 		return err
+	}
+
+	if bootstrap {
+		// Joiners will add their disks as part of the join process, so only add disks here for the system we bootstrapped, or already existed.
+		peer := s.Name
+		microCeph := initializedServices[types.MicroCeph]
+		if microCeph != "" {
+			peer = microCeph
+		}
+
+		for name := range systems[peer].InitializedServices[types.MicroCeph] {
+			// There may be existing cluster members that are not a part of MicroCloud, so ignore those.
+			if systems[name].ServerInfo.Name == "" {
+				continue
+			}
+
+			var c *client.Client
+			for _, disk := range systems[name].MicroCephDisks {
+				if c == nil {
+					c, err = s.Services[types.MicroCeph].(*service.CephService).Client(name, systems[name].ServerInfo.AuthSecret)
+					if err != nil {
+						return err
+					}
+				}
+
+				logger.Debug("Adding disk to MicroCeph", logger.Ctx{"name": name, "disk": disk.Path})
+				_, err = cephClient.AddDisk(context.Background(), c, &disk)
+				if err != nil {
+					return err
+				}
+			}
+		}
 	}
 
 	fmt.Println("Configuring cluster-wide devices ...")
--- a/microcloud/cmd/microcloud/main_init.go
+++ b/microcloud/cmd/microcloud/main_init.go
@@ -330,29 +330,12 @@ func lookupPeers(s *service.Handler, autoSetup bool, iface *net.Interface, subne
 	return nil
 }
 
-// waitForJoin issues a token and instructs a system to request a join,
+// waitForJoin requests a system to join each service's respective cluster,
 // and then waits for the request to either complete or time out.
 // If the request was successful, it additionally waits until the cluster appears in the database.
 func waitForJoin(sh *service.Handler, clusterSize int, secret string, peer string, cfg types.ServicesPut) error {
-	mut := sync.Mutex{}
-	err := sh.RunConcurrent(false, false, func(s service.Service) error {
-		token, err := s.IssueToken(context.Background(), peer)
-		if err != nil {
-			return fmt.Errorf("Failed to issue %s token for peer %q: %w", s.Type(), peer, err)
-		}
-
-		mut.Lock()
-		cfg.Tokens = append(cfg.Tokens, types.ServiceToken{Service: s.Type(), JoinToken: token})
-		mut.Unlock()
-
-		return nil
-	})
-	if err != nil {
-		return err
-	}
-
 	cloud := sh.Services[types.MicroCloud].(*service.CloudService)
-	err = cloud.RequestJoin(context.Background(), secret, peer, cfg)
+	err := cloud.RequestJoin(context.Background(), secret, peer, cfg)
 	if err != nil {
 		return fmt.Errorf("System %q failed to join the cluster: %w", peer, err)
 	}
@@ -407,6 +390,27 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
 	if err != nil {
 		return fmt.Errorf("Failed to inspect existing cluster: %w", err)
 	}
+	// Concurrently issue a token for each joiner.
+	for peer := range systems {
+		mut := sync.Mutex{}
+		err := sh.RunConcurrent(false, false, func(s service.Service) error {
+			token, err = s.IssueToken(context.Background(), peer)
+			if err != nil {
+				return fmt.Errorf("Failed to issue %s token for peer %q: %w", s.Type(), peer, err)
+			}
+
+			mut.Lock()
+			cfg := joinConfig[peer]
+			cfg.Tokens = append(cfg.Tokens, types.ServiceToken{Service: s.Type(), JoinToken: token})
+			joinConfig[peer] = cfg
+			mut.Unlock()
+
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+ 	}
 
 	clusterSize := len(cluster)
 
--- a/microcloud/cmd/microcloud/main_init.go
+++ b/microcloud/cmd/microcloud/main_init.go
@@ -333,7 +333,7 @@ func lookupPeers(s *service.Handler, autoSetup bool, iface *net.Interface, subne
 // waitForJoin requests a system to join each service's respective cluster,
 // and then waits for the request to either complete or time out.
 // If the request was successful, it additionally waits until the cluster appears in the database.
-func waitForJoin(sh *service.Handler, clusterSize int, secret string, peer string, cfg types.ServicesPut) error {
+func waitForJoin(sh *service.Handler, clusterSizes map[types.ServiceType]int, secret string, peer string, cfg types.ServicesPut) error {
 	cloud := sh.Services[types.MicroCloud].(*service.CloudService)
 	err := cloud.RequestJoin(context.Background(), secret, peer, cfg)
 	if err != nil {
@@ -341,24 +341,30 @@ func waitForJoin(sh *service.Handler, clusterSize int, secret string, peer strin
 	}
 
 	clustered := make(map[types.ServiceType]bool, len(sh.Services))
-	for service := range sh.Services {
-		clustered[service] = false
+	for _, tokenInfo := range cfg.Tokens {
+		clustered[tokenInfo.Service] = false
 	}
 
+	// Iterate over all services until the database is updated with the new node across all of them.
 	now := time.Now()
 	for len(clustered) != 0 {
 		if time.Since(now) >= time.Second*30 {
 			return fmt.Errorf("Timed out waiting for cluster member %q to appear", peer)
 		}
 
+		// Check the size of the cluster for each service.
 		for service := range clustered {
 			systems, err := sh.Services[service].ClusterMembers(context.Background())
 			if err != nil {
 				return err
 			}
 
-			if len(systems) == clusterSize+1 {
+			// If the size of the cluster has been incremented by 1 from its initial value,
+			// then we don't need to check the corresponding service anymore.
+			// So remove the service from consideration and update the current cluster size for the next node.
+			if len(systems) == clusterSizes[service]+1 {
 				delete(clustered, service)
+				clusterSizes[service] = clusterSizes[service] + 1
 			}
 		}
 	}
@@ -369,60 +375,116 @@ func waitForJoin(sh *service.Handler, clusterSize int, secret string, peer strin
 }
 
 func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
-	joinConfig := make(map[string]types.ServicesPut, len(systems))
-	secrets := make(map[string]string, len(systems))
-	for peer, info := range systems {
-		if peer == sh.Name {
-			continue
+	// Grab the systems that are clustered from the InitSystem map.
+	initializedServices := map[types.ServiceType]string{}
+	existingSystems := map[types.ServiceType]map[string]string{}
+	for serviceType := range sh.Services {
+		for peer, system := range systems {
+			if system.InitializedServices != nil && system.InitializedServices[serviceType] != nil {
+				initializedServices[serviceType] = peer
+				existingSystems[serviceType] = system.InitializedServices[serviceType]
+				break
+			}
 		}
+	}
 
+	// Prepare a JoinConfig to send to each joiner.
+	joinConfig := make(map[string]types.ServicesPut, len(systems))
+	for peer, info := range systems {
 		joinConfig[peer] = types.ServicesPut{
 			Tokens:     []types.ServiceToken{},
 			Address:    info.ServerInfo.Address,
 			LXDConfig:  info.JoinConfig,
 			CephConfig: info.MicroCephDisks,
 		}
-
-		secrets[peer] = info.ServerInfo.AuthSecret
 	}
 
-	cluster, err := sh.Services[types.MicroCloud].ClusterMembers(context.Background())
-	if err != nil {
-		return fmt.Errorf("Failed to inspect existing cluster: %w", err)
+	_, bootstrap := systems[sh.Name]
+	clusterSize := map[types.ServiceType]int{}
+	if bootstrap {
+		for serviceType, clusterMembers := range existingSystems {
+			clusterSize[serviceType] = len(clusterMembers)
+		}
 	}
+
 	// Concurrently issue a token for each joiner.
 	for peer := range systems {
 		mut := sync.Mutex{}
 		err := sh.RunConcurrent(false, false, func(s service.Service) error {
-			token, err = s.IssueToken(context.Background(), peer)
-			if err != nil {
-				return fmt.Errorf("Failed to issue %s token for peer %q: %w", s.Type(), peer, err)
-			}
+			// Only issue a token if the system isn't already part of that cluster.
+			if existingSystems[s.Type()][peer] == "" {
+				clusteredSystem := systems[initializedServices[s.Type()]]
 
-			mut.Lock()
-			cfg := joinConfig[peer]
-			cfg.Tokens = append(cfg.Tokens, types.ServiceToken{Service: s.Type(), JoinToken: token})
-			joinConfig[peer] = cfg
-			mut.Unlock()
+				var token string
+				var err error
+
+				// If the local node is part of the pre-existing cluster, or if we are growing the cluster, issue the token locally.
+				// Otherwise, use the MicroCloud proxy to ask an existing cluster member to issue the token.
+				if clusteredSystem.ServerInfo.Name == sh.Name || clusteredSystem.ServerInfo.Name == "" {
+					token, err = s.IssueToken(context.Background(), peer)
+					if err != nil {
+						return fmt.Errorf("Failed to issue %s token for peer %q: %w", s.Type(), peer, err)
+					}
+				} else {
+					cloud := sh.Services[types.MicroCloud].(*service.CloudService)
+					token, err = cloud.RemoteIssueToken(context.Background(), clusteredSystem.ServerInfo.Address, clusteredSystem.ServerInfo.AuthSecret, peer, s.Type())
+					if err != nil {
+						return err
+					}
+				}
+
+				// Fetch the current cluster sizes if we are adding a new node.
+				var currentCluster map[string]string
+				if !bootstrap {
+					currentCluster, err = s.ClusterMembers(context.Background())
+					if err != nil {
+						return fmt.Errorf("Failed to check for existing %s cluster size: %w", s.Type(), err)
+					}
+				}
+
+				mut.Lock()
+
+				if !bootstrap {
+					clusterSize[s.Type()] = len(currentCluster)
+				}
+
+				cfg := joinConfig[peer]
+				cfg.Tokens = append(cfg.Tokens, types.ServiceToken{Service: s.Type(), JoinToken: token})
+				joinConfig[peer] = cfg
+				mut.Unlock()
+			}
 
 			return nil
 		})
 		if err != nil {
 			return err
 		}
- 	}
-
-	clusterSize := len(cluster)
+	}
 
 	fmt.Println("Awaiting cluster formation ...")
+
+	// If the local node needs to join an existing cluster, do it first so we can proceed as normal.
+	if len(joinConfig[sh.Name].Tokens) > 0 {
+		cfg := joinConfig[sh.Name]
+		err := waitForJoin(sh, clusterSize, "", sh.Name, cfg)
+		if err != nil {
+			return err
+		}
+
+		// Sleep 3 seconds to give the cluster roles time to reshuffle before adding more members.
+		time.Sleep(3 * time.Second)
+	}
+
 	for peer, cfg := range joinConfig {
+		if len(cfg.Tokens) == 0 || peer == sh.Name {
+			continue
+		}
+
 		logger.Debug("Initiating sequential request for cluster join", logger.Ctx{"peer": peer})
 		err := waitForJoin(sh, clusterSize, systems[peer].ServerInfo.AuthSecret, peer, cfg)
 		if err != nil {
 			return err
 		}
-
-		clusterSize = clusterSize + 1
 
 		// Sleep 3 seconds to give the cluster roles time to reshuffle before adding more members.
 		time.Sleep(3 * time.Second)
--- a/microcloud/cmd/microcloud/main_init.go
+++ b/microcloud/cmd/microcloud/main_init.go
@@ -30,6 +30,8 @@ import (
 type InitSystem struct {
 	ServerInfo mdns.ServerInfo // Data reported by mDNS about this system.
 
+	InitializedServices map[types.ServiceType]map[string]string // A map of services and their cluster members, if initialized.
+
 	AvailableDisks []lxdAPI.ResourcesStorageDisk // Disks as reported by LXD.
 
 	MicroCephDisks     []cephTypes.DisksPost                  // Disks intended to be passed to MicroCeph.
@@ -128,6 +130,11 @@ func (c *cmdInit) RunInteractive(cmd *cobra.Command, args []string) error {
 	}
 
 	err = lookupPeers(s, c.flagAutoSetup, iface, subnet, nil, systems)
+	if err != nil {
+		return err
+	}
+
+	err = c.common.askClustered(s, c.flagAutoSetup, systems)
 	if err != nil {
 		return err
 	}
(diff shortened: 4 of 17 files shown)


microcloud/service/microcloud.go
```diff
@@ -95,6 +97,26 @@
...
 
+// RemoteIssueToken issues a token for the given peer on a remote MicroCloud where we are authorized by mDNS.
+func (s CloudService) RemoteIssueToken(ctx context.Context, clusterAddress string, secret string, peer string, serviceType types.ServiceType) (string, error) {
+	c, err := s.client.RemoteClient(util.CanonicalNetworkAddress(clusterAddress, CloudPort))
+	if err != nil {
+		return "", err
+	}
+
+	c.Client.Client.Transport = &http.Transport{
+		TLSClientConfig:   &tls.Config{InsecureSkipVerify: true},
```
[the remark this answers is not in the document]
masnax: We use `InsecureSkipVerify` in MicroCloud so that we can instead authenticate with the mDNS secret prior to setting up the cluster.

masnax:
@markylaing @roosterfish @tomponline 

This one should be ready for review now.

tomponline:
Thanks @masnax 

Once @markylaing and @roosterfish have approved ill do  a final pass. Ta

microcloud/cmd/microcloud/main_init.go
```diff
@@ -368,8 +351,9 @@ func waitForJoin(sh *service.Handler, clusterSize int, secret string, peer strin
 				return err
 			}
 
-			if len(systems) == clusterSize+1 {
+			if len(systems) == clusterSize[service]+1 {
 				delete(clustered, service)
+				clusterSize[service] = clusterSize[service] + 1
```
roosterfish, line 356: I fear I don't quite get what this is doing. Could you please add a comment?

roosterfish (changes requested):
Looks good to me, only a few smaller suggestions.

microcloud/service/microceph.go
```diff
@@ -131,10 +133,33 @@ func (s CephService) Join(ctx context.Context, joinConfig JoinConfig) error {
...
+func (s CephService) ClusterMembers(ctx context.Context, info *mdns.ServerInfo) (map[string]string, error) {
+	var client *client.Client
+	var err error
+	if info != nil && info.Name != s.name {
+		client, err = s.m.RemoteClient(util.CanonicalNetworkAddress(info.Address, CloudPort))
+		if err != nil {
+			return nil, err
+		}
+
+		client.Client.Client.Transport = &http.Transport{
```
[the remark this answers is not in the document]
masnax: I'll have a PR coming soon that gets rid of these entirely, thanks to https://github.com/canonical/microcluster/pull/83

To keep this PR simple I've maintained the pre-existing behaviour. There's actually a few significant refactors that need to be made to the client generation in MicroCloud because we inefficiently open clients all over the place. I've been thinking of maintaining a map of clients in memory so that we don't need to constantly instantiate new ones throughout the init process.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -414,6 +421,76 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
...
 }
 
+// checkClustered checks whether any of the selected systems have already initialized a service.
+func checkClustered(s *service.Handler, autoSetup bool, serviceType types.ServiceType, systems map[string]InitSystem) (string, map[string]string, error) {
+	var firstInitializedSystem string
+	var existingMembers map[string]string
+
+	// We don't allow appending to an existing LXD cluster.
+	if serviceType == types.LXD {
+		return "", nil, nil
```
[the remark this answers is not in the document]
masnax: When MicroCloud first starts, it already checks whether LXD is initialized before it restarts LXD [here](https://github.com/canonical/microcloud/blob/81d501c885e4fd559842b2654714672f667d55d0/microcloud/cmd/microcloud/main_init.go#L89) to ensure the correct symlinks to MicroOVN and MicroCeph are present. We do this so we don't accidentally restart LXD while it's running instances for example.

This condition is just a convenience that lets us exit early from the `checkClustered` function in the case of LXD because it will always be uninitialized if we got to this point. 

I thought it might keep the invocation simple since we can just iterate over every `serviceType` and call `checkClustered` for all of them and get a sane value. The alternative would be that we need to remember to always call `checkClustered` in an if block where `serviceType != types.LXD`.

microcloud/service/microceph.go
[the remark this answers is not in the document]
roosterfish: Ok, thanks for the details and the link. Fine for me then.

microcloud/cmd/microcloud/main_init.go
masnax: Sure, I'll add a comment too, but it's basically just ensuring that the after a node joins a service (the call to `cloud.RequestJoin`), that we actually wait until the node is present in the corresponding daemon's database by checking that the cluster size has actually been incremented. 

We repeat this check until the cluster member appears (the number of rows in the table increases), at which point we can delete the service from the `clustered` list so we don't need to check it in the next loop, and we update the expected size of the cluster so that the next node that joins is checked against the updated cluster size.

microcloud/cmd/microcloud/main_init.go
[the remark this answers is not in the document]
roosterfish: >  because it will always be uninitialized if we got to this point.

I was not thinking about this, thanks, clear now! I think it would be helpful to extend the comment with the reason why an empty string is returned here.

microcloud/cmd/microcloud/main_init.go
masnax: It might be helpful here to remember that `RequestJoin` will attempt join 1 node to all services concurrently, so we iterate over all services until the cluster_members table reflects the change. 

This is so we don't get into a weird situation where dqlite somehow mishandles 2 nodes concurrently joining.

microcloud/cmd/microcloud/main_init.go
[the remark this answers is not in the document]
masnax: Updated the comment and labelled the args so it should be a bit clearer now. Don't know why the preview is still showing the old version.

microcloud/cmd/microcloud/main_init.go
masnax: Added a commit which comments all the behaviours in this block.

roosterfish: I see so `len(systems)` is incremented by 1 as soon as the node has joined for the respective service.
And since `clusterSize` is a map you increment the reference here so that follow up calls to `waitForJoin` use the updated value.

masnax:
@markylaing @tomponline Do either of you want to take a look at this before merging?

Also @ru-fu Looks like there's something going on with the doc checks here. Got any ideas?

markylaing:
> @markylaing @tomponline Do either of you want to take a look at this before merging?
> 
> Also @ru-fu Looks like there's something going on with the doc checks here. Got any ideas?

Yes I'd like to take a look. Will get to it first thing tomorrow.

ru-fu:
This sounds like it needs doc updates.

markylaing:
> * The preseed file will have a new key `reuse_existing_clusters` which can be set to true or false. If true, we will reuse any clusters we find, and if false, we will skip that service entirely and set up as normal.

If this key is false, does it error out if we find existing Ceph or OVN clusters as it does with the `--auto` flag? I'm presuming here that we can't set up a new Ceph or OVN cluster on top of one that pre-exists.

microcloud/service/microceph.go
```diff
@@ -131,10 +133,33 @@ func (s CephService) Join(ctx context.Context, joinConfig JoinConfig) error {
 }
 
 // ClusterMembers returns a map of cluster member names and addresses.
-func (s CephService) ClusterMembers(ctx context.Context) (map[string]string, error) {
-	client, err := s.Client("", "")
-	if err != nil {
-		return nil, err
+// Optionally sends a remote request using the information in ServerInfo.
+func (s CephService) ClusterMembers(ctx context.Context, info *mdns.ServerInfo) (map[string]string, error) {
```
markylaing, line 137: Same here, I think it would be clearer to have a separate function when we want to use the remote client.

microcloud/service/microcloud.go
```diff
@@ -95,6 +95,26 @@ func (s CloudService) IssueToken(ctx context.Context, peer string) (string, erro
...
+		DisableKeepAlives: true,
+		Proxy: func(r *http.Request) (*url.URL, error) {
+			r.Header.Set("X-MicroCloud-Auth", secret)
+
+			return shared.ProxyFromEnvironment(r)
+		},
+	}
```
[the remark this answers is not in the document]
markylaing: I've just seen this https://github.com/canonical/microcloud/pull/259/commits/b9479a9645ac45ec5634991810561fa73ade4ec1#r1559631449 so ignore this comment :)

microcloud/cmd/microcloud/ask.go
```diff
@@ -816,3 +817,44 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
...
+	}
+
+	for serviceType := range expectedServices {
+		initializedSystem, err := checkClustered(s, autoSetup, serviceType, systems)
+		if err != nil {
+			return err
+		}
+
+		if initializedSystem != "" {
+			question := fmt.Sprintf("%q is already part of a %s cluster. Use this cluster to MicroCloud, or skip %s? (reuse/skip) [default=reuse]", initializedSystem, serviceType, serviceType)
```
markylaing: The wording here is a bit off. Perhaps:
```golang
fmt.Sprintf("%q is already part of a %s cluster. Do you want to add this cluster to Microcloud? (add/skip) [default=add]", initializedSystem, serviceType)
```

Also, do we need to capitalise or otherwise format the `serviceType` string?

microcloud/cmd/microcloud/ask.go
```diff
@@ -830,13 +830,13 @@ func (c *CmdControl) askClustered(s *service.Handler, autoSetup bool, systems ma
...
 	for serviceType := range expectedServices {
-		initializedSystem, err := checkClustered(s, autoSetup, serviceType, systems)
+		initializedSystem, _, err := checkClustered(s, autoSetup, serviceType, systems)
 		if err != nil {
 			return err
 		}
 
 		if initializedSystem != "" {
-			question := fmt.Sprintf("%q is already part of a %s cluster. Use this cluster to MicroCloud, or skip %s? (reuse/skip) [default=reuse]", initializedSystem, serviceType, serviceType)
+			question := fmt.Sprintf("%q is already part of a %s cluster. Use this cluster with MicroCloud, or skip %s? (reuse/skip) [default=reuse]: ", initializedSystem, serviceType, serviceType)
```
markylaing, line 839: Should these changes be in the previous commit?

microcloud/cmd/microcloud/main_init.go
```diff
@@ -327,33 +327,16 @@ func lookupPeers(s *service.Handler, autoSetup bool, iface *net.Interface, subne
...
-		cfg.Tokens = append(cfg.Tokens, types.ServiceToken{Service: s.Type(), JoinToken: token})
-		mut.Unlock()
-
-		return nil
-	})
-	if err != nil {
-		return err
-	}
-
+func waitForJoin(sh *service.Handler, clusterSize map[types.ServiceType]int, secret string, peer string, cfg types.ServicesPut) error {
```
markylaing, line 330: Three functions have been changed in this commit, could you please add a commit for each change with a comment describing the reason for the change. E.g. This function no longer issues a token, why was it moved to `AddPeers`?

microcloud/service/microcloud.go
```diff
@@ -127,19 +127,28 @@ func (s CloudService) RequestJoin(ctx context.Context, secret string, name strin
...
+			TLSClientConfig:   &tls.Config{InsecureSkipVerify: true},
+			DisableKeepAlives: true,
+			Proxy: func(r *http.Request) (*url.URL, error) {
+				r.Header.Set("X-MicroCloud-Auth", secret)
 
-			return shared.ProxyFromEnvironment(r)
-		},
+				return shared.ProxyFromEnvironment(r)
+			},
+		}
```
markylaing, line 151: Now that https://github.com/canonical/microcluster/pull/83 is merged does this need to be updated? Same for the other cases where this is being set.

microcloud/cmd/microcloud/main_init_preseed.go
```diff
@@ -391,6 +392,22 @@ func (p *Preseed) Parse(s *service.Handler, bootstrap bool) (map[string]InitSyst
...
+
+	for serviceType := range expectedServices {
+		initializedSystem, _, err := checkClustered(s, false, serviceType, systems)
+		if err != nil {
+			return nil, err
+		}
+
+		if initializedSystem != "" && !p.ReuseExistingClusters {
+			delete(s.Services, serviceType)
+		}
```
markylaing, line 408: I'm not sure if we should error out here or not. I wouldn't expect ceph to be running on a LXD cluster member unless it was being used in the that LXD cluster, so for me this would be indicative of misconfiguration.

microcloud/test/includes/microcloud.sh
```diff
@@ -287,7 +287,7 @@ validate_system_lxd_ovn() {
 
   # Check that the created UPLINK network has the right DNS servers.
   if [ -n "${dns_namesersers}" ] ; then
-    dns_addresses=$(lxc exec "${name}" -- sh -c "lxc network get UPLINK dns.nameservers")
+    dns_addresses=$(lxc exec "local:${name}" -- sh -c "lxc network get UPLINK dns.nameservers")
```
markylaing, line 290: Is this commit specific to this PR or can it be moved into a different one?

microcloud/cmd/microcloud/main_init.go
```diff
@@ -339,18 +339,23 @@ func waitForJoin(sh *service.Handler, clusterSize map[types.ServiceType]int, sec
...
+		// Check the size of the cluster for each service.
 		for service := range clustered {
 			systems, err := sh.Services[service].ClusterMembers(context.Background(), nil)
 			if err != nil {
 				return err
 			}
 
+			// If the size of the cluster has been incremented by 1 from its initial value,
+			// then we don't need to check the corresponding service anymore.
+			// So remove the service from consideration and update the current cluster size for the next node.
```
markylaing, line 358: Perhaps this should be in the previous commit which edited the `waitForJoin` function?

markylaing (changes requested):
I think the structure of it looks good but some parts could be made a bit clearer.

Also, one concern I have with this is the reliance on mDNS for issuing the tokens. Will it be possible to add existing services if the join process changes such that explicit verification is required on both ends? I believe @roosterfish is working on this currently so I think it's worth a discussion around that.

tomponline:
> @markylaing @tomponline Do either of you want to take a look at this before merging?

I'll look over once approved by the others.

roosterfish:
> Also, one concern I have with this is the reliance on mDNS for issuing the tokens. Will it be possible to add existing services if the join process changes such that explicit verification is required on both ends? 

Good point, I have also thought about this the other day when reviewing the PR. Essentially what we will probably not have anymore is the secret used for the `X-MicroCloud-Auth` header to talk to the other MicroCloud daemon. 
So it has to be an action invoked by the administrator on the target system. 
I have it on my list of things for the spec so we can discuss it there :)

tomponline:
> Also, one concern I have with this is the reliance on mDNS for issuing the tokens. Will it be possible to add existing services if the join process changes such that explicit verification is required on both ends? I believe @roosterfish is working on this currently so I think it's worth a discussion around that.

It will almost certainly have to change as mdns isn't going to be used as much. But that shouldn't necessarily block this PR - although it will likely have to be reworked as part of @roosterfish planned changes.

masnax:
> Also, one concern I have with this is the reliance on mDNS for issuing the tokens. Will it be possible to add existing services if the join process changes such that explicit verification is required on both ends? I believe @roosterfish is working on this currently so I think it's worth a discussion around that.

At the moment, pretty much everything MicroCloud does requires mDNS to work because we can't be trusted on the other systems until we've clustered with them. We need mDNS for selecting disks, selecting networks, and selecting the nodes themselves, and now issuing tokens from existing clusters with this PR. 

I'd assume the verification process would establish a long-lived trust for the duration of the join process similar to the mDNS auth secret, allowing us to remotely issue tokens, view available disks and network interfaces, and request a node to join the clusters.

masnax:
> > I'd assume the verification process would establish a long-lived trust for the duration of the join process similar to the mDNS auth secret, allowing us to remotely issue tokens, view available disks and network interfaces, and request a node to join the clusters.
> 
> An option I am trying to validate for the spec is moving the forming of the MicroCloud MicroCluster right after the discovery of new peers. This establishes trust between all the peers and allow them talking to each other using mutual TLS to perform the actions we are currently performing using the auth secret.
> 
> With this approach we don't anymore require any type of auth secret besides a join token for each of the peers that want to join the MicroCloud.

For some historical context, we actually had a similar process in place very early on because I was hesitant to rely on the mDNS component so much. 

Problem was mainly user experience. 90% of errors will happen around the configuration so a mistake would be costly as we would have to tear down the clusters. Plus it puts a significant pause (could be minutes for larger clusters) before we even get to the bulk of the user interactions.

microcloud/cmd/microcloud/main_init_preseed.go
masnax: It's tough since MicroCeph/MicroOVN might have other applications besides LXD. 

If we error out, that's like saying "you can't have a MicroCloud with these systems until you fundamentally change your current MicroCeph cluster". If we don't error out, it's "You can still have a MicroCloud, but it won't include MicroCeph until you fundamentally change your current MicroCeph cluster".

Since MicroCeph and MicroOVN are optional for MicroCloud, I think the second one makes sense and the first is too restrictive. PR #260 which builds on this PR will allow for adding MicroCeph/MicroOVN to MicroCloud if it was optionally/implicitly skipped before.

roosterfish:
> 90% of errors will happen around the configuration so a mistake would be costly as we would have to tear down the clusters. Plus it puts a significant pause (could be minutes for larger clusters) before we even get to the bulk of the user interactions.

What do you mean by we have to tear down the clusters? 

If I understood you right you are talking about the pause which will be an effect of joining in the cluster members manually? I guess this is something we have to accept as part of making the overall concept more secure. But it's worth evaluating options to make this as fast as possible for the end user.

masnax:
> What do you mean by we have to tear down the clusters?
The user runs `microcloud init`, but mistakenly hasn't prepared disks/network interfaces appropriately yet. If the systems are clustered before we ask the storage/network questions, the user only realizes the mistake after the nodes have been clustered. Now aborting the init process so the user can fix the mistake becomes much more expensive because we have already changed state on every system, before the user has made any choices.

microcloud/service/lxd.go
```diff
@@ -225,8 +225,16 @@ func (s LXDService) IssueToken(ctx context.Context, peer string) (string, error)
 }
 
 // ClusterMembers returns a map of cluster member names and addresses.
-func (s LXDService) ClusterMembers(ctx context.Context) (map[string]string, error) {
-	client, err := s.Client(ctx, "")
+// Optionally sends a remote request using the information in ServerInfo.
+func (s LXDService) ClusterMembers(ctx context.Context, info *mdns.ServerInfo) (map[string]string, error) {
```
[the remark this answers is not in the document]
masnax: I've added `RemoteClusterMembers` to do this now.

microcloud/cmd/microcloud/ask.go
masnax: No, the type represents consrs which are actual service names, capitalized appropriately. So `MicroCloud`, `MicroOVN`, `MicroCeph`, `LXD`.

microcloud/cmd/microcloud/ask.go
masnax: Oops, bad squash.

microcloud/cmd/microcloud/main_init.go
masnax: I've split it up now and added some more info in the commit message. 

Basically I thought that since the token issuing process was becoming more complex, and `AddPeers` was going to handle other components of the existing-cluster detection, it made more sense to move all the related logic over there and keep `waitForJoin` to strictly request a join and wait for it to complete.

microcloud/test/includes/microcloud.sh
masnax: Removed it.

microcloud/service/microcloud.go
masnax: Instead of filling this PR with all of those changes, I've added a new PR that moves all this implementation into a helper so we can just call that instead: #287

masnax:
Rebased.

microcloud/cmd/microcloud/main_init_preseed.go
masnax: I've also added a warning that the service will be skipped now, so the user is aware of it.

masnax:
@roosterfish @markylaing Could you two please give this PR another look-over? Thanks!

roosterfish:
> @roosterfish @markylaing Could you two please give this PR another look-over? Thanks!

Is the pipeline expected to fail or waiting on other PRs?

masnax:
> > @roosterfish @markylaing Could you two please give this PR another look-over? Thanks!
> 
> Is the pipeline expected to fail or waiting on other PRs?

Ah yeah the fix for the tests has been merged now, so I've just rebased to re-run them (though the test suite is still unreliable so there may be some expected failures).

masnax:
OK, had to make 2 small changes:

* `ReuseExistingCluster` is explicitly blocked on `microcloud add --preseed` because there's too much complexity to consider when accommodating all the existing nodes into the joining node's pre-existing cluster. Instead there's #260 which should facilitate this.

* Fixed a bug introduced into `microcloud add` which wasn't correctly considering the cluster size when joining new nodes: https://github.com/canonical/microcloud/pull/259/files#diff-a0869b68249cc46dfcc32cb652f562f713b0711198ccff1bd0cc7b5a47d7fbb0R436-R443

roosterfish:
The new `test_reuse_cluster()` test is failing right at the beginning. From the output it looks it uses some wrong values for the question/answer dialog.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -333,32 +333,38 @@ func lookupPeers(s *service.Handler, autoSetup bool, iface *net.Interface, subne
 // waitForJoin issues a token and instructs a system to request a join,
```
markylaing, line 333: Looks like the comment needs to be updated as it no longer issues a token.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -333,32 +333,38 @@ func lookupPeers(s *service.Handler, autoSetup bool, iface *net.Interface, subne
...
 // and then waits for the request to either complete or time out.
 // If the request was successful, it additionally waits until the cluster appears in the database.
-func waitForJoin(sh *service.Handler, clusterSize int, secret string, peer string, cfg types.ServicesPut) error {
+func waitForJoin(sh *service.Handler, clusterSize map[types.ServiceType]int, secret string, peer string, cfg types.ServicesPut) error {
```
markylaing, line 336: Similarly, the `clusterSize` variable no longer describes what the variable is.

markylaing (changes requested):
Just a couple of nits but otherwise looks good :+1: 

Tests are failing though :cry:

masnax:
Test issues were just an incorrect parameter (I was passing `reuse` to the new question instead of `add` which was suggested in review).

masnax:
@ru-fu I've added a commit that documents this feature (and another related feature in a bit more detail). I'm not sure if this is quite the right place so if you have any ideas I'd be happy to hear them :)

https://github.com/canonical/microcloud/pull/259/commits/2ad62dd4af6f8d9a9069110b68267d98c0884149

masnax:
Another odd error with the test suite: https://github.com/canonical/microcloud/actions/runs/8822867943/job/24222020581?pr=259#step:9:8673

This should be `lxd_disk1`, so that means 1 of 2 things happened:
* The sorting of `lxc info --resources` doesn't necessarily match up with the ID given to the disk.

* Somehow the disk wasn't reset properly. This one would be really weird since we manually destroy and add back the volumes between each test, before we restore from the snapshot that was taken during the initial setup of the test suite.

@simondeziel Maybe we should add something like
```bash
lxc exec "${name}" -- lxc query "/1.0/resources" | jq -r '.storage.disks[] | {id, device_id, device_path}'
```

during `validate_system_lxd` before we validate storage so that we can catch it if it happens again?

simondeziel:
> Another odd error with the test suite: https://github.com/canonical/microcloud/actions/runs/8822867943/job/24222020581?pr=259#step:9:8673

That's the exact error I often ran into.

> @simondeziel Maybe we should add something like
> 
> ```shell
> lxc exec "${name}" -- lxc query "/1.0/resources" | jq -r '.storage.disks[] | {id, device_id, device_path}'
> ```
> 
> during `validate_system_lxd` before we validate storage so that we can catch it if it happens again?

I'm all for it, short and useful output!

masnax:
Not sure what's going on with the test, since it passes fine locally. 

I added a debug commit to cat the test logs to figure it out. The error seems to be on the interactive input, so maybe the debug binary isn't applying properly?

doc/how-to/initialise.rst
```diff
@@ -90,6 +90,39 @@ Complete the following steps to initialise MicroCloud:
 
 See an example of the full initialisation process in the :ref:`Get started with MicroCloud <initialisation-process>` tutorial.
 
+Excluding MicroCeph or MicroOVN from MicroCloud
+-----------------------------------------------
```
ru-fu, line 94:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This section should be a subsection of "Interactive configuration", right?
Then we'll need to use another symbol for the heading (not sure about the length of the suggestion :wink: - should be as long as the title).

doc/how-to/initialise.rst
```diff
@@ -90,6 +90,39 @@ Complete the following steps to initialise MicroCloud:
...
+
+If the MicroOVN or MicroCeph snap is not installed on the system that runs :command:`microcloud init`, you will be prompted with the following question:
+
+  .. terminal::
```
ru-fu, line 98:
```suggestion
If the MicroOVN or MicroCeph snap is not installed on the system that runs :command:`microcloud init`, you will be prompted with the following question::

```

No need for `terminal::` here - we use this when we want to show input and output. For only output, code style is sufficient.

doc/how-to/initialise.rst
```diff
@@ -90,6 +90,39 @@ Complete the following steps to initialise MicroCloud:
...
+     MicroCeph not found. Continue anyway? (yes/no) [default=yes]:
+
+     MicroOVN not found. Continue anyway? (yes/no) [default=yes]:
+
+  #. If you choose ``yes``,  only existing services will be configured on all systems.
+
+  #. If you choose ``no`` the setup will be aborted.
```
ru-fu, line 105:
```suggestion
If you choose ``yes``, only existing services will be configured on all systems.
If you choose ``no``, the setup will be stopped.
```

No need for a list here imo, the paragraph is short enough. :)

doc/how-to/initialise.rst
```diff
@@ -90,6 +90,39 @@ Complete the following steps to initialise MicroCloud:
...
+
+All other systems must have at least the same set of snaps installed as the system that runs :command:`microcloud init`, otherwise they will not be available to select from the list of systems.
+Any questions associated to these systems will be skipped. For example, if MicroCeph is not installed, you will not be prompted for distributed storage configuration.
+
+Reusing an existing MicroCeph or MicroOVN with MicroCloud
+---------------------------------------------------------
```
ru-fu, line 111:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

doc/how-to/initialise.rst
```diff
@@ -90,6 +90,39 @@ Complete the following steps to initialise MicroCloud:
...
+
+If some of the systems are already part of a MicroCeph or MicroOVN cluster, you may reuse this cluster when initialising MicroCloud when prompted with the following question:
```
ru-fu, line 113:
```suggestion
If some of the systems are already part of a MicroCeph or MicroOVN cluster, you can choose to reuse this cluster when initialising MicroCloud when prompted with the following question::
```

doc/how-to/initialise.rst
```diff
@@ -90,6 +90,39 @@ Complete the following steps to initialise MicroCloud:
...
+
+  .. terminal::
```
ru-fu, line 115:
```suggestion
```

doc/how-to/initialise.rst
```diff
@@ -90,6 +90,39 @@ Complete the following steps to initialise MicroCloud:
...
+     "micro01" is already part of a MicroCeph cluster. Do you want to add this cluster to MicroCloud? (add/skip) [default=add]:
+
+     "micro01" is already part of a MicroOVN cluster. Do you want to add this cluster to MicroCloud? (add/skip) [default=add]:
+
+  #. If you choose ``add``, MicroCloud will add the remaining systems selected for initialisation to the pre-existing cluster.
+
+  #. If you choose ``skip``, MicroCloud will not configure the service to work with MicroCloud on any system.
```
ru-fu, line 122:
```suggestion
If you choose ``add``, MicroCloud will add the remaining systems selected for initialisation to the pre-existing cluster.
If you choose ``skip``, MicroCloud will not configure the service to work with MicroCloud on any system.
```

ru-fu: If you choose skip, will MicroOVN/Ceph be set up in the usual way (forming a new cluster)? Or will it not be set up at all?
That's unclear from the description.

doc/how-to/initialise.rst
```diff
@@ -90,6 +90,39 @@ Complete the following steps to initialise MicroCloud:
...
+
+If more than one MicroCeph or MicroOVN cluster exists among the systems, the MicroCloud initialisation will abort.
```
ru-fu, line 124:
```suggestion
If more than one MicroCeph or MicroOVN cluster exists among the systems, the MicroCloud initialisation will be cancelled.
```

ru-fu (commented):
Thanks! I think this is a good place for the information.
Just some general comments. :)

microcloud/test/suites/basic.sh
```diff
@@ -822,3 +822,135 @@ test_auto() {
...
+  lxc exec micro03 -- microovn cluster bootstrap
+  microcloud_interactive | lxc exec micro01 -- sh -c "microcloud init > out"
+  services_validator
+
+  reset_systems 3 3 3
+  echo Create a MicroCloud that re-uses an existing existing service with multiple nodes from this cluster
+  export REUSE_EXISTING_COUNT=1
+  export REUSE_EXISTING="add"
+  lxc exec micro02 -- microceph cluster bootstrap
+  token="$(lxc exec micro02 -- microceph cluster add micro01)"
```
simondeziel, line 892: I didn't know`microceph` diverged in behavior from LXD where we need to pass `--quiet` to avoid showing the text header. Is that the case for `micro*`?

microcloud/test/suites/basic.sh
```diff
@@ -822,3 +822,135 @@ test_auto() {
...
+  export SETUP_OVN="yes"
+  export OVN_FILTER="enp6s0"
+  export IPV4_SUBNET="10.1.123.1/24"
+  export IPV4_START="10.1.123.100"
+  export IPV4_END="10.1.123.254"
+  export DNS_ADDRESSES="10.1.123.1,8.8.8.8"
+  export IPV6_SUBNET="fd42:1:1234:1234::1/64"
+
+  reset_systems 3 3 3
+  echo Create a MicroCloud that re-uses an existing service
```
simondeziel, line 858: Could you quote the text after all those `echo`, please? IIRC, this helps the text highlighter in some IDEs.

microcloud/test/suites/basic.sh
```diff
@@ -822,3 +822,135 @@ test_auto() {
...
+  lxc exec micro04 -- snap disable microcloud
+  lxc exec micro02 -- microceph cluster bootstrap
+  token="$(lxc exec micro02 -- microceph cluster add micro04)"
+  lxc exec micro04 -- microceph cluster join "${token}"
+  microcloud_interactive | lxc exec micro01 -- sh -c "microcloud init > out"
+  services_validator
+  validate_system_microceph micro04 1
+
+  reset_systems 3 3 3
+  echo Fail to create a MicroCloud due to an existing service if --auto specified
```
simondeziel, line 918: This is one good example where double quoting the string passed to `echo` would help the text highlighting because this `if` shows as a command instead of text. :)

microcloud/test/suites/basic.sh
```diff
@@ -822,3 +822,135 @@ test_auto() {
...
+storage:
+  local:
+    - find: id == sdb
+      wipe: true
+  ceph:
+    - find: id == sdc
+      wipe: true
+  cephfs: true
+"
+  lxc exec micro01 -- sh -c "cat /root/preseed.yaml | TEST_CONSOLE=0 microcloud init --preseed"
```
simondeziel, line 954: Is the intermediate `/root/preseed.yaml` needed/useful for debugging? If not, would it be possible to use something like:
```
lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud init --preseed << EOF

lookup_subnet: ${addr}/24
lookup_interface: enp5s0
reuse_existing_clusters: true
systems:
...
EOF
```

microcloud/test/suites/basic.sh
masnax: the `micro*` projects don't have any header, they just spit out the token.

doc/how-to/initialise.rst
masnax: Right, I meant that it won't be set up at all. 

How about:
`If you choose ``skip``, MicroCloud will ignore the service on all systems`?

ru-fu: Or maybe:
```
If you choose ``skip``, the respective service will not be set up at all.
```
"Ignoring the service" could also be interpreted as "ignoring the existing service/cluster".

doc/how-to/initialise.rst
```diff
@@ -90,6 +90,37 @@ Complete the following steps to initialise MicroCloud:
...
+
+If the MicroOVN or MicroCeph snap is not installed on the system that runs :command:`microcloud init`, you will be prompted with the following question:
+
+     MicroCeph not found. Continue anyway? (yes/no) [default=yes]:
+
+     MicroOVN not found. Continue anyway? (yes/no) [default=yes]:
+
+  If you choose ``yes``,  only existing services will be configured on all systems.
+
+  If you choose ``no`` the setup will be cancelled.
```
ru-fu, line 104:
```suggestion
    MicroCeph not found. Continue anyway? (yes/no) [default=yes]:

    MicroOVN not found. Continue anyway? (yes/no) [default=yes]:

If you choose ``yes``,  only existing services will be configured on all systems.
If you choose ``no``, the setup will be cancelled.
```

Too much indentation (must be 4 spaces afaik).
Also, missing comma in the last sentence, and I would combine the last two sentences into one paragraph (which means get rid of the empty line between them).

doc/how-to/initialise.rst
```diff
@@ -90,6 +90,35 @@ Complete the following steps to initialise MicroCloud:
 
 See an example of the full initialisation process in the :ref:`Get started with MicroCloud <initialisation-process>` tutorial.
 
+Excluding MicroCeph or MicroOVN from MicroCloud
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If the MicroOVN or MicroCeph snap is not installed on the system that runs :command:`microcloud init`, you will be prompted with the following question:
```
ru-fu, line 96:
```suggestion
If the MicroOVN or MicroCeph snap is not installed on the system that runs :command:`microcloud init`, you will be prompted with the following question::
```

Sorry, missed that in my previous review. I had it in my first one though. :stuck_out_tongue_closed_eyes: 
You need two colons to introduce a code block.

Same change below as well.

masnax:
@roosterfish @tomponline Could one of you please look this over one last time before merging? Thanks!

masnax:
> > (That 1 test failure is an issue with the test suite being too hard on the runners, it passes consistently when run locally).
> 
> @masnax does this mean if we merge this we will get that test failing going forward then?
> 
> If so then thats a merge blocker to me.

The test will be just as flaky on any PR regardless of if this one is merged. The test was not affected by this PR, the issue is that we just can't consistently run 4 VMs on the github runners without them running into issues like LXD being unable to start up.

tomponline:
I see @masnax thanks for clarifying that. It sounded like this PR was introducing the tests that were flaky.

I'll leave the final review to @roosterfish

microcloud/cmd/microcloud/main_init.go
```diff
@@ -532,6 +539,83 @@ func validateSystems(s *service.Handler, systems map[string]InitSystem) (err err
...
+		for k, v := range remoteClusterMembers {
+			host, _, err := net.SplitHostPort(v)
+			if err != nil {
+				return "", nil, err
+			}
+
+			clusterMembers[k] = host
+		}
+
+		if autoSetup {
```
roosterfish, line 646: As `autoSetup` doesn't get modified in this func, would it be a better choice not passing this var into `checkClustered` and performing the check outside?

microcloud/cmd/microcloud/main_init.go
```diff
@@ -369,61 +375,117 @@ func waitForJoin(sh *service.Handler, clusterSize int, secret string, peer strin
...
+	// If the local node needs to join an existing cluster, do it first so we can proceed as normal.
+	if len(joinConfig[sh.Name].Tokens) > 0 {
+		cfg := joinConfig[sh.Name]
+		err := waitForJoin(sh, clusterSize, "", sh.Name, cfg)
+		if err != nil {
+			return err
+		}
+
+		// Sleep 3 seconds to give the cluster roles time to reshuffle before adding more members.
+		time.Sleep(3 * time.Second)
```
roosterfish, line 475: I can remember this raised my attention the last time when reviewing it.

As this pattern is used at other places this shouldn't be a blocker but would it be another option retrying the "adding more members" operations instead with a timeout in case it doesn't work?

microcloud/service/microcloud.go
roosterfish: Ok so https://github.com/canonical/microcloud/pull/287 has to come after this one then?

microcloud/test/suites/basic.sh
```diff
@@ -822,3 +822,134 @@ test_auto() {
...
+  export REUSE_EXISTING_COUNT=1
+  export REUSE_EXISTING="add"
+  lxc exec micro02 -- microceph cluster bootstrap
+  token="$(lxc exec micro02 -- microceph cluster add micro01)"
+  lxc exec micro01 -- microceph cluster join "${token}"
+  microcloud_interactive | lxc exec micro01 -- sh -c "microcloud init > out"
+  services_validator
+
+  reset_systems 3 3 3
+  echo "Create a MicroCloud that re-uses an existing existing service with all nodes from this cluster"
```
roosterfish, line 898: Two times `existing`.

roosterfish (changes requested):
Just a few more small comments, rest still looking good to me.

microcloud/api/types/services.go
```diff
@@ -36,3 +36,9 @@ type ServiceToken struct {
 	Service   ServiceType `json:"service" yaml:"service"`
 	JoinToken string      `json:"join_token" yaml:"join_token"`
 }
+
+// ServiceTokensPost represents a request to issue a join token for a MicroCloud service.
+type ServiceTokensPost struct {
+	ClusterAddress string `json:"cluster_address" yaml:"cluster_address"`
+	JoinerName     string `json:"joiner_name"     yaml:"joiner_name"`
```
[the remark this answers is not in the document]
masnax: No it won't become obsolete, the name is used to identify the secret join key in the issuer's database. The change in microcluster removed the name only from the base64 encoded payload which included the secret key and the join addresses: https://github.com/canonical/microcluster/blob/ceb18595e3c6e75150d1a5106a6b363082a1c794/internal/rest/types/tokens.go#L11-L28

You should see the first struct still has a `Name` field.

microcloud/cmd/microcloud/main_init.go
masnax: This is here because retrying the connection too quickly can result in dqlite locking up and you would get hit with a `context deadline exceeded` error. 

This was mainly an issue with LXD's dqlite implementation which I've since addressed, but I haven't gotten around to testing if I can safely remove these timeouts in MicroCloud. I'll file an issue for this, thanks for reminding me.

microcloud/cmd/microcloud/main_init.go
masnax: It's just for efficiency. We can move it outside but that would mean that we would have to run through the rest of `checkClustered`'s implementation. Where it is now, as soon as at least 1 system gets to this point, we can exit if `autoSetup` is true, rather than continuing to build the list and checking more systems, only to error out afterward anyway.

microcloud/cmd/microcloud/main_init.go
roosterfish: Right makes sense.

microcloud/api/types/services.go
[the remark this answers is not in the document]
roosterfish: Right makes sense.

microcloud/cmd/microcloud/main_init.go
roosterfish: Ah sorry I think I got distracted reviewing whilst trying to lookup the issue. Have you logged it in another repo?

roosterfish (approved):

microcloud/cmd/microcloud/main_init.go
masnax: Instead of that I'm just working on a PR to remove them :)

roosterfish: Cool, thanks!

PR #261, opened 2024-02-07
microcloud/service: Only set source.wipe if user explicitly agrees

masnax:
If LXD doesn't have the corresponding schema extension for the source.wipe config key, MicroCloud will never ask the user if they want to wipe disks, but will still populate the config with a 'false' value, causing an error.

Instead, we can simply omit the key if it's false, and it will be implicitly considered false.

### DIFF
--- a/microcloud/service/lxd_config.go
+++ b/microcloud/service/lxd_config.go
@@ -101,11 +101,16 @@ func (s LXDService) DefaultOVNNetwork(ipv4Gateway string, ipv4Range string, ipv6
 // DefaultPendingZFSStoragePool returns the default local storage configuration when
 // creating a pending pool on a specific cluster member target.
 func (s LXDService) DefaultPendingZFSStoragePool(wipe bool, path string) api.StoragePoolsPost {
+	cfg := map[string]string{"source": path}
+	if wipe {
+		cfg["source.wipe"] = strconv.FormatBool(wipe)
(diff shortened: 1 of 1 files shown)


roosterfish (commented):
Looking good. Do you know what's wrong with the doc check in the pipeline?

ru-fu:
> No clue, looks like a spellcheck issue in the docs. I can reproduce it when running `make spelling -C doc` Got any ideas @ru-fu?

This is crazy ... :rofl: 

So, there is suddenly a `vfalse` in a lot of the MicroCloud doc pages. Looking through the HTML code, it appears for links to the LXD docs - `title="(in Canonical LXD vfalse)"`.
These links are generated through Intersphinx, which is part of the Sphinx documentation generator that we use. However, there hasn't been an update to neither Sphinx nor Intersphinx ...

Checking the published docs for what it is supposed to say instead of `vfalse` - oh, it's the version!
`title="(in Canonical LXD v5.20)"`

The version is set in https://github.com/canonical/lxd/blob/def587b98baf65d93ff15d64c6dfdec84b95c1ad/doc/custom_conf.py#L30 
And we just added some stuff to the `flex.go` file, which broke this horrible parsing ... :wink: 

Let me see how I can fix that extraction.

roosterfish (approved):

PR #262, opened 2024-02-08
github: drop DCO check now that we have the DCO app enabled

simondeziel:


roosterfish (approved):

tomponline (approved):

simondeziel:
@tomponline it looks like the old "Signed-off-by (DCO)" will never complete because I'm dropping it in this very PR. Can you loosen the requirements for merging?

PR #263, opened 2024-02-13
lxd: parametrize LXD with `instances.migration.stateful` set to `true`

gabrielmougard:
When [this](https://github.com/canonical/lxd/pull/12832) is merged, this PR adds the integration with MicroCloud. After that, all the created VM instances in a MicroCloud deployment should be able to be live migrated with no additional user settings.

### DIFF
--- a/microcloud/service/lxd.go
+++ b/microcloud/service/lxd.go
@@ -114,6 +114,9 @@ func (s LXDService) Bootstrap(ctx context.Context) error {
 	newServer := currentServer.Writable()
 	newServer.Config["core.https_address"] = "[::]:8443"
 	newServer.Config["cluster.https_address"] = addr
+	if client.HasExtension("migration_stateful_default") {
+		newServer.Config["instances.migration.stateful"] = "true"
+	}
 
 	// Apply it.
 	err = client.UpdateServer(newServer, etag)


roosterfish (commented):
Wouldn't this also apply for VMs backed by local storage? I thought right now we can do this just for Ceph based VMs since we don't require the `size.state` in this case.

tomponline:
> Wouldn't this also apply for VMs backed by local storage? I thought right now we can do this just for Ceph based VMs since we don't require the `size.state` in this case.

This isn't related to size.state, its related to migration.stateful, see https://github.com/canonical/lxd/pull/12832#discussion_r1485854257

tomponline (changes requested):
This needs to detect if the LXD server has the appropriate API extension and only set if it if does.

roosterfish:
> > Wouldn't this also apply for VMs backed by local storage? I thought right now we can do this just for Ceph based VMs since we don't require the `size.state` in this case.
> 
> This isn't related to size.state, its related to migration.stateful, see [canonical/lxd#12832 (comment)](https://github.com/canonical/lxd/pull/12832#discussion_r1485854257)

I forgot the fact that the new config key `instances.migration.stateful` only applies for remote storage. Thanks for the clarification.

tomponline:
> I forgot the fact that the new config key `instances.migration.stateful` only applies for remote storage. Thanks for the clarification.

No, it doesnt, it applies to all storage pools.

roosterfish:
> > I forgot the fact that the new config key `instances.migration.stateful` only applies for remote storage. Thanks for the clarification.
> 
> No, it doesnt, it applies to all storage pools.

Sorry my wording was wrong, I implied the effect of having `instances.migration.stateful` now makes stateful migration feel seamless for remote backed VMs.
If I now got it right, also a VM on local storage will get its `migrate.stateful` (if not already set) config key set, but will still complain on `size.state` if not there.

tomponline:
> If I now got it right, also a VM on local storage will get its `migrate.stateful` (if not already set) config key set, but will still complain on `size.state` if not there.

Correct.

masnax:
Is this a global config key? I see you're setting it on bootstrap, but you may need to include it in the `JoinConfig` payloads that configure other nodes when they join the cluster.

tomponline (approved):

tomponline:
Will wait for @masnax approval before merging.

masnax:
Ah I thought this was still being worked on. It looks good to me, so we can merge it.

PR #264, opened 2024-02-19
doc: clarify which snap channels to use for production-ready version

ru-fu:


tomponline (approved):

roosterfish (approved):
Looks good! Just for clarification, if you don't specify the channel, it still picks the stable one right?

PR #269, opened 2024-03-15
Use both IPv4 and IPv6 chosen addresses for default DNS addresses

masnax:
This PR does 2 things:

* Moves the DNS question to be after both the ipv4 and ipv6 questions. Right now it's in the middle, which feels awkward.
* Because of the above, when choosing the default DNS nameservers, we only have the user's chosen IPv4 address at that point. By moving the question to the end, we can include both the chosen IPv4 and IPv6 addresses by default.

### DIFF
--- a/microcloud/cmd/microcloud/ask.go
+++ b/microcloud/cmd/microcloud/ask.go
@@ -696,8 +696,8 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 	fmt.Println("")
 
 	// Prepare the configuration.
+	var dnsAddresses string
 	ipConfig := map[string]string{}
-	dnsConfig := map[string]string{}
 	if bootstrap {
 		for _, ip := range []string{"IPv4", "IPv6"} {
 			validator := func(s string) error {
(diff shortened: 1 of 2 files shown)


microcloud/cmd/microcloud/ask.go
```diff
@@ -740,25 +740,27 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
...
-					}
 				} else {
 					ipConfig[gateway] = ""
 				}
 			}
 		}
+
+		gateways := []string{}
+		for gateway := range ipConfig {
+			gatewayAddr, _, err := net.ParseCIDR(gateway)
```
roosterfish, line 751: Why do you run `net.ParseCIDR` here? Is it for the purpose of validating that the gateway is valid? I think the validation is already done in the step before when invoking `AskString`.

Below you again get the string from the `net.IP` and append it to the list of gateways. So could you just take the gateway from the map and append it to the list?

roosterfish: Nevermind, you want to extract the IP only without the subnetmask.

roosterfish (approved):
LGTM!

gabrielmougard (approved):
LGTM

PR #270, opened 2024-03-15
doc: Update docs for cephfs and dns nameservers

masnax:
Should be merged after #269 because I wrote it based on that.

Updates the docs for the newer features (cephfs and dns nameservers), and also mentions that the mDNS lookup list updates as new machines are found periodically.

### DIFF
--- a/doc/explanation/initialisation.rst
+++ b/doc/explanation/initialisation.rst
@@ -15,9 +15,9 @@ Automatic server detection
 MicroCloud uses :abbr:`mDNS (multicast DNS)` to automatically detect other servers on the network.
 This method works in physical networks, but it is usually not supported in a cloud environment.
 
-By default, the scan is limited to the local subnet.
+The scan can be limited to the default local subnet of the network interface you select.
 
-MicroCloud will display all servers that it detects, and you can then select the ones you want to add to the MicroCloud cluster.
+MicroCloud will display all servers that it detects and periodically update the list. You can select the servers you want to add to the MicroCloud cluster.
 
 .. _bootstrapping-process:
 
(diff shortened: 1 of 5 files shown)


doc/how-to/preseed.yaml
```diff
@@ -29,6 +29,7 @@ ovn:
   ipv4_gateway: 10.0.0.1/24
   ipv4_range: 10.0.0.100-10.0.0.254
   ipv6_gateway: cafe::1/64
+  dns_servers: 10.0.0.1,cafe::1
```
roosterfish, line 32: Just out of interested, does the order matter here? I saw it in https://github.com/canonical/microcloud/pull/269 too.

roosterfish (changes requested):
Looks good to me but spell check is failing. I guess `CephFS` needs to be added to this word list.

tomponline:
Is there  reason we aren't using the  IP test ranges in the docs, like we do in LXD?

doc/explanation/initialisation.rst
```diff
@@ -15,9 +15,9 @@ Automatic server detection
 MicroCloud uses :abbr:`mDNS (multicast DNS)` to automatically detect other servers on the network.
 This method works in physical networks, but it is usually not supported in a cloud environment.
 
-By default, the scan is limited to the local subnet.
+By default, the scan is limited to the default local subnet and default interface, but can be configurable to use a different interface and subnet.
```
ru-fu, line 18:
```suggestion
By default, the scan is limited to the default interface and default local subnet, but can be configured to use a different interface and subnet.
```

ru-fu: How can it be configured?

doc/explanation/initialisation.rst
```diff
@@ -15,9 +15,9 @@ Automatic server detection
...
 
-MicroCloud will display all servers that it detects, and you can then select the ones you want to add to the MicroCloud cluster.
+MicroCloud will display all servers that it detects and periodically update the list. You can select the ones you want to add to the MicroCloud cluster.
```
ru-fu, line 20:
```suggestion
MicroCloud will display all servers that it detects and periodically update the list. You can select the servers you want to add to the MicroCloud cluster.
```

doc/how-to/initialise.rst
```diff
@@ -71,6 +71,8 @@ Complete the following steps to initialise MicroCloud:
       You must select at least three disks.
    #. Select whether you want to wipe any of the disks.
       Wiping a disk will destroy all data on it.
+
+   #. You can choose to optionally set up a CephFS distributed file-system.
```
ru-fu, line 75:
```suggestion
   #. You can choose to optionally set up a CephFS distributed file system.
```

doc/how-to/preseed.yaml
```diff
@@ -39,7 +40,9 @@ ovn:
 # String values must not be in quotes unless the string contains a space.
 # Single quotes are fine, but double quotes must be escaped.
 # `find_min` and `find_max` can be used to validate the number of disks each filter finds.
+# cephfs: true can be used to optionally set up a CephFS file-system alongside Ceph distributed storage.
```
ru-fu, line 43:
```suggestion
# `cephfs: true` can be used to optionally set up a CephFS file system alongside Ceph distributed storage.
```

doc/tutorial/get_started.rst
```diff
@@ -271,6 +271,7 @@ Complete the following steps:
    #. Select ``yes`` to confirm that there are fewer disks available than machines.
    #. Select all listed disks (these should be ``remote1``, ``remote2``, and ``remote3``).
    #. You don't need to wipe any disks (because we just created them).
+   #. Select ``yes`` to optionally configure the CephFS distributed file-system.
```
ru-fu, line 274:
```suggestion
   #. Select ``yes`` to optionally configure the CephFS distributed file system.
```

ru-fu (changes requested):
Some nitpicks.

(Just as explanation: it's generally "file system", but when used as a modifier it gets a hyphen - "file-system storage".)

ru-fu:
> Is there reason we aren't using the IP test ranges in the docs, like we do in LXD?

No reason, it must have slipped through, and we should replace the IPs.
@masnax : Here's the list of IP ranges https://en.wikipedia.org/wiki/Reserved_IP_addresses

doc/explanation/initialisation.rst
masnax: It's the first question prompt, or both `lookup_interface` and `lookup_subnet` in the preseed file.

ru-fu: Ah okay, so the selection for "Select an address for MicroCloud's internal traffic"? But then there is no default, since you always have to select something?

ru-fu: So like this?

```suggestion
The scan can be limited to the default local subnet of the network interface you select.
```

doc/how-to/preseed.yaml
masnax: In the LXD OVN driver, [we split the list by ipv4 and ipv6](https://github.com/canonical/lxd/blob/01f0097066abd69f9b7d53f2b2278f724998a1ed/lxd/network/driver_ovn.go#L1125-L1143) so I I don't believe the order matters.

roosterfish (approved):

ru-fu (approved):
Looks good, thanks!

PR #274, opened 2024-04-02
Ceph: Add interactive and preseed options to configure a dedicated Ceph network in MicroCloud 

gabrielmougard:
See https://discourse.ubuntu.com/t/add-options-for-configuring-a-dedicated-ceph-network-in-microcloud/43687
JIRA ticket: https://warthogs.atlassian.net/jira/software/c/projects/LXD/boards/54?assignee=63d9eb6128cddcc70770aff1&selectedIssue=LXD-636

* [x] Bootstrapping with a dedicated Ceph Network
* [x] Adding new cluster members to a cluster with an existing dedicated Ceph network
  * [x] If the new cluster members have the right physical network interfaces with IPs that are within the subnet, it should work.  Else, error out.


microcloud/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/canonical/microcloud/microcloud
 
-go 1.18
+go 1.21.4
```
tomponline, line 3: is this required for the change, why not go 1.22 like LXD?

gabrielmougard: My laptop is running 1.21.4 (unlike my workstation that runs 1.22). I'll update this to 1.22 (alongside with my laptop go version)

simondeziel: Was that (like in https://github.com/canonical/microcloud/actions/runs/8525721170/job/23353341519?pr=268):

```
 # github.com/canonical/microcluster/internal/db
Error: ../../../../go/pkg/mod/github.com/canonical/microcluster@v0.0.0-20240311124409-8d4c3238a91b/internal/db/db.go:134:21: not enough arguments in call to query.Retry
	have (func() error)
	want (context.Context, func(ctx context.Context) error)
make: *** [Makefile:9: build] Error 1
Error: Process completed with exit code 2.
```

microcloud/cmd/microcloud/main_init.go
```diff
@@ -417,10 +418,21 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
 // setupCluster Bootstraps the cluster if necessary, adds all peers to the cluster, and completes any post cluster
 // configuration.
 func setupCluster(s *service.Handler, systems map[string]InitSystem) error {
-	_, bootstrap := systems[s.Name]
+	bootstapSystem, bootstrap := systems[s.Name]
```
sabaini, line 421: Nit: s/bootstapSystem/bootstrapSystem/

microcloud/cmd/microcloud/main_init.go
```diff
@@ -417,10 +418,21 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
...
+				// make the public and cluster traffic flow through the same interfaces.
+				microCephBootstrapConf := map[string]string{
+					"PublicNet":  bootstapSystem.MicroCephNetworkSubnet,
+					"ClusterNet": bootstapSystem.MicroCephNetworkSubnet,
+				}
+
+				s.SetConfig(microCephBootstrapConf)
+			}
+
 			err := s.Bootstrap(context.Background())
```
sabaini, line 436: I wonder about the address the microcephd binds to; it looks to me that this would choose the same address as the other microcloud services right? Should this be changed to choose an address from the MicroCephNetworkSubnet?

sabaini (commented):
Generally lgtm, two comments below. 

As a heads-up separating public and cluster nets for Ceph was a relatively frequently requested feature so I wonder if this should be really conflated (will add to the spec as well).

microcloud/cmd/microcloud/main_init.go
gabrielmougard: @sabaini Indeed, you're right. When `NewCephService` is called, the MicroCloud's internal network IP address of the bootstraping cluster member is chosen for microcephd to bind to. If a public subnet for Ceph has been specified, we should choose an address available on this subnet on this cluster member.

masnax: I'm not so familiar with where the boundary is between the cluster and public networks of microceph, and microcluster's network, but judging from `microceph cluster bootstrap` I believe you're right @gabrielmougard.

Assuming I have 3 network interfaces: 
- `10.0.0.100/24` (default)
- `20.0.0.100/24`
- `30.0.0.100/24`

And I run `microceph cluster bootstrap --cluster-network `20.0.0.100/24` --public-network `30.0.0.100/24`, then the resulting microcluster address displayed in `microceph cluster list` is still the default `10.0.0.100:7443`.

So the internal microcluster network should still be propagated to microceph from `NewCephService`.

doc/.wordlist.txt
```diff
@@ -37,3 +37,4 @@ Jira
 VM
 YAML
 CephFS
+cephbr
```
ru-fu, line 40: This is not a word and shouldn't be here. ;)

doc/explanation/microcloud.rst
```diff
@@ -75,6 +75,9 @@ Disk sizes can vary.
 
 For distributed storage, MicroCloud uses MicroCeph, which is a lightweight way of deploying a Ceph cluster.
 To use distributed storage, you must have at least three disks (attached to at least three different machines).
+It is also possible to specify a Ceph network through the initialisation wizard or through the preseed file.
+If you choose to have a dedicated Ceph network, each of your cluster member must have at least one network interface with
```
ru-fu, line 79:
```suggestion
If you choose to have a dedicated Ceph network, each of your cluster members must have at least one network interface with
```

doc/tutorial/get_started.rst
```diff
@@ -145,6 +145,17 @@ Complete the following steps to set up this network:
      lxc network get microbr0 ipv4.address
      lxc network get microbr0 ipv6.address
 
+You also have the option to create a dedicated network for the Ceph traffic:
```
ru-fu, line 148: This is a tutorial, so we shouldn't offer choices but decide whether this is something we want the reader to do or not.

Is it a rather special use case, or should anybody who tries out MicroCloud for the first time do this?

(An alternative could be to add a second tutorial for this scenario.)

doc/tutorial/get_started.rst
```diff
@@ -183,6 +194,13 @@ Complete the following steps:
      lxc config device add micro3 eth1 nic network=microbr0 name=eth1
      lxc config device add micro4 eth1 nic network=microbr0 name=eth1
 
+#. If you created a `cephbr0` network, create and add network interfaces that use the dedicated Ceph network to each VM::
```
ru-fu, line 197:
```suggestion
#. If you created a ``cephbr0`` network, create and add network interfaces that use the dedicated Ceph network to each VM::
```

This will fix the spelling issue. ;)
In RST, one single quote gives italics, and you need two for code markup.

doc/.wordlist.txt
gabrielmougard: fixed

doc/explanation/microcloud.rst
gabrielmougard: fixed

doc/tutorial/get_started.rst
gabrielmougard: fixed

gabrielmougard:
@markylaing @masnax @ru-fu ready for review :)

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,160 @@
...
+
+Here is a picture of the different Ceph network setups we talked about:
+
+.. figure:: /images/ceph_networking_setups.png
+   :alt: The different Ceph networking setups available
+   :align: center
+
+Now, let's see how to configure the different Ceph networking setups in the MicroCloud initialisation process.
+
+1. First, just like when you created an uplink network for MicroCloud so that the cluster members could have external connectivity, you will need to create a dedicated network for the Ceph cluster members to communicate with each other. Let's call it ``cephbr0``::
```
ru-fu, line 30: These are a lot of steps, which makes it hard to follow. Can you group the ones that belong together and use sub-steps?

As an example, you could combine steps 1-4 (I'm not writing the full text here, this is just to give you an idea what I mean):

1. Create the dedicated networks for Ceph:
    a. Create a network for internal traffic: ...
    b. For a fully disaggregated setup, create a second network for the public traffic: ...
    c. Find the assigned IP addresses:
       Maybe also for the second network:

ru-fu: Would it maybe be sufficient to give the full instructions for a fully disaggregated setup only, and then add a paragraph at the end that for a partially disaggregated setup, you do this with only one network and leave either the internal or the public traffic on the default MicroCloud internal subnet?

I feel that would make the instructions way more straight-forward, and it should be easy to follow.

doc/tutorial/get_started.rst
```diff
@@ -128,7 +128,7 @@ Complete the following steps to create the required disks in a LXD storage pool:
       | custom | remote3 |             | block        | 0       |
       +--------+---------+-------------+--------------+---------+
 
-3. Create a network
+3. Create networks
```
ru-fu, line 131: This (and the following changes) should be undone now that we have a separate page for it.

doc/tutorial/get_started.rst
```diff
@@ -272,6 +298,8 @@ Complete the following steps:
    #. Select all listed disks (these should be ``remote1``, ``remote2``, and ``remote3``).
    #. You don't need to wipe any disks (because we just created them).
    #. Select ``yes`` to optionally configure the CephFS distributed file system.
+   #. Enter the IPv4 or IPv6 CIDR subnet address that you noted down for your Ceph public network or leave it empty.
+   #. Enter the IPv4 or IPv6 CIDR subnet address that you noted down for your Ceph internal network or leave it empty.
```
ru-fu, line 302: Leave empty for both.

ru-fu (commented):
This looks really good for the documentation already! A few more changes needed though (of course :wink: ).

doc/explanation/microcloud.rst
```diff
@@ -64,6 +64,16 @@ MicroCloud will still be usable, but you will see some limitations:
   As a result of this, network forwarding works at a basic level only, and external addresses must be forwarded to a specific cluster member and don't fail over.
 - There is no support for hardware acceleration, load balancers, or ACL functionality within the local network.
 
+If you decide to set dedicated networks for Ceph to differentiate between public and internal Ceph traffic, you can do it through the initialisation wizard or through the preseed file.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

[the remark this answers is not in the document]
gabrielmougard: fixed

doc/explanation/microcloud.rst
```diff
@@ -64,6 +64,16 @@ MicroCloud will still be usable, but you will see some limitations:
...
+
+- You can choose to have a dedicated network for the Ceph public traffic and a dedicated network for the Ceph internal (cluster) traffic. This is a fully disaggregated Ceph network setup.
+  In the fully disaggregated setup, each of your cluster members must have at least two network interfaces with
+  both IP addresses being respectively in the same subnet as the Ceph public network and the Ceph internal network you specified.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/explanation/microcloud.rst
```diff
@@ -64,6 +64,16 @@ MicroCloud will still be usable, but you will see some limitations:
...
+- You can also choose to have a dedicated network where both the public and internal Ceph traffic will flow in. This is a partially disaggregated Ceph network setup.
+  In the partially disaggregated setup, each of your cluster members must have at least one network interface with an IP address in the same subnet as the Ceph dedicated network you specified.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/explanation/microcloud.rst
```diff
@@ -64,6 +64,16 @@ MicroCloud will still be usable, but you will see some limitations:
...
+
+See :ref:`howto-ceph-networking` for how to set it up.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,160 @@
+.. _howto-ceph-networking:
+
+How to configure Ceph networking
+================================
+
+In the ``microcloud init`` setup, you are asked if you want to provide custom subnets for the Ceph cluster.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,160 @@
...
+Here are the questions you will be asked:
+
+- `What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph public traffic on? [default: 203.0.113.0/24]: <answer>`
+- `What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph internal traffic on? [default: 203.0.113.0/24]: <answer>`
+
+You can choose to skip both questions (just click `Enter`) and use the default value which is the subnet used for the internal MicroCloud traffic.
+This is referred as a `usual Ceph networking setup`.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,160 @@
...
+
+But sometimes, you want to be able to use different network interfaces for some Ceph related usages.
+Let's imagine you have machines with network interfaces that are tailored for high throughput and low latency data transfer,
+like 100GbE+ QSFP links, and other ones that might be more suited for ‘management’ traffic, like 1GbE or 10GbE links.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,160 @@
...
+
+In this case, it would probably be ideal to set your Ceph internal (or cluster) traffic on the high throughput network interface and the Ceph public traffic on the management network interface. This is referred as a `fully disaggregated Ceph networking setup`.
+You could also decide to put both type of traffic on the same high throughput and low latency network interface. This is referred as a `partially disaggregated Ceph networking setup`.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,160 @@
...
+
+Here is a picture of the different Ceph network setups we talked about:
+
+.. figure:: /images/ceph_networking_setups.png
+   :alt: The different Ceph networking setups available
+   :align: center
+
+Now, let's see how to configure the different Ceph networking setups in the MicroCloud initialisation process.
```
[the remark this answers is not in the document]
gabrielmougard: Yes I agree. And yes, in a real life scenario, the Ceph network should already exist.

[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,160 @@
...
+like 100GbE+ QSFP links, and other ones that might be more suited for ‘management’ traffic, like 1GbE or 10GbE links.
+
+In this case, it would probably be ideal to set your Ceph internal (or cluster) traffic on the high throughput network interface and the Ceph public traffic on the management network interface. This is referred as a `fully disaggregated Ceph networking setup`.
+You could also decide to put both type of traffic on the same high throughput and low latency network interface. This is referred as a `partially disaggregated Ceph networking setup`.
+
+Here is a picture of the different Ceph network setups we talked about:
+
+.. figure:: /images/ceph_networking_setups.png
+   :alt: The different Ceph networking setups available
+   :align: center
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,160 @@
...
+    # If the `cephbr1` gateway address is `10.0.2.1/24` (subnet should be `10.0.2.0/24`)
+    ip link set enp8s0 up
+    # `X` should be a number between 2 and 254, different for each VM
+    ip addr add 10.0.2.X/24 dev enp8s0
+
+10. Now, you can start the MicroCloud initialisation process and provide the subnets you noted down in step 3 and 4 when asked for the Ceph networking subnets.
+
+#. If you only created ``cephbr0``, you can set the public and internal Ceph traffic on the same ``cephbr0`` subnet, but you don't have to. You can also choose to let one type of traffic on the default MicroCloud internal subnet::
+
+    What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph public traffic on? [default: 203.0.113.0/24] 10.0.1.0/24
```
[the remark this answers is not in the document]
gabrielmougard: `203.0.113.0/24` is already marked as a subnet used for the documentation in the page you provided. and `10.0.1.0/24` is within `10.0.0.0/8` that is used for private networks.

doc/tutorial/get_started.rst
gabrielmougard: fixed

doc/tutorial/get_started.rst
gabrielmougard: fixed

gabrielmougard:
@ru-fu doc udpated

microcloud/service/lxd.go
```diff
@@ -400,6 +400,105 @@ func (s LXDService) GetUplinkInterfaces(ctx context.Context, bootstrap bool, pee
...
+			if !shared.ValueInSlice(network.Type, []string{"physical", "bridge", "bond", "vlan"}) {
+				continue
+			}
+
+			state, err := clients[peer].GetNetworkState(network.Name)
+			if err != nil {
+				continue
+			}
+
+			// Only full L2 devices.
```
markylaing, line 466: Are there any docs you can link to to see the requirements?

microcloud/service/lxd.go
```diff
@@ -499,6 +499,58 @@ func (s LXDService) GetCephInterfaces(ctx context.Context, bootstrap bool, peers
...
+	for peer, interfaces := range interfacesMap {
+		for _, iface := range interfaces {
+			for _, addr := range iface.Addresses {
+				ip := net.ParseIP(addr)
+				if ip == nil {
+					return nil, fmt.Errorf("invalid IP address: %v", addr)
+				}
+
+				if subnet.IP.To4() != nil {
+					if ip.To4() != nil && subnet.Contains(ip) {
```
markylaing, line 525: Can the predicate here be moved into the one above to remove a layer of nesting? E.g.
```go
if subnet.IP.To4() != nil && ip.To4() != nil && subnet.Contains(ip) {
    // do IPv4 stuff
} else if subnet.IP.To6() != nil && ip.To6() != nil && subnet.Contains(ip) {
    // do IPv6 stuff
}
```

microcloud/cmd/microcloud/main_init.go
```diff
@@ -31,12 +31,14 @@ type InitSystem struct {
...
-	StoragePools       []lxdAPI.StoragePoolsPost              // Cluster-wide storage pool configuration.
-	StorageVolumes     map[string][]lxdAPI.StorageVolumesPost // Cluster wide storage volume configuration.
+	MicroCephDisks                 []cephTypes.DisksPost                  // Disks intended to be passed to MicroCeph.
+	MicroCephPublicNetworkSubnet   string                                 // Subnet for the Ceph public network.
+	MicroCephInternalNetworkSubnet string                                 // Subnet for the Ceph internal network.
+	TargetNetworks                 []lxdAPI.NetworksPost                  // Target specific network configuration.
+	TargetStoragePools             []lxdAPI.StoragePoolsPost              // Target specific storage pool configuration.
+	Networks                       []lxdAPI.NetworksPost                  // Cluster-wide network configuration.
+	StoragePools                   []lxdAPI.StoragePoolsPost              // Cluster-wide storage pool configuration.
+	StorageVolumes                 map[string][]lxdAPI.StorageVolumesPost // Cluster wide storage volume configuration.
```
markylaing, line 41: Can you please move all of these comments to the line above the field definition and begin them with the field name so that they are Godoc compatible?

markylaing (changes requested):
Overall it looks really good :) 

One thing that's missing though are some tests. Is that something you would be able to incorporate?

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,155 @@
...
+
+#. We will use ``cephbr0`` for the Ceph internal traffic and ``cephbr1`` for the Ceph public traffic. In the real life, you'd choose the fast subnet for the internal Ceph traffic::
+
+    What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph public traffic on? [default: 203.0.113.0/24]: 10.0.2.0/24
+
+    Interface "enp7s0" ("10.0.2.3") detected on cluster member "micro2"
+    Interface "enp7s0" ("10.0.2.4") detected on cluster member "micro3"
+    Interface "enp7s0" ("10.0.2.2") detected on cluster member "micro1"
+
+    What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph internal traffic on? [default: 203.0.113.0/24]: 10.0.1.0/24
```
tomponline, line 93: Because `10.0.2.0/24` was specified on the first question, the 2nd question should default to `10.0.2.0/24` as well right?

microcloud/cmd/microcloud/ask.go
```diff
@@ -599,6 +623,104 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
...
+			microCloudInternalNetworkAddr := microCloudInternalSubnet.IP.Mask(microCloudInternalSubnet.Mask)
+			ones, _ := microCloudInternalSubnet.Mask.Size()
+			microCloudInternalNetworkAddrCIDR := fmt.Sprintf("%s/%d", microCloudInternalNetworkAddr.String(), ones)
+
+			publicCephSubnet, err := c.asker.AskString(fmt.Sprintf("What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph public traffic on? [default: %s] ", microCloudInternalNetworkAddrCIDR), microCloudInternalNetworkAddrCIDR, validate.IsNetwork)
+			if err != nil {
+				return err
+			}
+
+			internalCephSubnet, err := c.asker.AskString(fmt.Sprintf("What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph internal traffic on? [default: %s] ", microCloudInternalNetworkAddrCIDR), microCloudInternalNetworkAddrCIDR, validate.IsNetwork)
```
tomponline, line 650: The default value here should be publicCephSubnet and not microCloudInternalNetworkAddrCIDR like we discussed the other day.

tomponline (commented):
@mionaalex @ru-fu should the diagrams in this PR be included or held back until the Design team have made official ones?

doc/explanation/microcloud.rst
```diff
@@ -64,6 +64,18 @@ MicroCloud will still be usable, but you will see some limitations:
   As a result of this, network forwarding works at a basic level only, and external addresses must be forwarded to a specific cluster member and don't fail over.
 - There is no support for hardware acceleration, load balancers, or ACL functionality within the local network.
 
+Dedicated networks for Ceph
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+You can set up dedicated networks for Ceph to differentiate between public and internal Ceph traffic.
+
+- You can choose to have a dedicated network for the public Ceph traffic and another dedicated network for the Ceph internal (cluster) traffic. This is a fully disaggregated Ceph network setup.
```
tomponline, line 71: This could be rephrased to something like as this seemed overlay verbose when I read it:

> You can choose to separate Ceph public and internal traffic onto separate dedicated networks (fully disaggregated) or you can choose to have all Ceph traffic use a single separate network (partially disaggregated).
> In either setups all of your cluster members must have an IP address configured on a network interface for the Ceph network(s) specified.

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,155 @@
...
+    b. Do the same for ``cephbr1`` on each VM::
+
+        # If the `cephbr1` gateway address is `10.0.2.1/24` (subnet should be `10.0.2.0/24`)
+        ip link set enp8s0 up
+        # `X` should be a number between 2 and 254, different for each VM
+        ip addr add 10.0.2.X/24 dev enp8s0
+
+5. Now, you can start the MicroCloud initialisation process and provide the subnets you noted down in step 1.c when asked for the Ceph networking subnets.
+
+#. We will use ``cephbr0`` for the Ceph internal traffic and ``cephbr1`` for the Ceph public traffic. In the real life, you'd choose the fast subnet for the internal Ceph traffic::
```
ru-fu, line 85:
```suggestion
#. We will use ``cephbr0`` for the Ceph internal traffic and ``cephbr1`` for the Ceph public traffic. In a production setup, you'd choose the fast subnet for the internal Ceph traffic::
```
VMs are also real life! :laughing:

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,155 @@
...
+        .. terminal::
+            :input: microceph.ceph config dump
+            :user: root
+            :host: micro1
+            :scroll:
+
+            WHO     MASK  LEVEL     OPTION                       VALUE        RO
+            global        advanced  cluster_network              10.0.1.0/24  *
+            global        advanced  public_network               10.0.2.0/24  *
+            global        advanced  osd_pool_default_crush_rule  2
```
ru-fu, line 112: Too much indentation here as well (this time it's visible because of the line in front of the terminal view).

And the same for the following substeps.

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,155 @@
...
+    Interface "enp7s0" ("10.0.2.2") detected on cluster member "micro1"
+
+    What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph internal traffic on? [default: 203.0.113.0/24]: 10.0.1.0/24
+
+    Interface "enp7s0" ("10.0.1.3") detected on cluster member "micro2"
+    Interface "enp7s0" ("10.0.1.4") detected on cluster member "micro3"
+    Interface "enp7s0" ("10.0.1.2") detected on cluster member "micro1"
+
+7. The MicroCloud initialisation process will now continue as usual and the Ceph cluster will be configured with the networking setup you provided.
+8. You can inspect the Ceph network setup:
```
ru-fu, line 100:
```suggestion
8. You can now inspect the Ceph network setup:
```

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,155 @@
...
+            :user: root
+            :host: micro1
+            :scroll:
+
+            WHO     MASK  LEVEL     OPTION                       VALUE        RO
+            global        advanced  cluster_network              10.0.1.0/24  *
+            global        advanced  public_network               10.0.2.0/24  *
+            global        advanced  osd_pool_default_crush_rule  2
+
+    b. Inspect your Ceph related network traffic:
```
ru-fu, line 114:
```suggestion
    b. Inspect your Ceph-related network traffic:
```

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,155 @@
...
+            17:48:48.604688 IP micro1.33328 > 10.0.1.4.6803: Flags [P.], seq 359774515:359839675, ack 938811, win 1886, options [nop,nop,TS val 3647909568 ecr 3552095031], length 65160
+            17:48:48.604733 IP micro1.33328 > 10.0.1.4.6803: Flags [P.], seq 359839675:359904835, ack 938811, win 1886, options [nop,nop,TS val 3647909568 ecr 3552095031], length 65160
+            17:48:48.604751 IP 10.0.1.4.6803 > micro1.33328: Flags [.], ack 359709355, win 24317, options [nop,nop,TS val 3552095035 ecr 3647909568], length 0
+            17:48:48.604757 IP micro1.33328 > 10.0.1.4.6803: Flags [P.], seq 359904835:359910746, ack 938811, win 1886, options [nop,nop,TS val 3647909568 ecr 3552095035], length 5911
+            17:48:48.604797 IP micro1.33328 > 10.0.1.4.6803: Flags [P.], seq 359910746:359975906, ack 938811, win 1886, options [nop,nop,TS val 3647909568 ecr 3552095035], length 65160
+
+We showed you how to set up a MicroCloud with a fully disaggregated Ceph networking setup, but you can also set up a partially disaggregated Ceph networking setup by using the same network subnet for both the Ceph public and internal traffic.
```
ru-fu, line 155:
```suggestion
This guide showed how to set up a MicroCloud with a fully disaggregated Ceph networking setup, but you can also set up a partially disaggregated Ceph networking setup by using the same network subnet for both the Ceph public and internal traffic.
```

doc/tutorial/get_started.rst
```diff
@@ -385,6 +387,11 @@ See the full initialisation process here:
     Using 1 disk(s) on "micro3" for remote storage pool
 
    Would you like to set up CephFS remote storage? (yes/no) [default=yes]:  yes
+   Configure a dedicated Ceph network? (yes/no) [default=no]: yes
+   Choose either an IPv4 or IPv6 subnet (CIDR notation) to describe your Ceph dedicated cluster: 192.168.0.0/24
+   Interface "enp7s0" ("192.168.0.3") detected on cluster member "micro2"
+   Interface "enp7s0" ("192.168.0.4") detected on cluster member "micro3"
+   Interface "enp7s0" ("192.168.0.2") detected on cluster member "micro1"
```
ru-fu, line 394: Wait, are we configuring it here or not?
Should the answer for line 274 be `no`?

ru-fu:
> @mionaalex @ru-fu should the diagrams in this PR be included or held back until the Design team have made official ones?

I think they are readable enough to include them (and then replace them after we request official ones).

tomponline:
@gabrielmougard have the diagram images been optimized to reduce their size  cc @simondeziel like we do in LXD ones?

ru-fu:
> @gabrielmougard have the diagram images been optimized to reduce their size cc @simondeziel like we do in LXD ones?

See https://documentation.ubuntu.com/lxd/en/latest/contributing/#contribute-to-the-documentation (last sentence).

doc/how-to/ceph_networking.rst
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/explanation/microcloud.rst
```diff
@@ -64,6 +64,18 @@ MicroCloud will still be usable, but you will see some limitations:
   As a result of this, network forwarding works at a basic level only, and external addresses must be forwarded to a specific cluster member and don't fail over.
 - There is no support for hardware acceleration, load balancers, or ACL functionality within the local network.
 
+Dedicated networks for Ceph
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+You can set up dedicated networks for Ceph to differentiate between public and internal Ceph traffic.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/explanation/microcloud.rst
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,155 @@
...
+How to configure Ceph networking
+================================
+
+When running :command:`microcloud init`, you are asked if you want to provide custom subnets for the Ceph cluster.
+Here are the questions you will be asked:
+
+- ``What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph public traffic on? [default: 203.0.113.0/24]: <answer>``
+- ``What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph internal traffic on? [default: 203.0.113.0/24]: <answer>``
+
+You can choose to skip both questions (just hit `Enter`) and use the default value which is the subnet used for the internal MicroCloud traffic.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,155 @@
...
+This is referred to as a *usual* Ceph networking setup.
+
+.. figure:: /images/ceph_network_usual_setup.png
+   :alt: All the Ceph traffic is on the same network interface
+   :align: center
+
+But sometimes, you want to be able to use different network interfaces for some Ceph related usages.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,155 @@
...
+Let's imagine you have machines with network interfaces that are tailored for high throughput and low latency data transfer,
+like 100 GbE+ QSFP links, and other ones that might be more suited for management traffic, like 1 GbE or 10 GbE links.
+
+In this case, it would probably be ideal to set your Ceph internal (or cluster) traffic on the high throughput network interface and the Ceph public traffic on the management network interface. This is referred to as a *fully disaggregated* Ceph networking setup.
+
+.. figure:: /images/ceph_network_full_setup.png
+   :alt: Each type of Ceph traffic is on a different network interface, tailored for its usage.
+   :align: center
+
+You could also decide to put both type of traffic on the same high throughput and low latency network interface. This is referred to as a *partially disaggregated* Ceph networking setup.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,155 @@
...
+
+.. figure:: /images/ceph_network_partial_setup.png
+   :alt: Both the Ceph public and internal traffic are on the same high throughput network interface.
+   :align: center
+
+To use a fully or partially disaggregated Ceph networking setup with your MicroCloud, specify the corresponding subnets during the MicroCloud initialisation process.
+The following instructions build on the :ref:`get-started` tutorial and show how you can test setting up a MicroCloud with disaggregated Ceph networking inside a LXD setup.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,155 @@
...
+
+1. Create the dedicated networks for Ceph:
+    a. First, just like when you created an uplink network for MicroCloud so that the cluster members could have external connectivity, you will need to create a dedicated network for the Ceph cluster members to communicate with each other. Let's call it ``cephbr0``::
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
gabrielmougard: fixed

doc/tutorial/get_started.rst
gabrielmougard: This bit needs to go, the two above questions represent the new usage.

gabrielmougard: fixed

microcloud/cmd/microcloud/ask.go
gabrielmougard: fixed

gabrielmougard:
> > @gabrielmougard have the diagram images been optimized to reduce their size cc @simondeziel like we do in LXD ones?
> 
> See https://documentation.ubuntu.com/lxd/en/latest/contributing/#contribute-to-the-documentation (last sentence).

The images should be compressed now.

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,161 @@
...
+    Interface "enp7s0" ("10.0.1.3") detected on cluster member "micro2"
+    Interface "enp7s0" ("10.0.1.4") detected on cluster member "micro3"
+    Interface "enp7s0" ("10.0.1.2") detected on cluster member "micro1"
+
+7. The MicroCloud initialisation process will now continue as usual and the Ceph cluster will be configured with the networking setup you provided.
+8. You can now inspect the Ceph network setup:
+
+   a. Inspect the Ceph configuration file:
+
+   .. terminal::
```
ru-fu, line 109: This is now too little indentation. ;)
It should have three spaces more (and so be aligned with "Inspect").

microcloud/service/lxd.go
gabrielmougard: There isn't any actually. I implied that we desire net devices that are not receiving broadcast frames but regular L2 net devices. I'm removing this comment. But this function is very similar to the above `GetUplinkInterfaces` function which is also filtering out net devices of type `broadcast`

gabrielmougard: fixed

ru-fu (approved):
The docs look good now (one comment regarding indentation, but the output displays okay-ish as is).

microcloud/cmd/microcloud/main_init.go
```diff
@@ -81,7 +81,7 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
 func (c *cmdInit) RunInteractive(cmd *cobra.Command, args []string) error {
 	// Initially restart LXD so that the correct MicroCloud service related state is set by the LXD snap.
 	fmt.Println("Waiting for LXD to start...")
-	lxdService, err := service.NewLXDService(context.Background(), "", "", c.common.FlagMicroCloudDir)
+	lxdService, err := service.NewLXDService(context.Background(), "", "", c.common.FlagMicroCloudDir, make(map[string]string))
```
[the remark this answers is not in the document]
gabrielmougard: fixed

microcloud/service/lxd.go
gabrielmougard: fixed

microcloud/cmd/microcloud/main_init.go
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
gabrielmougard: fixed

ru-fu (approved):
:+1: for the docs.

gabrielmougard:
@masnax @markylaing updated

microcloud/test/includes/microcloud.sh
```diff
@@ -663,6 +729,84 @@ reset_systems() {
...
+        new_ip=$(generate_ip "${num_hosts}" "${ip_num}")
+        if echo " ${used_ips} " | grep -q " ${new_ip} "; then
+          continue
+        fi
+
+        used_ips="${used_ips} ${new_ip} "
+        break
+      done
+
+      lxc exec "${name}" -- ip addr add "${new_ip}/${range}" dev enp8s0
```
masnax, line 808: All of this is really really complex, and will affect all existing tests, when for most tests it should be fine to use the default values for MicroCeph's cluster and public networks.

Would it be possible to remove all of this and just call the following when running the test?
```bash
# in basic.sh, when running the test:
reset_systems 3 3 3

iface="enp8s0"
ceph_subnet="hardcoded_ip/hardcoded_range"
for m in micro01 micro02 micro03 ; do
  lxc exec "${m}" -- ip addr add "${ceph_subnet}" dev "${iface}"
done

export CEPH_PUBLIC_NETWORK="${ceph_subnet}"
microcloud_interactive | lxc exec micro01 -- sh -c "microcloud init > out"
```

microcloud/test/suites/add.sh
```diff
@@ -99,7 +99,9 @@ test_add_auto() {
 }
 
 test_add_interactive() {
-  reset_systems 4 2 1
+  ceph_public_subnet="10.0.1.0/24"
+  ceph_cluster_subnet="10.0.2.0/24"
+  reset_systems 4 2 4 "${ceph_public_subnet}" "${ceph_cluster_subnet}"
```
masnax, line 104: By default, each machine is only created expecting a maximum of 3 disks and 3 interfaces. If you want to increment this higher (like 4, which you have here), you would have raise the values passed [here](https://github.com/canonical/microcloud/blob/0c6fff36559abb2277e4dcb08d146992a31b8400/microcloud/test/main.sh#L128) to `new_systems 4 3 4`

masnax (changes requested):
Looks good overall, I'd suggest just a couple changes, mainly about duplication and changing the approach for assigning addresses during tests.

microcloud/cmd/microcloud/main_init.go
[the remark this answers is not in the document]
gabrielmougard: fixed

microcloud/service/lxd.go
```diff
@@ -400,6 +400,104 @@ func (s LXDService) GetUplinkInterfaces(ctx context.Context, bootstrap bool, pee
...
+				for _, address := range state.Addresses {
+					if address.Scope != "global" {
+						continue
+					}
+
+					addresses = append(addresses, address.Address)
+				}
+			}
+
+			if len(addresses) == 0 {
```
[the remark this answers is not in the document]
gabrielmougard: fixed

ru-fu:
> > @mionaalex @ru-fu should the diagrams in this PR be included or held back until the Design team have made official ones?
> 
> I think they are readable enough to include them (and then replace them after we request official ones).

Just FYI - I've requested the diagrams, target date is May 10.
(And as I said before, this shouldn't prevent the PR from being merged.)

microcloud/cmd/microcloud/main_init_preseed.go
```diff
@@ -648,6 +680,90 @@ func (p *Preseed) Parse(s *service.Handler, bootstrap bool) (map[string]InitSyst
...
+			}
+
+			if existingCephPublicSubnet != "" {
+				err = validateCephInterfacesForSubnet(lxd, systems, cephInterfaces, existingCephPublicSubnet)
+				if err != nil {
+					return nil, err
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
gabrielmougard: fixed

microcloud/test/includes/microcloud.sh
```diff
@@ -23,6 +23,8 @@ microcloud_interactive() {
   CEPH_WARNING=${CEPH_WARNING:-}                 # (yes/no) input for warning about eligible disk detection.
   CEPH_FILTER=${CEPH_FILTER:-}                   # filter string for CEPH disks.
   CEPH_WIPE=${CEPH_WIPE:-}                       # (yes/no) to wipe all disks.
+  CEPH_PUBLIC_NETWORK=${CEPH_PUBLIC_NETWORK:-}   # (default: MicroCloud internal subnet) input for setting up a public network.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

microcloud/test/suites/add.sh
gabrielmougard: fixed

microcloud/test/includes/microcloud.sh
gabrielmougard: fixed

gabrielmougard:
@masnax sometimes in the preseed.sh test suite I see an `Failed to find at least 2 disks for filter "id == sdd"` error. And sometimes when I re run the build, the test goes through. Do you know what might happen here?

simondeziel:
> @masnax sometimes in the `preseed.sh` test suite I see an `Failed to find at least 2 disks for filter "id == sdd"` error. And sometimes when I re run the build, the test goes through. Do you know what might happen here? (I don't see the option for re running a part of the workflow in github btw)

Might be unrelated but I've seen a similar issue where the validation is expecting a specific ZFS disk to be there but the VM is missing it on some runs.

> Aside from that, I'm waiting for the `basic.sh` test suite to complete.

On some GHA runners (presumably with noisy neighbors) this one can take ~forever until it eventually times out after 6 hours :/

masnax:
If the validation is reporting that the disk is missing, that means it wasn't properly wiped and still has partitions from the previous run. 

I've noticed `zpool destroy` (and microceph teardown in the non-snapshot restore version) will occasionally not let go of the disk and so even if we run `wipefs -a` the partitions remain.

microcloud/test/includes/microcloud.sh
```diff
@@ -74,6 +77,35 @@ EOF
 )
 fi
 
+if [ -z "${IGNORE_CEPH_NETWORKING}" ]; then
+  if [ -n "${CEPH_PUBLIC_NETWORK}" ]; then
+    setup=$(cat << EOF
+${setup}
+${CEPH_PUBLIC_NETWORK}                                  # setup public network
+EOF
+)
```
simondeziel, line 86: Would you mind using the following equivalent syntax?
```suggestion
    setup="${setup}
${CEPH_PUBLIC_NETWORK}"                                 # setup public network
)
```

I'd put a PR doing the same for the existing code. The reason for this style change is that it produces less debug (`set -x`) noise:

```
$ sh -x
$ setup=""
+ setup=
$ setup="${setup}
> foo"
+ setup=
foo
$ setup="${setup}
> bar"
+ setup=
foo
bar
$ 
```

Versus:

```
$ sh -x
$ setup=""   
+ setup=
$ setup=$(cat << EOF
> ${setup}
> foo
> EOF
> )
+ cat
+ setup=
foo
$ setup=$(cat << EOF
> ${setup}
> bar
> EOF
> )
+ cat
+ setup=
foo
bar
$ 
```
```

Comment (gabrielmougard) [reply microcloud/test/includes/microcloud.sh]: fixed

**Comment by gabrielmougard** [comment]
@simondeziel @masnax the `interactive_combinations` fail a little bit closer to the end each time I'm re-triggering the test in GitHub (this time it goes to 4 machines, two disks, one extra interface which is one test case away from the end). I can't really reproduce this issue locally as I don't experience this 'LXD socket connection issue' .. what should we do ?

**Comment by masnax** [comment]
> @simondeziel @masnax the `interactive_combinations` fail a little bit closer to the end each time I'm re-triggering the test in GitHub (this time it goes to 4 machines, two disks, one extra interface which is one test case away from the end). I can't really reproduce this issue locally as I don't experience this 'LXD socket connection issue' .. what should we do ?

I think the 4 machine tests might just be too hard on the github runners?

**Comment by simondeziel** [comment]
> > @simondeziel @masnax the `interactive_combinations` fail a little bit closer to the end each time I'm re-triggering the test in GitHub (this time it goes to 4 machines, two disks, one extra interface which is one test case away from the end). I can't really reproduce this issue locally as I don't experience this 'LXD socket connection issue' .. what should we do ?
> 
> I think the 4 machine tests might just be too hard on the github runners?

I'm working on a PR that will hopefully make the VM a bit faster to boot but yeah, we are apparently pushing the envelop a bit it seems.

File: microcloud/cmd/microcloud/main_init_preseed.go [hunk 48]
```diff
@@ -653,6 +685,47 @@ func (p *Preseed) Parse(s *service.Handler, bootstrap bool) (map[string]InitSyst
 		systems[peer] = system
 	}
 
+	// Configure Ceph networks.
+	infos := make([]mdns.ServerInfo, 0, len(systems))
+	for _, system := range systems {
+		infos = append(infos, system.ServerInfo)
+	}
+
+	cephInterfaces, err := lxd.GetCephInterfaces(context.Background(), bootstrap, infos)
```
masnax, line 694: For a bit of efficiency, should we wrap this in a `if p.Ceph.PublicNetwork != "" || p.Ceph.InternalNetwork != "" || !bootstrap`?

microcloud/test/includes/microcloud.sh
```diff
@@ -74,6 +77,23 @@ $(true)                                                 # workaround for set -e
 "
 fi
 
+if [ -z "${IGNORE_CEPH_NETWORKING}" ]; then
+  if [ -n "${CEPH_PUBLIC_NETWORK}" ]; then
+    setup="${setup}
+${CEPH_PUBLIC_NETWORK}"                                 # setup public network
+  else
+    setup="${setup}
+_ENTER_"
```
masnax, line 86: I'm guessing the `_ENTER_` is to just select the default value that MicroCloud will find? I think that might make it hard to validate what the outcome is.

I think for the test suite, we should always explicitly enter values so that we can pass the expected value to the `validate_microceph` function. 

One thing that might be useful in a future PR is an additional test of MicroCloud's default values in general, probably by parsing the `out` file which contains the stdout from `microcloud init`.

masnax (approved):
Looking good, just a few more small nits and I think we should be able to merge soon!

masnax (changes requested):
Oops, accidentally pressed "Approve" when I meant to press "Request changes" :)

simondeziel:
> I'm working on a PR that will hopefully make the VM a bit faster to boot but yeah, we are apparently pushing the envelop a bit it seems.

https://github.com/canonical/microcloud/pull/298 is ready for review and did improve things for me in the repeated runs I've done.

doc/images/ceph_network_full_setup.png
ru-fu, line 1: We got the updated images now - please update them here before merging. Thanks!

ru-fu (changes requested):

microcloud/service/microovn.go
```diff
@@ -130,3 +132,8 @@ func (s OVNService) Address() string {
 func (s OVNService) Port() int64 {
 	return s.port
 }
+
+// SetConfig sets the config of this Service instance.
+func (s *OVNService) SetConfig(config map[string]string) {
+	s.config = config
```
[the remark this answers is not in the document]
gabrielmougard: fixed

microcloud/service/lxd.go
```diff
@@ -398,6 +407,78 @@ func (s LXDService) GetUplinkInterfaces(ctx context.Context, bootstrap bool, pee
...
+			addresses := []string{}
+			if network.Type != "bridge" {
+				for _, address := range state.Addresses {
+					if address.Scope != "global" {
+						continue
+					}
+
+					addresses = append(addresses, address.Address)
+				}
+			}
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/images/ceph_network_full_setup.png
gabrielmougard: fixed

gabrielmougard:
@simondeziel I don't understand the new `microcloud_interactive` system in the tests. I keep having `Invalid input, try again.` when I want to pass my two ceph network addresses. Do you see what I might be doing wrong?

simondeziel:
> @simondeziel I don't understand the new `microcloud_interactive` system in the tests. I keep having `Invalid input, try again.` when I want to pass my two ceph network addresses. Do you see what I might be doing wrong?

I'm not sure what's the expected answer list would look like but if we that was in the last failed run:

```
echo '
enp5s0                                         # filter the lookup interface
select          # select the interface
---
yes                                             # limit lookup subnet (yes/no)
  # skip MicroOVN/MicroCeph (yes/no)
expect 2                                      # wait until the systems show up
select-all                                                  # select all the systems
---
                                                 # workaround for set -e

yes                                            # add local disks (yes/no)
wait 300ms    # wait for the table to populate
lxd_disk1                                           # filter zfs disks
select-all    # select all disk matching the filter
---
select-all    # wipe all disks
---
                                                 # workaround for set -e

yes                                           # add remote disks (yes/no)
                                         # continue with some peers missing disks? (yes/no)
wait 300ms   # wait for the table to populate
lxd_disk2                                          # filter ceph disks
select-all   # select all disk matching the filter
---
select-all   # wipe all disks
---
yes
                                                 # workaround for set -e

10.0.1.0/24
                                                 # workaround for set -e

10.0.1.0/24
                                                 # workaround for set -e

yes                                           # agree to setup OVN
                                         # continue with some peers missing an interface? (yes/no)
wait 300ms   # wait for the table to populate
enp6s0                                          # filter interfaces
select-all   # select all interfaces matching the filter
---
10.1.123.1/24                                         # setup ipv4/ipv6 gateways and ranges
10.1.123.100
10.1.123.254
fd42:1:1234:1234::1/64
10.1.123.1,8.8.8.8
                                                 # workaround for set -e
' | sed '/^\s*#/d; s/\s*#.*//; /^$/d'
```

We get that:
```
enp5s0
select
---
yes
expect 2
select-all
---
yes
wait 300ms
lxd_disk1
select-all
---
select-all
---
yes
wait 300ms
lxd_disk2
select-all
---
select-all
---
yes
10.0.1.0/24
10.0.1.0/24
yes
wait 300ms
enp6s0
select-all
---
10.1.123.1/24
10.1.123.100
10.1.123.254
fd42:1:1234:1234::1/64
10.1.123.1,8.8.8.8
```

I don't have any better way to debug this than redoing the `echo | sed` manually to sanity check it. HTH

simondeziel:
@gabrielmougard rebasing now that #298 was merged should (fingers crossed) make your test go a bit faster.

doc/how-to/ceph_networking.rst
```diff
@@ -0,0 +1,161 @@
...
+Here are the questions you will be asked:
+
+- ``What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph public traffic on? [default: 203.0.113.0/24]: <answer>``
+- ``What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph internal traffic on? [default: 203.0.113.0/24]: <answer>``
+
+You can choose to skip both questions (just hit ``Enter``) and use the default value which is the subnet used for the internal MicroCloud traffic.
+This is referred to as a *usual* Ceph networking setup.
+
+.. figure:: /images/ceph_network_usual_setup.svg
+   :alt: All the Ceph traffic is on the same network interface
```
ru-fu, line 16: Can you add `:width: 100%` here (and below)?
That *should* make the images clickable.

masnax (changes requested):
So it seems in your tests, you're assigning a subnet for the ceph public/cluster questions each time, but in many of the tests the systems aren't set up with any interfaces configured to have those subnets. So MicroCloud is detecting there's no available subnets and skips the ceph public/cluster questions.

Depending on the test, you may want to just leave those variables blank, or properly apply the subnet you want to each system's interface.

microcloud/test/includes/microcloud.sh
```diff
@@ -487,7 +529,7 @@ reset_snaps() {
 # Makes only `num_disks` and `num_ifaces` disks and interfaces available for the next test.
 reset_system() {
   if [ "${SNAPSHOT_RESTORE}" = 1 ]; then
-    restore_system ${*}
+    restore_system "${*}"
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/how-to/ceph_networking.rst
gabrielmougard: fixed

ru-fu (dismissed):
The doc updates look good now. Thanks!

microcloud/test/suites/basic.sh
```diff
@@ -3,13 +3,17 @@
...
+
   echo "Creating a MicroCloud with all services but no devices"
   export LOOKUP_IFACE="enp5s0"
   export LIMIT_SUBNET="yes"
   export EXPECT_PEERS=2
   export SETUP_ZFS="no"
   export SETUP_CEPH="no"
   export SETUP_OVN="no"
+  export CEPH_PUBLIC_NETWORK="${microcloud_internal_net_addr}" # Default to MicroCloud internal network
+  export CEPH_CLUSTER_NETWORK="${microcloud_internal_net_addr}"
```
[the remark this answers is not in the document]
gabrielmougard: fixed

gabrielmougard:
@masnax do you think it should be mergeable now?

cmd/microcloud/ask.go
```diff
@@ -565,7 +647,7 @@ func (c *CmdControl) askRemotePool(systems map[string]InitSystem, autoSetup bool
 	return nil
 }
 
-func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSystem, autoSetup bool) error {
```
masnax, line 568: Have you tested if the modifications to the network questions will correctly be skipped if MicroCeph is not installed?

cmd/microcloud/ask.go
```diff
@@ -589,17 +671,82 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
...
+				if err != nil {
+					return err
+				}
+
+				bootstrapSystem := systems[sh.Name]
+				bootstrapSystem.MicroCephInternalNetworkSubnet = internalCephSubnet
+				systems[sh.Name] = bootstrapSystem
+			}
+		} else {
+			err = validateAddingMemberWithCephNetwork(sh, lxdService, systems, availableCephNetworkInterfaces, microCloudInternalSubnet.String())
```
masnax, line 738: Since we can be re-using an existing MicroCeph cluster now, we can't just rely on the unix socket anymore.

For example, lets say you want to set up MicroCloud with 3 systems.
- system1: all snaps are pristine
- system2: microceph has been bootstrapped, and is part of a 10-system cluster unrelated to microcloud
- system3: all snaps are pristine

In this case, we should end up with a 3-system MicroCloud/LXD/MicroOVN, and a 13 system MicroCeph.

But that means you need to connect to the existing (remote) MicroCeph deployment by specifying the address and mDNS auth secret of the remote system to the `Client` function.

You can grab the info of which systems are initialized from the `InitializedServices` map in `InitSystem`. You can look at this block to see how we handle something like this for issuing join tokens remotely if one of our systems is already clustered.
https://github.com/canonical/microcloud/blob/main/cmd/microcloud/main_init.go#L379-L435

masnax (changes requested):
@gabrielmougard Looking good, just a couple edge cases to consider.

cmd/microcloud/ask.go
masnax: To clarify, I'm talking about fetching the cluster config from MicroCeph.

masnax: In your case, since this endpoint already exists over the network in MicroCeph, you don't have to create a manual endpoint like we did for the join tokens, it would suffice to just specify the address & secret of system2 to the Client function.

cmd/microcloud/ask.go
```diff
@@ -589,17 +671,82 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
 		return nil
 	}
 
-	// Environments without OVN get a basic fan setup.
-	if sh.Services[types.MicroOVN] == nil {
```
masnax, line 593: So I think around here you'll need to check if MicroCeph is installed and only go through the new implementation if it is installed.

gabrielmougard: fixed

cmd/microcloud/ask.go
gabrielmougard: Understood! What if during the add, multiple systems are part of different MicroCeph clusters? If this is even possible, here is the validation process I had in mind:
```go
func validateAddingMemberWithCephNetwork(sh *service.Handler, lxd *service.LXDService, systems map[string]InitSystem, cephInterfaces map[string][]service.CephDedicatedInterface, existingMicroCloudInternalSubnet string) error {
	// In case of new machines added to the cluster that will be part of an existing Ceph cluster,
	// and if a dedicated Ceph cluster network has been configured, we need to ensure that the new machines
	// have at least one network interface with an IP within the existing Ceph network subnet.
	microCephService := sh.Services[types.MicroCeph].(*service.CephService)
	if microCephService != nil {
		// Separate the systems that have been initialized with the MicroCeph service from the ones that haven't.
		// There might be multiple 'remote' MicroCeph clusters to validate against because among the systems to be added,
		// there might be some that have already been initialized with the MicroCeph service. See the example below:
		//
		// 			    MicroCeph cluster A
		// 		            +------------------------------+
		// 			    |                              |
		// 			    |                              |
		// 			    |                              |
		// 			    |                +-------+     |
		// 		            |        +-------+       |     |
		// 			    |        |       |       |     |
		// 			    |        |       +-------+     |
		// 			    |        |       +-------+     |
		// 			    |        | +-----+       |     |
		// 			    |        | |     |       |     |
		// Existing MicroCloud      |    +---+-++    +-------+     |
		// +---------+              |    | ADD  |    +-------+     |
		// |         |    +-------------->      +----+       |     |
		// | +-----+ |    |         |    +------+    |       |     |
		// | |     | |    |         |                +-------+     |
		// | |     | |    |         +------------------------------+
		// | +-----+ |    |
		// |         |    |          Pristine node
		// | +-----+ |    |          +------+
		// | |  SH +-----------------> ADD  |
		// | |     | |    |          |      |
		// | +-----+ |    |          +------+
		// |         |    |
		// | +-----+ |    |         MicroCeph cluster B
		// | |     | |    |         +--------------------------------+
		// | |     | |    |         |                                |
		// | +-----+ |    |         |    +------+          +------+  |
		// |         |    +--------------> ADD  +----------+      |  |
		// +---------+              |    |      |          |      |  |
		// 	                    |    +----+-+          +------+  |
		// 	                    |         |                      |
		// 	                    |         |            +------+  |
		// 	                    |         +------------+      |  |
		// 	                    |                      |      |  |
		// 	                    |                      +------+  |
		// 	                    |                                |
		// 	                    +--------------------------------+
		//
		// * In such a scenario, the 'initialized' nodes to be added (with their remote Ceph cluster configuration),
		//   will be validated one by one against the existing MicroCloud with a Ceph deployment.
		// * The 'uninitialized' nodes will be validated all together against the existing MicroCloud with a Ceph deployment.

		initializedMicroCephSystems := make(map[string]InitSystem)
		uninitializedMicroCephSystems := make(map[string]InitSystem)
		for peer, system := range systems {
			if system.InitializedServices[types.MicroCeph][peer] != "" {
				initializedMicroCephSystems[peer] = system
			} else {
				uninitializedMicroCephSystems[peer] = system
			}
		}

		checkCephConfigs := func(systems map[string]InitSystem, initialized bool) error {
			var targetAddress string
			var targetAuthSecret string
			for peer, system := range systems {
				if initialized {
					targetAddress = system.ServerInfo.Address
					targetAuthSecret = system.ServerInfo.AuthSecret
				}

				cephConfigs, err := microCephService.ClusterConfig(context.Background(), targetAddress, targetAuthSecret)
				if err != nil {
					return err
				}

				// Read the Ceph network configuration from the existing Ceph cluster
				existingCephInternalSubnet := ""
				existingCephPublicSubnet := ""
				for key, value := range cephConfigs {
					if key == "cluster_network" && value != "" {
						// Sometimes, the default cluster_network value in the Ceph configuration
						// is not a network range but a regular IP address. We need to extract the network range.
						_, valueNet, err := net.ParseCIDR(value)
						if err != nil {
							return fmt.Errorf("failed to parse the Ceph cluster network configuration from the existing Ceph cluster: %v", err)
						}

						if valueNet.String() != existingMicroCloudInternalSubnet {
							existingCephInternalSubnet = valueNet.String()
						}
					}

					if key == "public_network" && value != "" {
						_, valueNet, err := net.ParseCIDR(value)
						if err != nil {
							return fmt.Errorf("failed to parse the Ceph public network configuration from the existing Ceph cluster: %v", err)
						}

						if valueNet.String() != existingMicroCloudInternalSubnet {
							existingCephPublicSubnet = valueNet.String()
						}
					}
				}

				if initialized {
					if existingCephInternalSubnet != "" {
						err = validateCephInterfacesForSubnet(lxd, map[string]InitSystem{peer: system}, cephInterfaces, existingCephInternalSubnet)
						if err != nil {
							return err
						}
					}

					if existingCephPublicSubnet != "" {
						err = validateCephInterfacesForSubnet(lxd, map[string]InitSystem{peer: system}, cephInterfaces, existingCephPublicSubnet)
						if err != nil {
							return err
						}
					}
				} else {
					if existingCephInternalSubnet != "" {
						err = validateCephInterfacesForSubnet(lxd, systems, cephInterfaces, existingCephInternalSubnet)
						if err != nil {
							return err
						}
					}

					if existingCephPublicSubnet != "" {
						err = validateCephInterfacesForSubnet(lxd, systems, cephInterfaces, existingCephPublicSubnet)
						if err != nil {
							return err
						}
					}

					// The uninitialized nodes can be validated all together as there is no remote Ceph cluster configuration to validate against,
					// so we can break the loop here.
					break
				}
			}

			return nil
		}

		err := checkCephConfigs(uninitializedMicroCephSystems, false)
		if err != nil {
			return err
		}

		if len(initializedMicroCephSystems) > 0 {
			err = checkCephConfigs(initializedMicroCephSystems, true)
			if err != nil {
				return err
			}
		}

		return nil
	}

	return nil
}
```

masnax: > @masnax Understood! What if during the add, multiple systems are part of different MicroCeph clusters? 

We can't reconcile 2 different clusters for the same service, so the supported configurations are strictly like this:

* All systems are unclustered before running `microcloud init`
* A subset (so 1 or more, up to all) of the systems being added to the MicroCloud are part of the same pre-existing cluster for MicroCeph or MicroOVN
  * The subset contains external systems we don't care about in MicroCloud. These can be ignored.

So the key here is that there will only ever be 0 or 1 external clusters for each service, but the participants can vary. Additionally, we only need to support this for `microcloud init`, not `microcloud add` since if running `microcloud add` the system we are interacting with must necessarily be a cluster member already.

So for this I think the logical flow should look something like this: 

* If we are bootstrapping: 
  * We check if there are any initialized systems for MicroCeph. If there is then we save the first one we encounter to issue remote requests to. We can assume if there is more than 1 that they are part of the same cluster.
  * We target the remote MicroCeph and fetch its cluster config. If it has not set up cluster/public networks, then we ask the questions. If it has set up cluster/public networks, then we try to see if the systems joining that cluster have the same subnets available.
* If we are not bootstrapping (so adding):
  * We can assume `systems[sh.Name]` (ourself) is clustered.
  * We target the local MicroCeph and fetch its cluster config. If it has not set up cluster/public networks, then we ask the questions. If it has set up cluster/public networks, then we try to see if the systems joining that cluster have the same subnets available.


I noticed currently you don't ask any questions in the `add` case. It might just be my misunderstanding, but what is MicroCeph's default behaviour if cluster/public networks have not been manually configured? Does it pick some defaults? If so then we probably don't need to ask questions in those cases (both for bootstrap with a pre-existing cluster & all add cases). But if the configs are unset then we should probably ask questions to set them.


Hope that's clear, let me know if it's confusing :)

gabrielmougard: Very clear! thanks!

gabrielmougard: To answer your question regarding the `add` case not having any questions:

The logic is to only validate that the to-be-added systems have network interfaces falling within the same subnet as the one configured in the cluster. If no ceph network have been setup in the cluster, this validation process is skipped.

The default behavior of MicroCeph if no cluster/public network have been configured is to use the MicroCloud internal network.

gabrielmougard:
@simondeziel there seems to be an issue with setting up ZFS in the tests: `Error: Error loading "zfs" module: Failed to run: modprobe -b zfs: exit status 126 (nsenter: failed to execute /bin/: Permission denied)`

cmd/microcloud/ask.go
masnax: I see, so if I `microceph cluster bootstrap`, it will automatically set some value for `cluster_network` and `public_network`?

simondeziel:
> @simondeziel @masnax there seems to be an issue with setting up ZFS in the tests: `Error: Error loading "zfs" module: Failed to run: modprobe -b zfs: exit status 126 (nsenter: failed to execute /bin/: Permission denied)`

Sorry for that, I broke the `latest/edge` snap with https://github.com/canonical/lxd-pkg-snap/pull/456. The fix is on the way in https://github.com/canonical/lxd-pkg-snap/pull/459

service/lxd.go
```diff
@@ -333,10 +333,10 @@ func (s *LXDService) GetResources(ctx context.Context, target string, address st
 	return client.GetServerResources()
 }
 
-// GetUplinkInterfaces returns a map of peer name to slice of api.Network that may be used with OVN.
-func (s LXDService) GetUplinkInterfaces(ctx context.Context, bootstrap bool, peers []mdns.ServerInfo) (map[string][]api.Network, error) {
-	clients := map[string]lxd.InstanceServer{}
-	networks := map[string][]api.Network{}
+// getNetworkInterfaces fetches network candidates based on peer information and filtering rules contained in the collector function.
+func getNetworkInterfaces[T api.Network | CephDedicatedInterface](s LXDService, ctx context.Context, bootstrap bool, peers []mdns.ServerInfo, collectFn func(c map[string]lxd.InstanceServer, acc map[string][]T, peer string, nets []api.Network)) (map[string][]T, error) {
```
markylaing, line 337: Do we need to have this much complexity here? I.e. would it be simpler to just have a different function when we want to return a different type?

markylaing (commented):
Looks great :smiley:. Lots of docs and tests. I just have one comment and one nit.

cmd/microcloud/ask.go
masnax: @gabrielmougard looks like the tests are not passing, something to do with the subnet being `nil`.

Also, could you please confirm the question I asked In my [previous comment](https://github.com/canonical/microcloud/pull/274#discussion_r1612038549)?

service/lxd.go
gabrielmougard: yes, I can try to simplify!

gabrielmougard: fixed

cmd/microcloud/ask.go
gabrielmougard: yes

service/lxd.go
markylaing: Did you push your changes? This still looks the same to me.

test/includes/microcloud.sh
```diff
@@ -1104,3 +1146,30 @@ lxd_wait_vm() {
...
+  line=$(ifconfig -a "$1" | grep netmask | tr -s " ")
+  ip=$(echo "$line" | cut -f 3 -d " ")
+  mask=$(echo "$line" | cut -f 5 -d " ")
+
+	IFS=. read -r io1 io2 io3 io4 <<< "$ip"
+	IFS=. read -r mo1 mo2 mo3 mo4 <<< "$mask"
+	net_addr="$((io1 & mo1)).$((io2 & mo2)).$((io3 & mo3)).$((io4 & mo4))"
+
+	echo "${net_addr}$(ip_prefix_by_netmask "${mask}")"
+}
```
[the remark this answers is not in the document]
gabrielmougard: fixed

masnax:
@gabrielmougard 

I seem to be consistently seeing this error at the same spot: https://github.com/canonical/microcloud/actions/runs/9347254250/job/25868665575?pr=274#step:13:9857

```
Error: failed to fetch cluster config: Get "http://control.socket/1.0/services/microceph/1.0/configs?target=10.186.56.49": EOF, Key: 
```

gabrielmougard:
@masnax tests should be passing

service/lxd.go
```diff
@@ -481,6 +481,47 @@ func (s LXDService) GetCephInterfaces(ctx context.Context, bootstrap bool, peers
...
+// and returns a map of peer name to interfaces that are in the subnet.
+func (s *LXDService) ValidateCephInterfaces(cephNetworkSubnetStr string, interfacesMap map[string][]CephDedicatedInterface) (map[string][][]string, error) {
+	_, subnet, err := net.ParseCIDR(cephNetworkSubnetStr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid CIDR subnet: %v", err)
+	}
+
+	ones, bits := subnet.Mask.Size()
+	if bits-ones == 0 {
+		return nil, fmt.Errorf("invalid Ceph network subnet (must be a subnet with at least one address)")
```
markylaing, line 494: What is the convention in microcloud for capitalisation of errors? In LXD they are all capitalised.

Also, if `bits-ones == 0`, wouldn't there be exactly one address? (e.g. `10.123.123.1/32`). Should we instead have `Invalid Ceph network subnet (must have more than one address)`?

service/lxd.go
```diff
@@ -481,6 +481,47 @@ func (s LXDService) GetCephInterfaces(ctx context.Context, bootstrap bool, peers
...
+	}
+
+	data := make(map[string][][]string)
+	for peer, interfaces := range interfacesMap {
+		for _, iface := range interfaces {
+			for _, addr := range iface.Addresses {
+				ip := net.ParseIP(addr)
+				if ip == nil {
+					return nil, fmt.Errorf("invalid IP address: %v", addr)
```
markylaing, line 503: Error capitalisation.

cmd/microcloud/main_init.go
```diff
@@ -28,20 +28,28 @@ import (
...
+	// TargetStoragePools contains the storage pool configuration for the target system.
+	TargetStoragePools []lxdAPI.StoragePoolsPost
+	// Networks is the cluster-wide network configuration.
+	Networks []lxdAPI.NetworksPost
+	// StoragePools is the cluster-wide storage pool configuration.
+	StoragePools []lxdAPI.StoragePoolsPost
+	// StorageVolumes is the cluster-wide storage volume configuration.
+	StorageVolumes map[string][]lxdAPI.StorageVolumesPost
+	// JoinConfig is the LXD configuration for joining members.
+	JoinConfig []lxdAPI.ClusterMemberConfigKey
```
markylaing, line 52: Could you add the new field in on commit and then fix the field comments in a separate commit?

doc/.wordlist.txt
```diff
@@ -37,3 +37,7 @@ Jira
 VM
 YAML
 CephFS
+disaggregated
+subnets
```
markylaing, line 41: @ru-fu I'm not sure why these need to be added to the wordlist. Where does sphinx/spellcheck get it's default wordlist from?

markylaing (changes requested):
A few nits but other than that it looks good to go :)

doc/.wordlist.txt
ru-fu: It's using aspell. Not sure why it doesn't know "disaggregated" (it's not the most common word though ;) ). Unfortunately, aspell doesn't understand plurals for added words, that's why "subnets" must be added manually.

That being said, these additions should all go into `.custom_wordlist.txt` though instead of `.wordlist.txt` @gabrielmougard .

markylaing: Ah fair enough, thanks :)

service/lxd.go
```diff
@@ -11,7 +11,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/canonical/lxd/client"
+	lxd "github.com/canonical/lxd/client"
```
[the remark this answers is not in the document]
gabrielmougard: fixed

service/lxd.go
gabrielmougard: fixed

service/lxd.go
gabrielmougard: fixed

cmd/microcloud/main_init.go
gabrielmougard: fixed

doc/.wordlist.txt
gabrielmougard: fixed

markylaing (approved):
LGTM thanks :)

masnax (approved):
We can finally put this one to rest :)

PR #276, opened 2024-04-03
Update gomod

MggMuggins:
I chose to cast all the parameters of `util.CanonicalNetworkAddress` to the larger `int64` rather than converting all port variables to `int64` (as in https://github.com/canonical/lxd/pull/13078) as it prevents needing to cast to a smaller int when [calling into `mdns`](https://github.com/canonical/microcloud/blob/306018db715ce129c0e19a88ffcaa799c7b60245/microcloud/mdns/mdns.go#L26). That said I'm not attached to one approach or another.


roosterfish (dismissed):
LGTM!

However I am wondering if there could ever be a port number bigger than the 16 bit unsigned int? Is it that just on some systems the underlying data type is different?
If so wouldn't it be more straightforward to use `int64` as type for the port number and convert this number to an int only for the mDNS library?
We could do some checks against `math.MaxInt` for the upper limit before we try to convert the port number.

tomponline:
I would tend to agree with this idea of only converting where you need it.

MggMuggins:
To clarify @tomponline, you would prefer that `int64` be used everywhere in microcloud as opposed to `int` for ports?

tomponline:
> To clarify @tomponline, you would prefer that `int64` be used everywhere in microcloud as opposed to `int` for ports?

Yeah I think that makes sense - as it avoids the variable size of int on non-64 bit platforms (even though in practice it'll most commonly be on 64 bit platforms).

microcloud/service/service_handler.go
```diff
@@ -148,7 +149,11 @@ func (s *Handler) Broadcast() error {
 				return err
 			}
 
-			server, err := cloudMDNS.NewBroadcast(info.LookupKey(), iface, info.Address, s.Port, service, bytes)
+			if s.Port > math.MaxInt {
+				return fmt.Errorf("Invalid port number %q for service %q", s.Port, s.Name)
```
roosterfish, line 153: Should we return more concrete information like "out of range" or something like this?

microcloud/go.mod
```diff
@@ -1,20 +1,20 @@
 module github.com/canonical/microcloud/microcloud
 
-go 1.22
+go 1.22.0
```
roosterfish, line 3: Not a blocker, just wanted to know why we update to `1.22.0` here?

simondeziel: When enabling CodeQL, I noticed it would choke on that short `1.22` version string: https://github.com/canonical/microcloud/actions/runs/8576568978/job/23507751328#step:4:90

```sh
Attempting to automatically build go code
  /opt/hostedtoolcache/CodeQL/2.17.0/x64/codeql/go/tools/autobuild.sh 
  2024/04/05 22:48:49 Autobuilder was built with go1.22.1, environment has go1.21.8
  2024/04/05 22:48:49 LGTM_SRC is /home/runner/work/microcloud/microcloud
  2024/04/05 22:48:49 Found no go.work files in the workspace; looking for go.mod files...
  2024/04/05 22:48:49 Found 1 go.mod file(s).
  2024/04/05 22:48:49 Import path is 'github.com/canonical/microcloud'
  go: downloading go1.22 (linux/amd64)
  go: download go1.22 for linux/amd64: toolchain not available
```

Just adding that `.0` made it happy.

roosterfish: Ah ok, thanks for the hint!

microcloud/service/service_handler.go
```diff
@@ -148,7 +149,11 @@ func (s *Handler) Broadcast() error {
 				return err
 			}
 
-			server, err := cloudMDNS.NewBroadcast(info.LookupKey(), iface, info.Address, s.Port, service, bytes)
+			if s.Port > math.MaxInt {
```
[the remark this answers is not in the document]
simondeziel: Since we are dealing with a signed source number, wouldn't it be better to check for `s.Port < 1` then?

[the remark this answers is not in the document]
MggMuggins: Yes :slightly_smiling_face:

simondeziel (approved):

markylaing (approved):
LGTM. Thanks :)

roosterfish (approved):
Thanks, looking good!

Issue #210, opened 2023-11-10
`microcloud init` does not perform sanity checks on the uplink network input


IsaacJT:
`microcloud init` does not perform many sanity checks on the inputs for configuring the OVN uplink network. I can identify two cases where the user can easily make a mistake while entering the network configuration, which causes the init to fail at the last stage, requiring that the user starts again.

Case 1: first and last addresses are outside the gateway's subnet 

```
$ sudo microcloud init
Waiting for LXD to start...
Using address "10.42.0.49" for MicroCloud
Limit search for other MicroCloud servers to 10.42.0.49/24? (yes/no) [default=yes]:
Scanning for eligible servers ...

 Selected "orangepi-1" at "10.42.0.198"
 Selected "orangepi-2" at "10.42.0.49"
 Selected "orangepi-3" at "10.42.0.217"

Would you like to set up local storage? (yes/no) [default=yes]: no
Would you like to set up distributed storage? (yes/no) [default=yes]:
Select from the available unpartitioned disks:

Select which disks to wipe:

 Using 1 disk(s) on "orangepi-2" for remote storage pool
 Using 1 disk(s) on "orangepi-3" for remote storage pool
 Using 1 disk(s) on "orangepi-1" for remote storage pool

Configure distributed networking? (yes/no) [default=yes]:
Select exactly one network interface from each cluster member:

 Using "enP4p65s0" on "orangepi-2" for OVN uplink
 Using "enP4p65s0" on "orangepi-3" for OVN uplink
 Using "enP4p65s0" on "orangepi-1" for OVN uplink

Specify the IPv4 gateway (CIDR) on the uplink network (empty to skip IPv4): 10.1.1.1/24
Specify the first IPv4 address in the range to use with LXD: 10.2.2.50
Specify the last IPv4 address in the range to use with LXD: 10.2.2.100
Specify the IPv6 gateway (CIDR) on the uplink network (empty to skip IPv6):
Initializing a new cluster
 Local MicroCloud is ready
 Local LXD is ready
 Local MicroOVN is ready
 Local MicroCeph is ready
Awaiting cluster formation ...
 Peer "orangepi-3" has joined the cluster
 Peer "orangepi-1" has joined the cluster
Configuring cluster-wide devices ...
Error: Failed allocating uplink port IPs on network "UPLINK": Failed to parse uplink IPv4 OVN ranges: IP range "10.2.2.50-10.2.2.100" does not fall within any of the allowed networks [10.1.1.0/24]
```

[...]

PR #277, opened 2024-04-03
Network sanity checks

MggMuggins:
I will rebase once #276 is merged, the only new commit here is [this one](https://github.com/canonical/microcloud/pull/277/commits/fba259c7c20688198732ea7783e41dc2486c7507).

### DIFF
--- a/microcloud/cmd/microcloud/main_init.go
+++ b/microcloud/cmd/microcloud/main_init.go
@@ -13,6 +13,7 @@ import (
 	"github.com/canonical/lxd/shared"
 	lxdAPI "github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/lxd/shared/validate"
 	cephTypes "github.com/canonical/microceph/microceph/api/types"
 	cephClient "github.com/canonical/microceph/microceph/client"
 	"github.com/canonical/microcluster/client"
@@ -137,6 +138,11 @@ func (c *cmdInit) RunInteractive(cmd *cobra.Command, args []string) error {
 	}
 
 	err = c.common.askNetwork(s, systems, c.flagAutoSetup)
+	if err != nil {
+		return err
+	}
+
+	err = validateSystems(s, systems)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


microcloud/cmd/microcloud/main_init.go
```diff
@@ -414,6 +419,82 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
 	return nil
 }
 
+func ensureSystemsOutsideGateway(gateway string, localAddr string, systems map[string]InitSystem) error {
```
MggMuggins, line 422: This is a bit inefficient as `ensureSystemsOutsideGateway` parses all the system addresses twice; I'm happy to rework this if needed but I thought I'd go for readability for this first draft.

MggMuggins:
@masnax Would you be willing to take a quick pass at this? I'm looking for feedback on if the tests look reasonable as much as anything. Realize I missed testing the case where other `InitSystem` addresses conflict with the uplink gateway. I'll add that and fix the linter this afternoon.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -414,6 +419,82 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
...
+	return nil
+}
+
+func validateSystems(s *service.Handler, systems map[string]InitSystem) error {
+	_, bootstrap := systems[s.Name]
+	if !bootstrap {
+		return nil
+	}
+
+	for _, curSystem := range systems {
```
masnax, line 457: Looks like you call `ensureSystemsOutsideGateway` twice for each qualifying network, and that is repeated for each system. Actually, the OVN networks should be identical on all systems, so we can factor away the outer loop, by only considering the `Networks` field on one `InitSystem`.

How about iterating over the systems first, collecting a list of parsed IPs for each system address, and then only iterating over `systems[s.Name].Networks`. Then we can change `ensureSystemsOutsideGateway` to have a signature like
`func ensureSystemsOutsideGateway(gateway string, systemIPs []net.IP) error`. 

That way we won't be redundantly parsing the system addresses and we won't validate the same uplink network `len(systems)` times. What do you think?

MggMuggins: I've restructured the code around the assumption that all the UPLINK nets are the same; I think the result is more readable and the tests make better sense. Let me know what you think.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -414,6 +419,68 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
...
+	// systems' management net addrs
+	hasIP4Gateway, hasIP6Gateway := false, false
+	var ip4Gateway, ip6Gateway string
+	var ip4GatewayNet, ip6GatewayNet *net.IPNet
+
+	for _, network := range curSystem.Networks {
+		if network.Type == "physical" && network.Name == "UPLINK" {
+			ip4Gateway, hasIP4Gateway = network.Config["ipv4.gateway"]
+			ip6Gateway, hasIP6Gateway = network.Config["ipv6.gateway"]
+			ovnRanges, hasOVNRanges := network.Config["ipv4.ovn.ranges"]
```
tomponline, line 439: https://documentation.ubuntu.com/lxd/en/latest/reference/network_physical/#network-physical-network-conf:ipv6.ovn.ranges is a thing in LXD, does microcloud offer to support it, if so we need to also check it here.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -414,6 +419,68 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
...
+		}
+	}
+
+	for systemName, system := range systems {
+		systemAddr := net.ParseIP(system.ServerInfo.Address)
+		if systemAddr == nil {
+			return fmt.Errorf("Invalid address %q for system %q", system.ServerInfo.Address, systemName)
+		}
+
+		if hasIP4Gateway && ip4GatewayNet.Contains(systemAddr) {
```
tomponline, line 472: A comment explaining why this is the case would be great.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -414,6 +419,68 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
...
+	var ip4GatewayNet, ip6GatewayNet *net.IPNet
+
+	for _, network := range curSystem.Networks {
+		if network.Type == "physical" && network.Name == "UPLINK" {
+			ip4Gateway, hasIP4Gateway = network.Config["ipv4.gateway"]
+			ip6Gateway, hasIP6Gateway = network.Config["ipv6.gateway"]
+			ovnRanges, hasOVNRanges := network.Config["ipv4.ovn.ranges"]
+
+			if hasIP4Gateway {
+				_, ip4GatewayNet, err = net.ParseCIDR(ip4Gateway)
```
tomponline, line 442: The gateway fields are expecting an IP address and subnet in CIDR format.
This check as it stands would allow an invalid entry where the address is the network address of the subnet, e.g.

`192.168.1.0/24` should be invalid, but `192.168.1.1/24` is valid.

See `IsNetworkAddressCIDRV4()` in LXD.

microcloud/cmd/microcloud/main_init.go
tomponline: I'd like to understand more about what this is check is for? Are we truly aiming to check that the system address isn't within the entire uplink subnet, or are we wanting to check something else?
Thanks

microcloud/cmd/microcloud/main_init.go
masnax: No, so far MicroCloud only offers a range for ipv4. Perhaps this is something we should add (but out of scope for this pr)

tomponline: should we add it here just in case it gets added in the future and missed ?

microcloud/cmd/microcloud/main_init.go
masnax: That's right, we want to check that the system addresses used for internal microcluster traffic don't collide with the uplink gateway subnet. 

The whole uplink subnet may not directly be in use at once (thanks to `ipv4.ovn.ranges`) but the user still did specify that they're reserving that subnet for the uplink network so it shouldn't collide with any of the addresses used for dqlite. 

Additionally I'm not sure how much of the uplink subnet is used for OVN if it's IPv6 as we don't specify a range. So I think in general having a unique subnet per required network interface is the way to go here. 

The corresponding issue mentions a "management subnet" but such a thing doesn't exist in MicroCloud, the subnet chosen for mDNS lookup is more of a filter, and won't necessarily be the smallest subnet that matches the management network. In fact the `/8` subnet used in the example would be impossible to saturate with MicroCloud.

tomponline: > The whole uplink subnet may not directly be in use at once (thanks to `ipv4.ovn.ranges`) but the user still did specify that they're reserving that subnet for the uplink network so it shouldn't collide with any of the addresses used for dqlite.

I'm not sure about this. It might prevent some valid existing setups. Let me try it out and get back to you.

masnax: Interesting, I was under the impression that the majority of MicroClouds are set up just onnce. Are there many MicroCloud setups that are regularly initialized?

tomponline: You misunderstood. I mean it may prevent setting up previously valid setups.

tomponline: Thinking about this I'm pretty sure having both the microcloud admin network interface and the ovn uplink network interface being connected to the same network (and thus same subnets) is fine and shouldn't be prevented (infact I should think its a likely common setup).

Any ideas where the idea to prevent it came from?

masnax: It's the second case mentioned in the corresponding isssue #210 

There's a notion of a "management network" that uses the subnet ` 10.42.0.49/8`, and then the uplink gateway collides with it at `10.254.254.1/24`. 

As I mentioned above, there's no such thing as a "management network" in MicroCloud based on a subnet input, rather the lookup subnet just filters the mDNS lookup. But I had thought the extra validation for actually selected systems would be useful. If you think it's too restrictive then we don't need to check each system address against the uplink subnet.

masnax (changes requested):
@MggMuggins There's one more component to the networking questions: the dns nameservers we pass to LXD.

Can you also include a check that verifies that the input is a comma-separated list of IPs?

It's part of the same UPLINK network struct as the other fields, accessed by the config key `dns.nameservers`

https://github.com/canonical/microcloud/blob/a28d437351bb9dc3ea38aa9791441404b77a89bd/microcloud/service/lxd_config.go#L88-L90

microcloud/cmd/microcloud/main_init.go
tomponline: I replied here https://github.com/canonical/microcloud/issues/210

tomponline (approved):
Apart from the hardcoded UPLINK, which can be addressed in a separate PR, this looks great thanks.

microcloud/cmd/microcloud/main_init.go
```diff
@@ -414,6 +420,112 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
...
+		return nil
+	}
+
+	// Assume that the UPLINK network on each system is the same, so grab just
+	// the gateways from the current node's UPLINK to verify against the other
+	// systems' management addrs
+	var ip4OVNRanges, ip6OVNRanges []*shared.IPRange
+
+	for _, network := range curSystem.Networks {
+		if network.Type == "physical" && network.Name == "UPLINK" {
```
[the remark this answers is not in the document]
masnax: @MggMuggins you can use [this method](https://github.com/canonical/microcloud/blob/0c6fff36559abb2277e4dcb08d146992a31b8400/microcloud/service/lxd_config.go#L75) to grab the corresponding network type. These methods are how we keep the payloads that we send to LXD consistent. 

 You can either instantiate a new `LXDService` or extract the receiver from `handler.Services[types.LXD].(*service.LXDService)` from the `service.Handler` that carries through the whole init.

PR #281, opened 2024-04-10
github: add system tests

simondeziel:


microcloud/test/includes/microcloud.sh
```diff
@@ -475,7 +475,7 @@ reset_snaps() {
       snap enable microcloud > /dev/null 2>&1 || true
       snap start lxd > /dev/null 2>&1 || true
       snap start microcloud > /dev/null 2>&1 || true
-      snap refresh lxd --channel latest/stable --cohort=+
+      snap refresh lxd --cohort=+
```
tomponline, line 478: Can this be latest/edge to catch LXD regressions with MicroCloud usage ?
Similar to how we use latest edge of liblxc and dqlite in LXD's tests.

microcloud/test/includes/microcloud.sh
```diff
@@ -849,8 +849,8 @@ setup_lxd_project() {
     fi
 
     lxc remote switch local
-	  lxc project create microcloud-test || true
-	  lxc project switch microcloud-test
+	lxc project create microcloud-test || true
```
tomponline: why `|| true` here?

microcloud/test/includes/microcloud.sh
```diff
@@ -959,9 +959,9 @@ setup_system() {
 
     if [ -n "${MICROCLOUD_SNAP_PATH}" ]; then
       lxc file push "${MICROCLOUD_SNAP_PATH}" "${name}"/root/microcloud.snap
-      lxc exec "${name}" -- sh -c "PATH=\$PATH:/snap/bin snap install --devmode /root/microcloud.snap"
+      lxc exec "${name}" -- snap install --devmode /root/microcloud.snap
```
tomponline, line 959: where does /root/microcloud.snap come from? Should we be sideloading the PR's binary into the edge microcloud snap instead?

tomponline: We should be using `--dangerous` rather than `--devmode` as the latter isn't confined and isn't a representative test.

microcloud/test/includes/microcloud.sh
```diff
@@ -845,8 +845,8 @@ setup_lxd_project() {
     fi
 
     lxc remote switch local
-	lxc project create microcloud-test || true
-	lxc project switch microcloud-test
+    lxc project create microcloud-test || true
```
tomponline, line 848: why `|| true`?

microcloud/test/includes/microcloud.sh
```diff
@@ -966,14 +966,13 @@ setup_system() {
...
-  lxc stop "${name}"
-
-  lxc snapshot "${name}" snap0
-
-  lxc start "${name}"
-
-  lxd_wait_vm "${name}"
+  # Create a snapshot so we can restore to this point.
+  if [ "${SNAPSHOT_RESTORE}" = 1 ]; then
+    lxc stop "${name}"
```
tomponline, line 975: lxc stop --force?

microcloud/test/includes/microcloud.sh
```diff
@@ -934,12 +934,12 @@ setup_system() {
     # Snaps can occasionally fail to install properly, so repeatedly try.
     lxc exec "${name}" -- sh -c "
       while ! test -e /snap/bin/microceph ; do
-        snap install microceph || true
+        snap install microceph --cohort='+' || true
```
tomponline, line 937: should we use `--edge` or at least `<LTS track>/edge`? for this and microovn to catch regressions early??

microcloud/test/includes/microcloud.sh
```diff
@@ -968,6 +968,9 @@ setup_system() {
 
   # Create a snapshot so we can restore to this point.
   if [ "${SNAPSHOT_RESTORE}" = 1 ]; then
+    echo "::warning:: ==> XXX: working around LXD shutdown bug"
```
tomponline, line 972: Please an you elaborate on what this is and reference a GH issue or PR tracking it?

tomponline (approved):
Thanks for this, there's plenty of cleanup in there.

I would like to understand why there are different lxd channels being used apparently in the tests.

But any tests are better than none!

roosterfish (commented):
Nice, great to have those tests running now!

microcloud/test/suites/basic.sh
```diff
@@ -110,7 +110,7 @@ test_instances_config() {
 
   # Setup a MicroCloud with 3 systems, ZFS storage, and a FAN network.
   addr=$(lxc ls micro01 -f csv -c4 | grep enp5s0 | cut -d' ' -f1)
-  lxc exec micro01 -- TEST_CONSOLE=0 microcloud init --preseed << EOF
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud init --preseed << EOF
```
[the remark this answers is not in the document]
simondeziel: In this specific case, the command replaced does seem to be valid as LXD tries to `execve` `TEST_CONSOLE=0` which is not a valid binary in the instance. Here's a simple reproducer:
```
$ lxc exec c1 -- FOO=1 env
Error: Command not found

$ lxc exec c1 --env FOO=1 -- env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
container=lxc
FOO=1
TERM=xterm-256color
HOME=/root
USER=root
LANG=C.UTF-8
```

There are many more examples where I replaced `lxc exec foo -- sh -c " ... "` to drop the wrapper shell mostly because I find it opens the door to doing a lot inside that subshell and it's easy to lose visibility (forgetting to use `sh -xc` for example). Mostly a question of style/taste I guess.

I've kept the `sh -c` where it was needed for IO redirections when the receiving file (usually `> out`) was meant to live inside the instance running the command.

microcloud/test/includes/microcloud.sh
```diff
@@ -1054,20 +1054,20 @@ lxd_wait_vm() {
...
   echo "==> ${name} Awaiting VM..."
-  for round in $(seq 640); do
+  for round in $(seq 160); do
     if lxc info "${name}" | grep -qF "Status: READY" ; then
       wait_snapd "${name}"
       echo "    ${name} VM is ready"
       return 0
     fi
 
     # Sometimes the VM just won't start, so retry after 3 minutes.
```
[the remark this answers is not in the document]
simondeziel: IIRC, I also changed the `sleep 1` to a `sleep 4` but maybe my maths are still off.

microcloud/test/suites/basic.sh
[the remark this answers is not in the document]
tomponline: In general when we see `sh -c` being used its usually a mistake.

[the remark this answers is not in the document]
roosterfish: I see, thanks for the detailed explanation!

microcloud/test/includes/microcloud.sh
masnax: If the project doesn't exist, it will create the project.

If the project does exist, it will won't break at trying to create the project and will continue with the test. This is very helpful when you're running tests locally and something breaks so the tests don't clean up properly.

microcloud/test/includes/microcloud.sh
masnax: We're snapshotting the VM here, so if we `lxc stop -f`, that makes the resulting snapshot less deterministic as we just force quit everything.

microcloud/test/includes/microcloud.sh
tomponline: hrm yeah but this could be applied to any state setup not just projects.
Seemed kind of odd to see this here as it suggests we dont know what state the host will be in when the tests are running.

microcloud/test/includes/microcloud.sh
masnax: We should probably have both. We should run tests on `stable` because that's what we know today will definitely work, and we should have some cases for `edge` to ensure we anticipate any upcoming breakages.

microcloud/test/includes/microcloud.sh
tomponline: i think in the lxd vm tests we execute a `sync` call and then force stop to ensure it runs more reliably

https://github.com/canonical/lxd-ci/blob/main/tests/storage-vm#L83-L86

microcloud/test/includes/microcloud.sh
tomponline: yeah agreed, with a preference towards edge builds initially.

microcloud/test/includes/microcloud.sh
```diff
@@ -948,7 +948,8 @@ setup_system() {
       fi
 
       while ! test -e /snap/bin/lxd ; do
-        snap install lxd --cohort='+' || true
+        echo \"::warning::pulling LXD from latest/edge (https://github.com/canonical/lxd-pkg-snap/pull/426)\"
+        snap install lxd --cohort='+' --channel=latest/edge || true
```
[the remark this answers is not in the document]
masnax: I believe this was in place to catch the MicroOVN content interface updates before they were pushed to stable, hence the comment linking the PR.

microcloud/test/includes/microcloud.sh
simondeziel: We run that function/take that snapshot only once so I think it's safer to play it safe and that the speed hit which should be small as those system were freshly created to there is little risk for them to hang during stop.

microcloud/test/includes/microcloud.sh
masnax: The only danger I can see is that if a user had a pre-existing `microcloud-test` project, then their default profile in that project is going to get wiped by running the test suite. That and maybe pre-existing project restrictions but that would cause the test suite to fail the initial setup anyway. 

As for the state of the hosts, it's inconsequential to the tests which other networks, profiles, storage pools, or instances exist on that project and we're strict about the creation of any of those that are necessary for the test suite. In the case of CI, since the whole LXD is replaced per run, then it doesn't matter for that either.

I think it's useful to have this for making it easy to debug the test suite but I'm fine adding it manually if I need it locally.

microcloud/test/includes/microcloud.sh
simondeziel: I added a link to the PR in question: https://github.com/canonical/microcloud/pull/286

simondeziel:
All the feedback is being addressed in #290 (still in draft).

PR #284, opened 2024-04-15
Fix DNS nameserver env var in tests

masnax:
@simondeziel this should fix your issue

### DIFF
--- a/microcloud/test/suites/add.sh
+++ b/microcloud/test/suites/add.sh
@@ -123,8 +123,8 @@ test_add_interactive() {
   export IPV4_SUBNET="10.1.123.1/24"
   export IPV4_START="10.1.123.100"
   export IPV4_END="10.1.123.254"
-  export CUSTOM_DNS_ADDRESSES="10.1.123.1,8.8.8.8" # comma-separated list of custom DNS addresses to be set for the OVN uplink.
   export IPV6_SUBNET="fd42:1:1234:1234::1/64"
+  export DNS_ADDRESSES="10.1.123.1,fd42:1:1234:1234::1"
   microcloud_interactive | lxc exec micro01 -- sh -c "microcloud init > out"
   lxc exec micro01 -- tail -1 out | grep "MicroCloud is ready" -q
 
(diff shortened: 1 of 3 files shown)


simondeziel (approved):
Thanks!

microcloud/test/suites/add.sh
```diff
@@ -123,8 +123,8 @@ test_add_interactive() {
   export IPV4_SUBNET="10.1.123.1/24"
   export IPV4_START="10.1.123.100"
   export IPV4_END="10.1.123.254"
-  export CUSTOM_DNS_ADDRESSES="10.1.123.1,8.8.8.8" # comma-separated list of custom DNS addresses to be set for the OVN uplink.
   export IPV6_SUBNET="fd42:1:1234:1234::1/64"
+  export DNS_ADDRESSES="10.1.123.1,fd42:1:1234:1234::1/64"
```
roosterfish, line 127: Should the IPv6 DNS address have a subnet mask? 
If so, why isn't it required for the IPv4 address?
From what I understood the validation will check if the addresses set in `DNS_ADDRESSES` are actually there.

I have never worked with this test suite so please bare with me :)

roosterfish (commented):
Looks good to me there is just one variable I am not sure about.

microcloud/test/suites/add.sh
tomponline: No it shouldnt have a subnet mask

tomponline: If the tests are passing with this value then the system isnt performing sufficient validation.

masnax: That's what #277 is for.

masnax: Though if the tests actually did pass that means LXD also isn't validating this config key when it receives it.

tomponline: Possibly, both are not doing it correctly, can you check and update where needed please.

simondeziel: The microcloud tests are failing on this unexpected CIDR notation, see https://github.com/canonical/microcloud/actions/runs/8711674992/job/23896307777?pr=281#step:7:2774:

```
---
10.1.123.1/24                                         # setup ipv4/ipv6 gateways and ranges
10.1.123.100
10.1.123.254
fd42:1:1234:1234::1/64
10.1.123.1,fd42:1:1234:1234::1/64
+ sed -e /^\s*#/d -e /^\s*$/d
Invalid input: Item "fd42:1:1234:1234::1/64": Not an IP address "fd42:1:1234:1234::1/64"
```

simondeziel: Dropping that bogus `/64` makes the tests go much further in #281

simondeziel (approved):

tomponline (approved):

PR #285, opened 2024-04-16
Tutorial updates

ru-fu:


roosterfish (approved):
LGTM! 

I was wondering if it would be helpful to also add the `lxc shell micro01` and `lxc config trust` example commands in the info box after the `lxc file push` to make it clear where it has to be executed but I guess this would be duplicated information then in both the UI instructions and in the docs. Just a thought should not be a blocker.

ru-fu:
Good point - updated. :)

roosterfish (approved):

PR #286, opened 2024-04-17
Only restart LXD if version is less than 5.21

masnax:
With the content interfaces in places in LXD to detect MicroOVN and MicroCeph, we don't need to restart LXD to pick them up anymore. 

However, if we don't restart, this breaks support for versions of LXD prior to 5.21 when the content interface was introduced. We can't directly fetch the snap version or the available interfaces of LXD due to the snap confinement, so the best I came up with was to hard-code `5.21` into MicroCloud, and compare it to the server version reported from `/1.0` in LXD. 

If LXD is not at least version 5.21, then we need to restart the LXD daemon to pick up MicroOVN and MicroCeph integration. Otherwise we can just expect it to be there.

### DIFF
--- a/microcloud/service/lxd.go
+++ b/microcloud/service/lxd.go
@@ -11,12 +11,13 @@ import (
 	"strings"
 	"time"
 
-	lxd "github.com/canonical/lxd/client"
+	"github.com/canonical/lxd/client"
 	"github.com/canonical/lxd/lxd/util"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/microcluster/microcluster"
+	"golang.org/x/mod/semver"
 
 	"github.com/canonical/microcloud/microcloud/api/types"
 	"github.com/canonical/microcloud/microcloud/mdns"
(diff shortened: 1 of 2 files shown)


microcloud/service/lxd.go
```diff
@@ -420,6 +421,17 @@ func (s *LXDService) Restart(ctx context.Context, timeoutSeconds int) error {
 		return fmt.Errorf("Detected pre-existing LXD storage pools. LXD might have already been initialized")
 	}
 
+	server, _, err := c.GetServer()
+	if err != nil {
+		return fmt.Errorf("Failed to get LXD server information: %w", err)
+	}
+
+	if semver.Compare(semver.Canonical(fmt.Sprintf("v%s", server.Environment.ServerVersion)), semver.Canonical(fmt.Sprintf("v%s", lxdMinVersion))) < 0 {
```
tomponline, line 429: I think future readers would really appreciate a comment here explaining *why* we are doing this.

tomponline (changes requested):
Ready for merge once the comments are improved.

microcloud/service/lxd.go
masnax: Fixed.

tomponline (approved):

simondeziel:
The test failures are addressed by https://github.com/canonical/microcloud/pull/289

tomponline:
@masnax please can you rebase to get the tests passing. Ta

PR #287, opened 2024-04-18
Add helper for setting up MicroCloud proxy

masnax:
Should be merged after #259 

In response to reviews from #259, I've added some additional commits to update the microcluster dependency, fix contexts, and add 2 new helpers:
* `UseAuthProxy` which applies the auth proxy used to manage the API of other services to a microcluster `Client`. This is now used in place of manually applying the proxy all over the place.
* `AuthProxy` which returns the actual proxy implementation, so we can also reduce the duplication for LXD's client which is not the same structure as `microcluster`'s.

### DIFF
--- a/service/lxd.go
+++ b/service/lxd.go
@@ -5,8 +5,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"net/http"
-	"net/url"
 	"os"
 	"strings"
 	"time"
@@ -20,6 +18,7 @@ import (
 	"golang.org/x/mod/semver"
 
 	"github.com/canonical/microcloud/microcloud/api/types"
+	cloudClient "github.com/canonical/microcloud/microcloud/client"
 	"github.com/canonical/microcloud/microcloud/mdns"
 )
 
@@ -60,14 +59,7 @@ func (s LXDService) Client(ctx context.Context, secret string) (lxd.InstanceServ
 	return lxd.ConnectLXDUnixWithContext(ctx, s.m.FileSystem.ControlSocket().URL.Host, &lxd.ConnectionArgs{
 		HTTPClient:    c.Client.Client,
 		SkipGetServer: true,
-		Proxy: func(r *http.Request) (*url.URL, error) {
-			r.Header.Set("X-MicroCloud-Auth", secret)
-			if !strings.HasPrefix(r.URL.Path, "/1.0/services/lxd") {
-				r.URL.Path = "/1.0/services/lxd" + r.URL.Path
-			}
-
-			return shared.ProxyFromEnvironment(r)
-		},
+		Proxy:         cloudClient.AuthProxy(secret, types.LXD),
 	})
 }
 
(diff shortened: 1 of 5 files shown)


masnax:
@roosterfish @markylaing This one is now ready for review, thanks.

masnax:
Ah crap, the microcluster changes clash with MicroCeph. I guess this is blocked on MicroCeph updating to the latest microcluster.

service/microceph.go
```diff
@@ -16,6 +15,7 @@ import (
 	"github.com/canonical/microcluster/microcluster"
 
 	"github.com/canonical/microcloud/microcloud/api/types"
+	cloudCli "github.com/canonical/microcloud/microcloud/client"
```
markylaing, line 18: Can we call this `microcloudClient` or `cloudClient` to be consistent with the `cephClient` above?

cmd/microcloud/waitready.go
```diff
@@ -31,10 +32,17 @@ func (c *cmdWaitready) Run(cmd *cobra.Command, args []string) error {
...
 
-	return m.Ready(c.flagTimeout)
+	ctx := context.Background()
+	if c.flagTimeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, time.Second*time.Duration(c.flagTimeout))
+		defer cancel()
+	}
+
+	return m.Ready(ctx)
```
markylaing, line 47: Ok I've just seen these commits related to passing contexts around are related to a change to microcluster. Can we move these to a separate PR for updating dependencies? Then this PR can focus just on adding the proxy helper.

markylaing (changes requested):
Looks good, I think it'd be worth splitting this into two PRs though. One for the dependency update and another for the auth proxy helpers.

cmd/microcloud/waitready.go
masnax: Sure, although in the interest of not growing the number of PRs that are unable to be merged, I'll hold off on that until it's safe to update microcloud's dependency :)

roosterfish (commented):
All the comments I had on those changes were already addressed over in https://github.com/canonical/microcloud/pull/260 as those commits are included over there too. 

The changes look good to me :) I'll approve as soon as https://github.com/canonical/microcloud/pull/320 is merged so this PR contains only the proxy related changes.

service/microceph.go
roosterfish: This one is still open, rest looks good.

service/lxd.go
```diff
@@ -20,6 +18,7 @@ import (
 	"golang.org/x/mod/semver"
 
 	"github.com/canonical/microcloud/microcloud/api/types"
+	cloudCli "github.com/canonical/microcloud/microcloud/client"
```
roosterfish, line 21: Same comment as https://github.com/canonical/microcloud/pull/287/commits/0191050b30712b99814e7e1032c536d27f1b8f05#r1621963740.

masnax: I've made it `cloudClient`, dropping `micro` to keep it consistent with the other usages of micro* project imports.

roosterfish (approved):
Thanks, looks good!

PR #289, opened 2024-04-19
Don't use mDNS payload address for local node validation

masnax:
The local node does not have an mDNS record (the struct is blank) so to validate its cluster address we need to grab it from the `Handler`.

### DIFF
--- a/microcloud/cmd/microcloud/main_init.go
+++ b/microcloud/cmd/microcloud/main_init.go
@@ -505,9 +505,15 @@ func validateSystems(s *service.Handler, systems map[string]InitSystem) (err err
 	// Ensure that no system's management address falls within the OVN ranges
 	// to prevent OVN from allocating an IP that's already in use.
 	for systemName, system := range systems {
-		systemAddr := net.ParseIP(system.ServerInfo.Address)
+		// If the system is ourselves, we don't have an mDNS payload so grab the address locally.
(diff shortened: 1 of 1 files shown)


masnax:
```
Error: Failed to issue LXD token for peer "micro03": There are no online cluster members
```

That's a fun one, looks like the first LXD didn't bootstrap fast enough before we attempted to generate tokens.

masnax:
> I would think bootstrapping LXD would only return after the node is actually operational.

Looking at the corresponding code block:  https://github.com/canonical/lxd/blob/9bc2974a84d15fb9c16de04666dc79041a8a19ca/lxd/api_cluster.go#L1323-L1339


It looks like we're considering nodes to be offline if they haven't received a heartbeat, but shouldn't the local node always be online to itself?

simondeziel:
> It looks like we're considering nodes to be offline if they haven't received a heartbeat, but shouldn't the local node always be online to itself?

I guess it depends on what you define as being online. Having received a heartbeat is IMHO a very good indication of being online but maybe it could be more flexible for itself? I don't know the code, but that definition of online seems to be a consistent way to treat everyone the same with regards to connectivity.

MggMuggins (approved):
I won't speak to the CI failure. Thanks for correcting my assumption.

tomponline (approved):

PR #290, opened 2024-04-19
System tests misc improvements

simondeziel:


simondeziel:
@masnax I know we agreed one splitting the interactive one even more but since I got a full green run, is that OK if we merge it as-is?

masnax (approved):
Yay! All green :)

I'll have another PR coming up in a sec that will reorder the runners and log `lxc info --resources` plus a few other workarounds for issues I faced locally.

PR #293, opened 2024-04-24
Fix test suite hangs

masnax:
For the debug binaries and ZFS, I'm not sure exactly why the timeouts are necessary but without them, I pretty consistently run into issues where the action is actually performed, but the command won't let go.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -92,7 +92,7 @@ jobs:
         go: ["1.22.x"]
         suite:
           - "add"
-          - "auto"
+          - "instances"
           - "basic"
           - "interactive"
           - "mismatch"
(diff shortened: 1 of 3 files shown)


masnax:
@simondeziel did you ever run into issues with the places I put timeouts in?

simondeziel:
> @simondeziel did you ever run into issues with the places I put timeouts in?

No

masnax:
> > @simondeziel did you ever run into issues with the places I put timeouts in?
> 
> No


You wouldn't have, since I removed the commits :)

The issue seems to be related to `RUN_CONCURRENT=1`

Seems when we detach the process and then run `lxc exec -- {something}`, if the command returns too quickly then it never exits. Not sure if it's just a bug with my shell.

simondeziel:
> Seems when we detach the process and then run `lxc exec -- {something}`, if the command returns too quickly then it never exits. Not sure if it's just a bug with my terminal.

@MusicDin does that ring a bell from when you investigated/worked on a `lxc exec` bug a while ago?

PR #294, opened 2024-04-25
doc/tutorial/get_started: LXD is not always refresh-able

simondeziel:
With 24.04 onward or if using any `minimal` image, the LXD snap won't be there. Also make it clear that we don't get the latest version but the latest LTS.

With that small tweak, I got a microcloud cluster running on 4x 24.04 VMs. While it's a bit early to change the doc to use 24.04 VMs, it's a first step.

### DIFF
--- a/doc/tutorial/get_started.rst
+++ b/doc/tutorial/get_started.rst
@@ -225,10 +225,13 @@ Complete the following steps on each VM (``micro1``, ``micro2``, ``micro3``, and
       The ``--cohort="+"`` flag in the command ensures that the same version of the snap is installed on all machines.
       See :ref:`howto-snap-cluster` for more information.
 
-#. The LXD snap is already installed.
-   Refresh it to the latest version::
+#. If the LXD snap is already installed, refresh it to the latest LTS version::
 
      snap refresh lxd --channel=5.21/stable --cohort="+"
+
+   Otherwise, install it::
+
+     snap install lxd --channel=5.21/stable --cohort="+"
 
 6. Initialise MicroCloud
 ------------------------


masnax (approved):

doc/tutorial/get_started.rst
```diff
@@ -225,11 +225,15 @@ Complete the following steps on each VM (``micro1``, ``micro2``, ``micro3``, and
       The ``--cohort="+"`` flag in the command ensures that the same version of the snap is installed on all machines.
       See :ref:`howto-snap-cluster` for more information.
 
-#. The LXD snap is already installed.
-   Refresh it to the latest version::
+#. If the LXD snap is already installed.
+   Refresh it to the latest LTS version::
```
roosterfish, line 229: I am by far not good at this, but wouldn't the sentence be more fluid using a comma between `installed, refresh`?

roosterfish (changes requested):

doc/tutorial/get_started.rst
tomponline: Whats with the double colons too?

simondeziel: I just copied the double colons scheme which seems to be rendered as a single one: https://canonical-microcloud--294.com.readthedocs.build/en/294/tutorial/get_started/#install-microcloud-on-each-vm

Before checking the preview, I assumed the 2 lines were separated but that's not the case so I agree that `installed, refresh` reads better.

tomponline: How strange, is this normal @ru-fu ?

ru-fu: Yes, that's RST syntax for introducing a code block. See https://canonical-documentation-with-sphinx-and-readthedocscom.readthedocs-hosted.com/style-guide/#code-blocks

ru-fu:
> With 24.04 onward or if using any `minimal` image, the LXD snap won't be there.

But the tutorial doesn't use 24.04. A tutorial should be one clear path and not offer options - adding the "if it is installed" part makes it sound like we don't know whether it is, but we're working with VMs and know what is installed.

doc/tutorial/get_started.rst
simondeziel: The double colons are apparently used before code blocks.

simondeziel:
> > With 24.04 onward or if using any `minimal` image, the LXD snap won't be there.
> 
> But the tutorial doesn't use 24.04. A tutorial should be one clear path and not offer options - adding the "if it is installed" part makes it sound like we don't know whether it is, but we're working with VMs and know what is installed.

I wanted to make it easier for those diverging a little from the tutorial, but yeah, I see what you mean. Thanks

PR #299, opened 2024-04-30
Fix hardcoded uplink

MggMuggins:
Per the discussion in #277

It feels a little silly to do a bunch of relatively more complex stuff at runtime to fix something that can be fixed by a single source of truth at compile time.

### DIFF
--- a/cmd/microcloud/main_init.go
+++ b/cmd/microcloud/main_init.go
@@ -524,7 +524,7 @@ func validateGatewayNet(config map[string]string, ipPrefix string, cidrValidator
 
 	for _, ipRange := range ovnIPRanges {
 		if ipRange.ContainsIP(gatewayIP) {
-			return nil, fmt.Errorf("UPLINK %s.ovn.ranges must not include gateway address %q", ipPrefix, gatewayNet.IP)
+			return nil, fmt.Errorf("%s %s.ovn.ranges must not include gateway address %q", service.DefaultUplinkNetwork, ipPrefix, gatewayNet.IP)
 		}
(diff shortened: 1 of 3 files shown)


microcloud/service/lxd_config.go
```diff
@@ -93,7 +96,7 @@ func (s LXDService) DefaultOVNNetwork(ipv4Gateway string, ipv4Range string, ipv6
 	}
 
 	ovnNetwork := api.NetworksPost{
-		NetworkPut: api.NetworkPut{Config: map[string]string{"network": "UPLINK"}, Description: "Default OVN network"},
+		NetworkPut: api.NetworkPut{Config: map[string]string{"network": DefaultOVNNetwork}, Description: "Default OVN network"},
 		Name:       "default",
```
masnax, line 100: Given that we also have `"default"` as an `ovn` type network, I think the const name might be a bit misleading.

How about using `DefaultOVNNetwork` for `"default"` and `OVNUplinkNetwork` for `"UPLINK"`?

masnax (changes requested):
Just one nit :)

microcloud/service/lxd_config.go
tomponline: I strongly agree. The uplink isn't an ovn network

masnax:
@MggMuggins Heads up that this needs a rebase now.

MggMuggins:
Woops; forgot about this one. Done. Test failure feels very unlikely to be related.

masnax (approved):

PR #300, opened 2024-05-02
Fix SNAPSHOT_RESTORE=0 tests

masnax:


microcloud/test/includes/microcloud.sh
```diff
@@ -466,15 +471,11 @@ reset_snaps() {
...
-      snap start lxd > /dev/null 2>&1 || true
-      snap start microcloud > /dev/null 2>&1 || true
-      snap refresh lxd --cohort=+
-
-      lxd waitready
-    "
+    lxc exec "${name}" -- snap enable lxd > /dev/null 2>&1 || true
+    lxc exec "${name}" -- snap enable microcloud > /dev/null 2>&1 || true
+    lxc exec "${name}" -- snap start lxd > /dev/null 2>&1 || true
+    lxc exec "${name}" -- snap start microcloud > /dev/null 2>&1 || true
```
simondeziel, line 477: I just found that `snap start --enable` is a thing. Would that work here?

simondeziel (dismissed):
LGTM, thanks

microcloud/test/includes/microcloud.sh
masnax: I think they do different things.

`snap enable` re-enables a previously disabled snap, which treats it as though it's not installed.

`snap start --enable` just makes the snap start automatically next time the system is booted.

microcloud/test/includes/microcloud.sh
```diff
@@ -933,7 +935,7 @@ setup_system() {
 
     # Install the snaps.
     lxc exec "${name}" -- apt-get update
-    if [ -n "${CLOUD_INSPECT:-}" ]; then
+    if [ -n "${CLOUD_INSPECT:-}" ] || [ "${SNAPSHOT_RESTORE}" = 0 ]; then
```
simondeziel, line 938: If I followed properly, this isn't needed anymore now that #301 is closed there is no need to have the test automatically issue `zpool` commands.

masnax: #301 just introduced a sleep to the `zpool destroy` call here: https://github.com/canonical/microcloud/blob/4f47db906f5846928fc56e8082e55c9a7d78f117/microcloud/test/includes/microcloud.sh#L524

In the non-snapshot case, we still need to make the call because sometimes LXD won't clean up the zpool properly, and you'll start another test with the disk in use.

simondeziel: Right. I take it that `blkdiscard --force` or `wipefs -a -f` is not enough?

masnax: Not if `zfs` won't let go of the disk. You'll get a `Device or resource busy`

masnax: Frustratingly, `wipefs -a` will fail with the error but `wipefs -a -f` will return successfully, but the disk will still have partitions and be in use.

masnax: Hmm, I'd like to try that. Might help with another issue I've seen with MicroCeph. 

I added `microceph.ceph osd set noup` because occasionally the OSDs would bring themselves back up mid-teardown, but if we just went ahead with the teardown of microceph, then we would nuke ceph, but the disk would remain in use.

I couldn't find anything obvious in `lsof`, `ps` or `/proc/self/mountinfo` that mentioned the disk was in use. 

Manually deleting disk records with `echo 1 > /sys/block/{disk_id}/device/delete` and rescanning with `echo '- - -' > /sys/class/scsi_host0/scan` would instead pick up 2 disks for the same disk. So if you had `sdb` grabbed by `ceph`, you after rescanning you would have `sdb` with partitions and `sdc` without any.

tomponline: > In the non-snapshot case, we still need to make the call because sometimes LXD won't clean up the zpool properly, and you'll start another test with the disk in use.

@masnax is this a bug in LXD? Logged an issue for it?

masnax: I think it's a bug with zfs. LXD appears to attempt to destroy the zpool, but it very rarely doesn't work. 

I can log an issue but I don't really have much of a reproducer.

tomponline: what error do you get when doing `lxc storage delete`?

masnax: None. The pool is gone from LXD, and there's no error reported but `zpool list` still reports a local zpool.

masnax: I should mention, `zpool destroy mypool` will return successfully but not destroy the pool in these cases, it needs `-f`. We can add a contingency to LXD thar tries to delete the pool with `-f` if the regular delete doesn't work, but I'm not 100% clear on the implications.

tomponline: `-f` unmounts data sets, so suggests its using a lazy unmount approach if not passed. LXD should probably pass `-f` then to ensure and fail if there are still mounted data sets (this should help identity any areas were mounts are being left over).

masnax:
@roosterfish Could this please get a final pass? This is basically just some workarounds to get the microcloud tests to run faster on local systems without needing to wait for the LXD agent in-between snapshot restores.

test/includes/microcloud.sh
```diff
@@ -533,6 +534,8 @@ reset_system() {
 
     reset_snaps "${name}"
 
+    # Attempt to destroy the zpool as we may have left dangling volumes when we wiped the LXD database earlier.
+    # This is slightly faster than deleting the volumes by hand on each system.
     lxc exec "${name}" -- zpool destroy -f local || true
```
masnax, line 539: So here, with the non-snapshot tests which are primarily optimized for speed, we wipe LXD's database (and entire state directory) out from under it and use the snap to restore it (`snap enable lxd` will restore `/var/snap/lxd/common/` if it doesn't exist). 

This means we won't properly clean up the zpool on each system (there may be dangling volumes) and so we need to manually `zpool destroy -f` to clear everything. We also do this with the `ceph` storage  driver, manually clearing everything out from under `ceph`.

test/includes/microcloud.sh
```diff
@@ -487,7 +487,6 @@ reset_snaps() {
     lxc exec "${name}" -- snap enable microcloud > /dev/null 2>&1 || true
     lxc exec "${name}" -- snap start lxd > /dev/null 2>&1 || true
     lxc exec "${name}" -- snap start microcloud > /dev/null 2>&1 || true
-    lxc exec "${name}" -- snap refresh lxd --cohort=+ || true
```
roosterfish: This line was added in the same PR in https://github.com/canonical/microcloud/pull/300/commits/cf4746f0aca0fea510425e7a169f5d1e2a0c609b. Can you please squash those commits?

test/includes/microcloud.sh
roosterfish: > We also do this with the ceph storage driver, manually clearing everything out from under ceph.

Are you referring to how it is done in LXD or some other test suite that does Ceph related cleanups?

roosterfish (changes requested):

test/includes/microcloud.sh
masnax: Huh, that's odd. Must've been removed and my rebase added it back :)

test/includes/microcloud.sh
masnax: I was referring to this block. https://github.com/canonical/microcloud/blob/c86c366230bac921eb6f9349295a6f88dfe31a0d/test/includes/microcloud.sh#L596

roosterfish: I see, thanks!

roosterfish (approved):
Looks good!

PR #301, opened 2024-05-02
Bandaid for `lxc exec` lockup

masnax:
If the test suite is run with SNAPSHOT_RESTORE=0 and CONCURRENT_SETUP=1 then `lxc exec` can occasionally get stuck and never return if the command being executed is fast enough.

This adds a workaround by just sleeping for 1s before such commands, but ideally we should address this in LXD.

### DIFF
--- a/microcloud/test/includes/microcloud.sh
+++ b/microcloud/test/includes/microcloud.sh
@@ -102,8 +102,18 @@ set_debug_binaries() {
 
   if [ -n "${MICROCLOUD_DEBUG_PATH}" ] && [ -n "${MICROCLOUDD_DEBUG_PATH}" ]; then
     echo "==> Add debug binaries for MicroCloud."
-    lxc exec "${name}" -- rm -f /var/snap/microcloud/common/microcloudd.debug
-    lxc exec "${name}" -- rm -f /var/snap/microcloud/common/microcloud.debug
+
+    # lxc exec may lock up on fast-running commands if running in the background, so we need to sleep 1s.
+    lxc exec "${name}" -- sh -c "
+      sleep 1
+      if test -e /var/snap/microcloud/common/microcloudd.debug ; then
+        rm -f /var/snap/microcloud/common/microcloudd.debug
+      fi
+
+      if test -e /var/snap/microcloud/common/microcloud.debug ; then
+        rm -f /var/snap/microcloud/common/microcloud.debug
+      fi
+    "
 
     lxc file push --quiet "${MICROCLOUDD_DEBUG_PATH}" "${name}"/var/snap/microcloud/common/microcloudd.debug
     lxc file push --quiet "${MICROCLOUD_DEBUG_PATH}" "${name}"/var/snap/microcloud/common/microcloud.debug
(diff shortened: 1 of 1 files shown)


masnax:
@MusicDin might this be related to the other lxc exec lockup issues?

tomponline:
> but ideally we should address this in LXD

Did you open a bug already for this?

masnax:
> > but ideally we should address this in LXD
> 
> Did you open a bug already for this?

I haven't yet, thanks for the reminder.

simondeziel:
The error message `The joining server version doesn't match (expected 5.21.1 with API count 386)` could be a bit more informative if it included the version/API it got that didn't match.

Speaking of cohort, this reminds me there is this unexpected (to me) [refresh of LXD](https://github.com/canonical/microcloud/blob/main/microcloud/test/includes/microcloud.sh#L474). I'm not seeing why a refresh would be needed there and also without a channel being specified.

masnax:
@tomponline The corresponding issue is here: https://github.com/canonical/lxd/issues/13425

@simondeziel Yeah that refresh looks redundant.

microcloud/test/includes/microcloud.sh
```diff
@@ -520,7 +538,8 @@ reset_system() {
 
     reset_snaps "${name}"
 
-    lxc exec "${name}" -- zpool destroy -f local || true
+    # lxc exec may lock up on fast-running commands if running in the background, so we need to sleep 1s.
+    lxc exec "${name}" -- sh -c "sleep 1 ; zpool destroy -f local" || true
```
simondeziel, line 542: Moving `sleep 1` out of the `lxc exec` would avoid needing this `sh -c`.

microcloud/test/includes/microcloud.sh
```diff
@@ -113,7 +123,15 @@ set_debug_binaries() {
...
     echo "==> Add a debug binary for LXD."
-    lxc exec "${name}" -- rm -f /var/snap/lxd/common/lxd.debug
+
+    # lxc exec may lock up on fast-running commands if running in the background, so we need to sleep 1s.
+    lxc exec "${name}" -- sh -c "
+      sleep 1
+      if test -e /var/snap/lxd/common/lxd.debug ; then
+        rm -f /var/snap/lxd/common/lxd.debug
+      fi
+    "
```
simondeziel, line 133: Why not move `sleep 1` out and then just `lxc exec "${name}" -- rm -f /var/snap/lxd/common/lxd.debug` which will be a noop if the file isn't there.

microcloud/test/includes/microcloud.sh
```diff
@@ -102,8 +102,18 @@ set_debug_binaries() {
...
+    lxc exec "${name}" -- sh -c "
+      sleep 1
+      if test -e /var/snap/microcloud/common/microcloudd.debug ; then
+        rm -f /var/snap/microcloud/common/microcloudd.debug
+      fi
+
+      if test -e /var/snap/microcloud/common/microcloud.debug ; then
+        rm -f /var/snap/microcloud/common/microcloud.debug
+      fi
+    "
```
simondeziel, line 116: I'd prefer `sleep 1` to be out of that bit and then simply do `lxc exec "${name}" -- rm -f /var/snap/microcloud/common/microcloudd.debug /var/snap/microcloud/common/microcloud.debug`

simondeziel:
> @simondeziel Yeah that refresh looks redundant.

Good, can you drop it in this bandaid PR? If not, I'm happy to do it in a separated one.

microcloud/test/includes/microcloud.sh
masnax: If I remove it out of the `lxc exec` then it would defeat the purpose. It's the `lxc exec` itself returning too quickly which causes the issue.

masnax:
> > @simondeziel Yeah that refresh looks redundant.
> 
> Good, can you drop it in this bandaid PR? If not, I'm happy to do it in a separated one.

I added it to #300

microcloud/test/includes/microcloud.sh
simondeziel: OK, that's embarrassing that I missed the whole point :/
Thanks :)

masnax:
Closing as I've narrowed this down to the specific containers I set up back in September. Not sure what's up with the container, but we don't need this PR anymore.

PR #303, opened 2024-05-03
microcloud/cmd/microcloud: Remove dqlite role shift timeouts

masnax:
After https://github.com/canonical/lxd/pull/12571 was merged, these timeouts are no longer necessary to avoid the join process breaking. 

The 4 system tests should stress-test this enough to capture any issues with removing the timeouts.

### DIFF
--- a/cmd/microcloud/main_init.go
+++ b/cmd/microcloud/main_init.go
@@ -470,9 +470,6 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
 		if err != nil {
 			return err
 		}
-
-		// Sleep 3 seconds to give the cluster roles time to reshuffle before adding more members.
-		time.Sleep(3 * time.Second)
 	}
 
 	for peer, cfg := range joinConfig {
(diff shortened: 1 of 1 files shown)


microcloud/cmd/microcloud/main_init.go
```diff
@@ -470,9 +470,6 @@ func AddPeers(sh *service.Handler, systems map[string]InitSystem) error {
 		if err != nil {
 			return err
 		}
-
```
roosterfish, line 473: Do we now require some retries with timeout? The interactive job gives `Get "http://unix.socket/1.0": dial unix /var/snap/lxd/common/lxd/unix.socket: connect: connection refused`.

Or is this unrelated to this change?

masnax:
Alas, the 4 system tests passed but now this needs a rebase :(

masnax:
@roosterfish Tests are passing now, could you please approve this PR?

roosterfish (approved):
Thanks for removing the sleeps, LGTM!

PR #308, opened 2024-05-14
ceph: Add support for full disk encryption of disks deployed as part of the distributed storage

gabrielmougard:
JIRA ticket: https://warthogs.atlassian.net/browse/LXD-910

We would like to be able to configure Ceph full disk encryption (FDE) as part of a MicroCloud deployment. The existing MicroCeph API already allows to pass an `Encrypt` boolean field to encrypt data at-rest at the disk level through the use of LUKS. Each generated keyring for the associated disk OSD is stored in the internel Ceph configuration folder. 

We propose to introduce a new `--encrypt` flag for the `microcloud init | add` commands to encrypt all the disks selected to be part of the Ceph deployment. We also introduce a way to select the disk to encrypt just like how it is currently done with the disk wipe operation.

One important thing to note is that, Ceph will use `dm_crypt` under the hood to create a virtual encrypted device for each selected disk device. If choosing to encrypt a disk on a certain node, the user must ensure that the `dm_crypt` kernel module is enabled (it is by default enabled for LXD `ubuntu:24.04` VM images : `grep -i dm_crypt config-6.8.0-31-generic` should output `CONFIG_DM_CRYPT=m`)

 ## TODO

* [x] Update MicroCloud integration test suite


ru-fu (commented):
Looks good, but we'll need the link to MicroCeph for enabling this.

gabrielmougard:
@ru-fu updated

gabrielmougard:
@simondeziel the issue shown [here](https://github.com/canonical/microcloud/actions/runs/9191562093/job/25278231667?pr=308#:~:text=grep%20%2Dq%20%27volatile.initial_source%3A%20.*disk1%27) looks similar to the one we saw in Madrid (same failure in the test case) while we discovered a potential race condition in the MicroCloud test suite. But this time, this is part of the preseed test so I don't see why it could be related to a race. As per the interactive test failing, this is an other GH runner timeout in the interactive combination so I don't think we can do much here...

simondeziel:
> But this time, this is part of the preseed test so I don't see why it could be related to a race.

@masnax, does the preseed case actually side-step the disk selection/table creation race?

masnax:
> > But this time, this is part of the preseed test so I don't see why it could be related to a race.
> 
> @masnax, does the preseed case actually side-step the disk selection/table creation race?

Yes, it doesn't use the tables at all

gabrielmougard:
@simondeziel @masnax tests are ok.

cmd/microcloud/add.go
```diff
@@ -17,6 +17,7 @@ type cmdAdd struct {
 
 	flagAutoSetup bool
 	flagWipe      bool
+	flagEncrypt   bool
```
masnax, line 20: I don't think this flag is necessary. It would only be useful for `--auto` but the whole point of `--auto` is that we are deciding for you. So I think it would suffice to either enforce encryption or enforce no encryption in this case.

In the interactive setup we will ask you questions, and the preseed setup has its own key for encryption, so I think it would be safe to drop this.

test/includes/microcloud.sh
```diff
@@ -61,8 +62,9 @@ fi
 if [ -n "${SETUP_ZFS}" ]; then
   setup="${setup}
 ${SETUP_ZFS}                                            # add local disks (yes/no)
-$([ "${SETUP_ZFS}" = "yes" ] && printf "wait 300ms")    # wait for the table to populate
+$([ "${SETUP_ZFS}" = "yes" ] && printf "wait 500ms")    # wait for the table to populate
```
masnax, line 65: I'm not so sure if this should be included here as we've seen from the preseed issue it doesn't seem to be related to the selectable table at all.

test/includes/microcloud.sh
```diff
@@ -904,7 +937,8 @@ create_system() {
       exec > /dev/null
     fi
 
-    lxc init ubuntu-minimal:22.04 "${name}" --vm -c limits.cpu=2 -c limits.memory=4GiB
+    # ubuntu-minimal:24.04 does not execute cloud-init to enable the READY status. Still working on finding why this happens...
+    lxc init ubuntu:24.04 "${name}" --vm -c limits.cpu=2 -c limits.memory=4GiB
```
masnax, line 941: Is this new? How are other PR tests working then?

test/suites/basic.sh
```diff
@@ -105,6 +105,21 @@ test_interactive() {
...
+
+  echo "Creating a MicroCloud with ZFS and Ceph storage, and OVN network"
+  export CEPH_ENCRYPT="yes"
+  microcloud_interactive | lxc exec micro01 -- sh -c "microcloud init > out"
+
+  lxc exec micro01 -- tail -1 out | grep "MicroCloud is ready" -q
+  for m in micro01 micro02 micro03 ; do
+    validate_system_lxd "${m}" 3 disk1 3 1 "${OVN_FILTER}" "${IPV4_SUBNET}" "${IPV4_START}"-"${IPV4_END}" "${IPV6_SUBNET}" "${DNS_ADDRESSES}"
+    validate_system_microceph "${m}" 1 disk2
+    validate_ceph_encrypt "${m}" disk2
```
masnax, line 120: Maybe move `validate_ceph_encrypt` into `validate_system_microceph` with a new argument before or after the `CephFS` argument which indicates whether we expect encryption? That way it will be checked for every test.

masnax (changes requested):
Looks good, just a few changes & questions

test/includes/microcloud.sh
simondeziel: Not sure about the `-minimal` vs `non-minimal` part but 24.04 was picked as it doesn't use the `-kvm` kernel which is lacking the `dm-crypt` module. With 24.04's kernel, all the necessary modules are easily loadable.

masnax: Ah I missed the change to `24.04`.

gabrielmougard: I'm facing an issue with `ubuntu-minimal:24.04` in regards to `cloud-init`: strangely, with the minimal flavor, the `ready.sh` script is not executed:
```
cat << EOF | lxc profile edit default
config:
  cloud-init.user-data: |
    #cloud-config
    write_files:
      - content: |
          #!/bin/sh
          exec curl --unix-socket /dev/lxd/sock lxd/1.0 -X PATCH -d '{"state": "Ready"}'
        path: /var/lib/cloud/scripts/per-boot/ready.sh
        permissions: "0755"
EOF
```

whereas it works fine with `ubuntu:24.04`

simondeziel: @gabrielmougard I cannot reproduce with `ubuntu-minimal-daily:24.04` here:
```
lxc init ubuntu-minimal-daily:24.04 foo --vm -c limits.cpu=2 -c limits.memory=4GiB
cat << EOF | lxc config set foo cloud-init.user-data -
#cloud-config
write_files:
  - content: |
     #!/bin/sh
     exec curl --unix-socket /dev/lxd/sock lxd/1.0 -X PATCH -d '{"state": "Ready"}'
    path: /var/lib/cloud/scripts/per-boot/ready.sh
    permissions: "0755"
EOF
lxc start foo
```

```
$ lxc ls foo
+------+-------+-----------------------+------------------------------------------------+-----------------+-----------+
| NAME | STATE |         IPV4          |                      IPV6                      |      TYPE       | SNAPSHOTS |
+------+-------+-----------------------+------------------------------------------------+-----------------+-----------+
| foo  | READY | 172.24.26.31 (enp5s0) | 2001:470:b1c3:7946:216:3eff:fe0d:bae8 (enp5s0) | VIRTUAL-MACHINE | 0         |
+------+-------+-----------------------+------------------------------------------------+-----------------+-----------+
```

gabrielmougard: Ok then I'll try with `ubuntu-minimal-daily:24.04` and see

simondeziel: `ubuntu-minimal:24.04` worked for me too but since it's a much older image, I think we should switch to `ubuntu-minimal-daily:24.04` if that also works for you.

masnax:
Just a general design question, not sure if it makes sense at all:

Is it reasonable to expect per-disk encryption? Since the data is replicated should the encryption be all-or-nothing and not per-disk?

cmd/microcloud/add.go
gabrielmougard: @masnax Then, for the --auto mode, I would suggest to NOT enforce encryption as this is a quite strong choice that has deep implication on the storage. But I think it also really depends on the philosophy of MicroCloud: what should be part of a default MicroCloud cluster.. are we putting the accent on simplicity or security ? (not to say that both aspects are always mutually exclusive, but in this case I think the question is relevant)

There is maybe also a third solution that is more complex: **encryption by default, when possible**. Which mean that we'll need some system validation logic. For example, is `dm_crypt` loaded in each cluster member containing the to be encrypted disks? If yes, proceed with encryption, else skip with an information message. What do you think?

gabrielmougard:
> Just a general design question, not sure if it makes sense at all:
> 
> Is it reasonable to expect per-disk encryption? Since the data is replicated should the encryption be all-or-nothing and not per-disk?

The MicroCeph API seems to only works with a per-disk encryption basis (see [here](https://canonical-microceph.readthedocs-hosted.com/en/reef-stable/explanation/full-disk-encryption/)). But if I'm following you, do you mean that, among the chosen disks for distributed storage, and if we chose to have encryption, we enforce the fact that either all the disks are encrypted or not at all?

cmd/microcloud/add.go
gabrielmougard: @mionaalex what do you think we should do here?

test/includes/microcloud.sh
gabrielmougard: fixed

test/includes/microcloud.sh
gabrielmougard: fixed

test/suites/basic.sh
gabrielmougard: fixed

masnax:
> > Just a general design question, not sure if it makes sense at all:
> > Is it reasonable to expect per-disk encryption? Since the data is replicated should the encryption be all-or-nothing and not per-disk?
> 
> The MicroCeph API seems to only works with a per-disk encryption basis (see [here](https://canonical-microceph.readthedocs-hosted.com/en/reef-stable/explanation/full-disk-encryption/)). But if I'm following you, do you mean that, among the chosen disks for distributed storage, and if we chose to have encryption, we enforce the fact that either all the disks are encrypted or not at all?


Yes I meant just having a yes/no question for disk encryption that sets the field for every disk payload that we send to the MicroCeph API.

ru-fu (dismissed):
I added a commit to the branch that makes it possible to link directly to the MicroCeph documentation. (Plus some small wording changes.)

tomponline:
> > @simondeziel in the 'basic' test suite (`test_instances_config`) I have a `Error: Failed to run: zpool create -f -m none -O compression=on local /dev/sdb: exit status 1 (/dev/sdb is in use and contains a unknown filesystem.)` .. it seems that somehow a device has not been properly wiped. Do you know from where it could come from?
> 
> IIRC, @masnax ran into this multiple times and a theory was that LXD didn't properly wipe the disks sometimes.
> 
> The "is in use" here is a bit more worrying than not being properly wiped though.

I believe he originally thought that but then found it was a duffrrebt issue

masnax:
> > > @simondeziel in the 'basic' test suite (`test_instances_config`) I have a `Error: Failed to run: zpool create -f -m none -O compression=on local /dev/sdb: exit status 1 (/dev/sdb is in use and contains a unknown filesystem.)` .. it seems that somehow a device has not been properly wiped. Do you know from where it could come from?
> > 
> > 
> > IIRC, @masnax ran into this multiple times and a theory was that LXD didn't properly wipe the disks sometimes.
> > The "is in use" here is a bit more worrying than not being properly wiped though.
> 
> I believe he originally thought that but then found it was a duffrrebt issue

No, it's the same issue, which is that `zpool destroy local` can sometimes silently fail to destroy the zpool. I believe the solution we came to was to update LXD to attempt `zpool destroy -f local` as a fallback, but I haven't actually made a PR for that yet.

tomponline:
> > > > @simondeziel in the 'basic' test suite (`test_instances_config`) I have a `Error: Failed to run: zpool create -f -m none -O compression=on local /dev/sdb: exit status 1 (/dev/sdb is in use and contains a unknown filesystem.)` .. it seems that somehow a device has not been properly wiped. Do you know from where it could come from?
> > > 
> > > 
> > > IIRC, @masnax ran into this multiple times and a theory was that LXD didn't properly wipe the disks sometimes.
> > > The "is in use" here is a bit more worrying than not being properly wiped though.
> > 
> > 
> > I believe he originally thought that but then found it was a duffrrebt issue
> 
> No, it's the same issue, which is that `zpool destroy local` can sometimes silently fail to destroy the zpool. I believe the solution we came to was to update LXD to attempt `zpool destroy -f local` as a fallback, but I haven't actually made a PR for that yet.
> 
> I've submitted a PR for it now: [canonical/lxd#13513](https://github.com/canonical/lxd/pull/13513)

Id like to know more about the scenario and when it occurs before we potentially paper over a separate issue with "-f" incorrectly.

masnax:
> > > > > @simondeziel in the 'basic' test suite (`test_instances_config`) I have a `Error: Failed to run: zpool create -f -m none -O compression=on local /dev/sdb: exit status 1 (/dev/sdb is in use and contains a unknown filesystem.)` .. it seems that somehow a device has not been properly wiped. Do you know from where it could come from?
> > > > 
> > > > 
> > > > IIRC, @masnax ran into this multiple times and a theory was that LXD didn't properly wipe the disks sometimes.
> > > > The "is in use" here is a bit more worrying than not being properly wiped though.
> > > 
> > > 
> > > I believe he originally thought that but then found it was a duffrrebt issue
> > 
> > 
> > No, it's the same issue, which is that `zpool destroy local` can sometimes silently fail to destroy the zpool. I believe the solution we came to was to update LXD to attempt `zpool destroy -f local` as a fallback, but I haven't actually made a PR for that yet.
> > I've submitted a PR for it now: [canonical/lxd#13513](https://github.com/canonical/lxd/pull/13513)
> 
> Id like to know more about the scenario and when it occurs before we potentially paper over a separate issue with "-f" incorrectly.

As discussed here: https://github.com/canonical/microcloud/pull/300#discussion_r1589304021

`zpool destroy <pool>` will sometimes return successfully but fail to actually do anything. In such cases, `zpool destroy -f <pool>` will correctly remove the pool.

tomponline:
> > > > > > @simondeziel in the 'basic' test suite (`test_instances_config`) I have a `Error: Failed to run: zpool create -f -m none -O compression=on local /dev/sdb: exit status 1 (/dev/sdb is in use and contains a unknown filesystem.)` .. it seems that somehow a device has not been properly wiped. Do you know from where it could come from?
> > > > > 
> > > > > 
> > > > > IIRC, @masnax ran into this multiple times and a theory was that LXD didn't properly wipe the disks sometimes.
> > > > > The "is in use" here is a bit more worrying than not being properly wiped though.
> > > > 
> > > > 
> > > > I believe he originally thought that but then found it was a duffrrebt issue
> > > 
> > > 
> > > No, it's the same issue, which is that `zpool destroy local` can sometimes silently fail to destroy the zpool. I believe the solution we came to was to update LXD to attempt `zpool destroy -f local` as a fallback, but I haven't actually made a PR for that yet.
> > > I've submitted a PR for it now: [canonical/lxd#13513](https://github.com/canonical/lxd/pull/13513)
> > 
> > 
> > Id like to know more about the scenario and when it occurs before we potentially paper over a separate issue with "-f" incorrectly.
> 
> As discussed here: [#300 (comment)](https://github.com/canonical/microcloud/pull/300#discussion_r1589304021)
> 
> `zpool destroy <pool>` will sometimes return successfully but fail to actually do anything. In such cases, `zpool destroy -f <pool>` will correctly remove the pool.

Ah ok so we should expect this to reveal the underlying issue rather than fix it. Got you.

masnax:
> Ah ok so we should expect this to reveal the underlying issue rather than fix it. Got you.

Well, not really. It's definitely just hand-waving the issue because we are successfully removing the pool with `--force`. As for *why* that works, and *why* not using `--force` can return without an error but with the pool still existing is a mystery to me.

simondeziel:
> > I've submitted a PR for it now: [canonical/lxd#13513](https://github.com/canonical/lxd/pull/13513)
> 
> Id like to know more about the scenario and when it occurs before we potentially paper over a separate issue with "-f" incorrectly.

I'd be curious to see if there is anything of interest in `zpool events -v <pool>` when the first `zpool destroy <pool>` command fails.

tomponline:
> As for _why_ that works, and _why_ not using `--force` can return without an error but with the pool still existing is a mystery to me.

This is why im hesitant to merge. Doesnt seem like weve got a good enough handle on what's happening yet to make an informed decision either eay.

masnax:
OK, got it to happen with the `test_instances_config` test:

`zpool events -v <pool>` reports over 7000 lines, I've added the last few here because they seem to be the most relevant:

```
May 28 2024 21:21:22.513675449 sysevent.fs.zfs.history_event
        version = 0x0
        class = "sysevent.fs.zfs.history_event"
        pool = "local"
        pool_guid = 0x75804eb4c6f3b15e
        pool_state = 0x0
        pool_context = 0x0
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1"
        history_internal_str = "quota=104857600"
        history_internal_name = "set"
        history_dsid = 0x122
        history_txg = 0xb5
        history_time = 0x66564ad2
        time = 0x66564ad2 0x1e9e10b9
        eid = 0x1ad

May 28 2024 21:21:22.521674776 sysevent.fs.zfs.history_event
        version = 0x0
        class = "sysevent.fs.zfs.history_event"
        pool = "local"
        pool_guid = 0x75804eb4c6f3b15e
        pool_state = 0x0
        pool_context = 0x0
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1"
        history_internal_str = "mountpoint=legacy"
        history_internal_name = "set"
        history_dsid = 0x122
        history_txg = 0xbd
        history_time = 0x66564ad2
        time = 0x66564ad2 0x1f182018
        eid = 0x1ae

May 28 2024 21:21:22.521674776 sysevent.fs.zfs.history_event
        version = 0x0
        class = "sysevent.fs.zfs.history_event"
        pool = "local"
        pool_guid = 0x75804eb4c6f3b15e
        pool_state = 0x0
        pool_context = 0x0
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1"
        history_internal_str = "canmount=2"
        history_internal_name = "set"
        history_dsid = 0x122
        history_txg = 0xbd
        history_time = 0x66564ad2
        time = 0x66564ad2 0x1f182018
        eid = 0x1af

May 28 2024 21:21:22.545672756 sysevent.fs.zfs.history_event
        version = 0x0
        class = "sysevent.fs.zfs.history_event"
        pool = "local"
        pool_guid = 0x75804eb4c6f3b15e
        pool_state = 0x0
        pool_context = 0x0
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1.block"
        history_internal_str = "volmode=2"
        history_internal_name = "set"
        history_dsid = 0x119
        history_txg = 0xc0
        history_time = 0x66564ad2
        time = 0x66564ad2 0x20864e34
        eid = 0x1b0

May 28 2024 21:21:23.061629524 sysevent.fs.zfs.history_event
        version = 0x0
        class = "sysevent.fs.zfs.history_event"
        pool = "local"
        pool_guid = 0x75804eb4c6f3b15e
        pool_state = 0x0
        pool_context = 0x0
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1"
        history_internal_str = "mountpoint=legacy"
        history_internal_name = "set"
        history_dsid = 0x122
        history_txg = 0xc2
        history_time = 0x66564ad3
        time = 0x66564ad3 0x3ac6454
        eid = 0x1b1

May 28 2024 21:21:23.061629524 sysevent.fs.zfs.history_event
        version = 0x0
        class = "sysevent.fs.zfs.history_event"
        pool = "local"
        pool_guid = 0x75804eb4c6f3b15e
        pool_state = 0x0
        pool_context = 0x0
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1"
        history_internal_str = "canmount=2"
        history_internal_name = "set"
        history_dsid = 0x122
        history_txg = 0xc2
        history_time = 0x66564ad3
        time = 0x66564ad3 0x3ac6454
        eid = 0x1b2

May 28 2024 21:21:23.085627583 sysevent.fs.zfs.history_event
        version = 0x0
        class = "sysevent.fs.zfs.history_event"
        pool = "local"
        pool_guid = 0x75804eb4c6f3b15e
        pool_state = 0x0
        pool_context = 0x0
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1.block"
        history_internal_str = "volmode=3"
        history_internal_name = "set"
        history_dsid = 0x119
        history_txg = 0xc4
        history_time = 0x66564ad3
        time = 0x66564ad3 0x51a92bf
        eid = 0x1b3

May 28 2024 21:21:23.109625642 sysevent.fs.zfs.history_event
        version = 0x0
        class = "sysevent.fs.zfs.history_event"
        pool = "local"
        pool_guid = 0x75804eb4c6f3b15e
        pool_state = 0x0
        pool_context = 0x0
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1"
        history_internal_str = "mountpoint=legacy"
        history_internal_name = "set"
        history_dsid = 0x122
        history_txg = 0xc5
        history_time = 0x66564ad3
        time = 0x66564ad3 0x688c12a
        eid = 0x1b4

May 28 2024 21:21:23.109625642 sysevent.fs.zfs.history_event
        version = 0x0
        class = "sysevent.fs.zfs.history_event"
        pool = "local"
        pool_guid = 0x75804eb4c6f3b15e
        pool_state = 0x0
        pool_context = 0x0
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1"
        history_internal_str = "canmount=2"
        history_internal_name = "set"
        history_dsid = 0x122
        history_txg = 0xc5
        history_time = 0x66564ad3
        time = 0x66564ad3 0x688c12a
        eid = 0x1b5

May 28 2024 21:21:24.225536052 sysevent.fs.zfs.history_event
        version = 0x0
        class = "sysevent.fs.zfs.history_event"
        pool = "local"
        pool_guid = 0x75804eb4c6f3b15e
        pool_state = 0x0
        pool_context = 0x0
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1.block"
        history_internal_str = "(bptree, mintxg=1)"
        history_internal_name = "destroy"
        history_dsid = 0x119
        history_txg = 0xc8
        history_time = 0x66564ad4
        time = 0x66564ad4 0xd716834
        eid = 0x1b6

May 28 2024 21:21:24.245534498 sysevent.fs.zfs.history_event
        version = 0x0
        class = "sysevent.fs.zfs.history_event"
        pool = "local"
        pool_guid = 0x75804eb4c6f3b15e
        pool_state = 0x0
        pool_context = 0x0
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1"
        history_internal_str = "(bptree, mintxg=1)"
        history_internal_name = "destroy"
        history_dsid = 0x122
        history_txg = 0xc9
        history_time = 0x66564ad4
        time = 0x66564ad4 0xea28f22
        eid = 0x1b7
```

At this point, the LXD VM has been deleted, and the pool is gone from LXD, but the zpool remains.

masnax:
@simondeziel And here's what I get when I grep for `destroy`:

```
May 28 2024 21:19:02.108750009 sysevent.fs.zfs.pool_destroy
        version = 0x0
        class = "sysevent.fs.zfs.pool_destroy"
--
        pool_state = 0x0
        pool_context = 0x6
        history_hostname = "micro01"
        history_internal_str = "feature@async_destroy=enabled"
--
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1.block"
        history_internal_str = "(bptree, mintxg=1)"
        history_internal_name = "destroy"
--
        history_hostname = "micro01"
        history_dsname = "local/virtual-machines/v1"
        history_internal_str = "(bptree, mintxg=1)"
        history_internal_name = "destroy"
```

simondeziel:
> "feature@async_destroy=enabled"

This seems like a feature we'd probably want to disable prior to doing the `zpool destroy` from LXD. Presumably this would get us the error synchronously.

test/includes/microcloud.sh
simondeziel: Could you give it another try with `ubuntu-minimal-daily:24.04`, please? It was marked as fixed but maybe some rebase lost the change.

test/includes/microcloud.sh
```diff
@@ -932,12 +973,6 @@ setup_system() {
       exec > /dev/null
     fi
 
-    # Disable unneeded services/timers/sockets/mounts (source of noise/slowdown)
-    lxc exec "${name}" -- systemctl mask --now apport.service cron.service e2scrub_reap.service grub-common.service grub-initrd-fallback.service lvm2-monitor.service networkd-dispatcher.service polkit.service secureboot-db.service serial-getty@ttyS0.service ssh.service systemd-journald.service systemd-journal-flush.service unattended-upgrades.service
-    lxc exec "${name}" -- systemctl mask --now apt-daily-upgrade.timer apt-daily.timer dpkg-db-backup.timer e2scrub_all.timer fstrim.timer motd-news.timer update-notifier-download.timer update-notifier-motd.timer
-    lxc exec "${name}" -- systemctl mask --now cloud-init-hotplugd.socket lvm2-lvmpolld.socket lxd-installer.socket iscsid.socket systemd-journald-dev-log.socket
-    lxc exec "${name}" -- systemctl mask --now dev-hugepages.mount sys-kernel-debug.mount sys-kernel-tracing.mount
```
simondeziel, line 939: Removing all of those feels a bit too big of a hammer. Could you try not disabling the `cloud-init` and `lvm2` ones and see if that works?

test/includes/microcloud.sh
```diff
@@ -215,9 +255,11 @@ validate_system_lxd_zfs() {
   grep -q "config: {}" <<< "${cfg}"
   grep -q "status: Created" <<< "${cfg}"
 
+  # In microcloud auto mode, the disk number is not known in advance, we just know it's the first discovered disk.
+  disk_number_pattern="[0-9]+"
```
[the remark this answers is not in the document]
gabrielmougard: fixed

test/includes/microcloud.sh
simondeziel: @gabrielmougard I was suggesting **not** to disable the `cloud-init` and `lvm2` ones :)

test/includes/microcloud.sh
simondeziel: @gabrielmougard how did it go with `ubuntu-minimal-daily:24.04`?

gabrielmougard: fixed

gabrielmougard:
@simondeziel do you know why we see a `Error: Failed instance creation: Failed creating instance from image: Failed to activate volume "f06aa34c14c9863617b1b2b2206bb62ffb085cd9027dc966103565fa6a10b10c"` in the test setup phase here?

test/includes/microcloud.sh
gabrielmougard: ah sorry, on me..

gabrielmougard: let's try with this

simondeziel:
> @simondeziel do you know why we see a `Error: Failed instance creation: Failed creating instance from image: Failed to activate volume "f06aa34c14c9863617b1b2b2206bb62ffb085cd9027dc966103565fa6a10b10c"` in the test setup phase here?

I believe that's a known (recent) regression @MusicDin's already working on fixing in https://github.com/canonical/lxd/pull/13662

simondeziel (approved):
LGTM if the tests are passing.

gabrielmougard:
@MusicDin @simondeziel are the failing tests here related to https://github.com/canonical/lxd/pull/13662 ? We actually don't have much information on why the VMs are failing to start here :/ Is it an other regression observed in LXD?

MusicDin:
The regresion in zfs volume activation should be fixed now. Am I properly interpreting that the tests are failing because the expected status is READY but reported status (by LXD) is RUNNING?

```
  + for round in $(seq 1 5 150)
  ++ lxc list -f csv -c s micro01
  + '[' RUNNING = READY ']'
  + '[' 26 = 0 ']'
  + sleep 5
  + echo '    micro01 VM failed to start'
      micro01 VM failed to start
```

tomponline:
> The regresion in zfs volume activation should be fixed now. Am I properly interpreting that the tests are failing because the expected status is READY but reported status (by LXD) is RUNNING?
> 
> ```
>   + for round in $(seq 1 5 150)
>   ++ lxc list -f csv -c s micro01
>   + '[' RUNNING = READY ']'
>   + '[' 26 = 0 ']'
>   + sleep 5
>   + echo '    micro01 VM failed to start'
>       micro01 VM failed to start
> ```

Should be fixed now in latest/edge https://launchpad.net/~canonical-lxd/lxd/+snap/lxd-latest-edge

tomponline:
wait i take that back...

tomponline:
> wait i take that back...

no it should be fine in latest/edge, but not 5.21/edge or 5.0/edge yet

MusicDin:
> no it should be fine in latest/edge, but not 5.21/edge or 5.0/edge yet

I may be wrong, but this does not seem as the volume activation issue - because at the end of the test it reports that the instance is running, it also gets the IP addresses. Yesterday it was failing with something similar to `Failed to activate volume ..`

```
 ++ lxc list -f csv -c s micro01
  + '[' RUNNING = READY ']'
  + '[' 26 = 0 ']'
  + sleep 5
  + echo '    micro01 VM failed to start'
      micro01 VM failed to start
  + return 1
  + cleanup
  + set -eux
  + lxc remote switch local
  + lxc project switch microcloud-test
  + set +e
  + '[' setup = setup ']'
  + '[' failure = success ']'
  + '[' -n '' ']'
  + lxc list --all-projects
  +-----------------+---------+---------+-----------------------+------------------------------------------------+-----------------+-----------+
  |     PROJECT     |  NAME   |  STATE  |         IPV4          |                      IPV6                      |      TYPE       | SNAPSHOTS |
  +-----------------+---------+---------+-----------------------+------------------------------------------------+-----------------+-----------+
  | microcloud-test | micro01 | RUNNING | 10.57.120.50 (enp5s0) | fd42:51b:91c0:6cc9:216:3eff:fe65:4879 (enp5s0) | VIRTUAL-MACHINE | 0         |
  +-----------------+---------+---------+-----------------------+------------------------------------------------+-----------------+-----------+
```

But the test is expecting READY state, not RUNNING?
https://github.com/canonical/microcloud/blob/main/test/includes/microcloud.sh#L1108-L1131

test/includes/microcloud.sh
gabrielmougard: @simondeziel I can reproduce:
```bash
lxc init ubuntu-minimal-daily:24.04 foo --vm -c limits.cpu=2 -c limits.memory=4GiB
cat << EOF | lxc config set foo cloud-init.user-data -
#cloud-config
write_files:
  - content: |
     #!/bin/sh
     exec curl --unix-socket /dev/lxd/sock lxd/1.0 -X PATCH -d '{"state": "Ready"}'
    path: /var/lib/cloud/scripts/per-boot/ready.sh
    permissions: "0755"
EOF
lxc start foo
```

but the CI still seems to complain even with `ubuntu-minimal-daily:24.04` ...

simondeziel:
> I may be wrong, but this does not seem as the volume activation issue - because at the end of the test it reports that the instance is running, it also gets the IP addresses. Yesterday it was failing with something similar to `Failed to activate volume ..`

Yep, the ZFS activation issue is no longer a blocker (thanks) and it seems like the `add` test failed due to a timeout reaching `READY` state, probably due to a slower than normal runner. I'll try to switch to the bigger runners now if that's quick.

gabrielmougard:
@tomponline @simondeziel after sshing into the runner with tmate (thanks @tomponline for the tip), I noticed that the cloud-init script not setting the VM to a `READY` state issue, was somehow randomly happening (I created many VMs with the same cloud-init profile and some were ever `RUNNING`, some were `READY` but there were just a few)

I noticed that settings `limits.cpu=4` instead of `2` dramatically increased the rate of success though. I also wrapped the `lxd_create / lxd_wait` logic in a short retry loop to eliminate the outliers... It seems to work now so far.

simondeziel:
> I noticed that settings `limits.cpu=4` instead of `2` dramatically increased the rate of success though. I also wrapped the `lxd_create / lxd_wait` logic in a short retry loop to eliminate the outliers... It seems to work now so far.

I'm fine with that especially since we are starting VMs serially (`CONCURRENT_SETUP=0`), this might cause issue to those running them concurrently like @masnax does locally IIRC.

masnax:
> > I noticed that settings `limits.cpu=4` instead of `2` dramatically increased the rate of success though. I also wrapped the `lxd_create / lxd_wait` logic in a short retry loop to eliminate the outliers... It seems to work now so far.
> 
> I'm fine with that especially since we are starting VMs serially (`CONCURRENT_SETUP=0`), this might cause issue to those running them concurrently like @masnax does locally IIRC.

No issue from me, but I hope the 3-node tests continue to be happy with the over-commitment.

masnax (dismissed):
LGTM, rerunning the failed tests once more before merging.

masnax:
@gabrielmougard looks like the test package needs a rebase now.

test/includes/microcloud.sh
```diff
@@ -1036,8 +1081,31 @@ new_system() {
   name=${1}
   num_disks=${2:-0}
 
-  create_system "${name}" "${num_disks}"
-  lxd_wait_vm "${name}"
+  (
```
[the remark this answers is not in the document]
gabrielmougard: @simondeziel I thought putting the logic in a subshell would avoid falling in the `cleanup` triggered by the `trap` at the beginning of the test suite if an error would occur in this retry loop. What could be a clean way to do it without triggering the `cleanup` during the retries?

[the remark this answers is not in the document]
gabrielmougard: @simondeziel it seems that all the tests fails now for the same reasons (VM not READY) in the while being restored.. I guess I need to put this logic in the restore section as well.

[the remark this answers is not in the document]
masnax: I don't think the test failures have to do with this PR. It seems the github runners are failing to start the VMs at all, even on other PRs.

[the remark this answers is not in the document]
simondeziel: Yeah, hopefully https://github.com/canonical/microcloud/pull/335 should make it work again. Thanks to Max for finding the root cause.

masnax:
@gabrielmougard Another rebase, sorry :)

masnax:
@gabrielmougard You need to remove the `systemd-journald` related masks from `microcloud.sh` as that seems to be what's breaking cloud-init.

simondeziel:
> @gabrielmougard You need to remove the `systemd-journald` related masks from `microcloud.sh` as that seems to be what's breaking cloud-init.

Indeed, looks like it was lost in the merge conflict resolution. There should only be one masked `systemd-journald` and that's the `systemd-journal-flush.service` one. The other 2 are gone.

test/includes/microcloud.sh
```diff
@@ -966,9 +1007,9 @@ setup_system() {
     fi
 
     # Disable unneeded services/timers/sockets/mounts (source of noise/slowdown)
-    lxc exec "${name}" -- systemctl mask --now apport.service cron.service e2scrub_reap.service esm-cache.service grub-common.service grub-initrd-fallback.service lvm2-monitor.service networkd-dispatcher.service polkit.service secureboot-db.service serial-getty@ttyS0.service ssh.service systemd-journal-flush.service unattended-upgrades.service
+    lxc exec "${name}" -- systemctl mask --now apport.service cron.service e2scrub_reap.service grub-common.service grub-initrd-fallback.service networkd-dispatcher.service polkit.service secureboot-db.service serial-getty@ttyS0.service ssh.service systemd-journal-flush.service unattended-upgrades.service
```
simondeziel, line 1010: It looks like `esm-cache.service` masking was dropped by accident in the rebase.

masnax:
@gabrielmougard This may need a rebase now, to capture the new microcluster changes.

masnax:
@gabrielmougard Yet another rebase please

test/includes/microcloud.sh
```diff
@@ -245,9 +285,10 @@ validate_system_lxd_zfs() {
   grep -q "config: {}" <<< "${cfg}"
   grep -q "status: Created" <<< "${cfg}"
 
+  # In microcloud auto mode, the disk number is not known in advance, we just know it's the first discovered disk.
   cfg="$(lxc storage show local --target "${name}")"
   grep -q "source: local" <<< "${cfg}"
-  grep -q "volatile.initial_source: .*${local_disk}" <<< "${cfg}"
+  grep -qE "volatile.initial_source: /[^/]+(/[^/]+)*" <<< "${cfg}"
```
masnax, line 291: This can be undone now that the disk ordering issue is fixed, but that can be in another PR.

masnax (approved):
LGTM!

test/includes/microcloud.sh
masnax: Fixed by #346

PR #311, opened 2024-05-31
Update documentation framework

ru-fu:
First step in the cleanup process: Update to the latest version of the starter pack and include the doc targets in the main Makefile.


Makefile
```diff
@@ -49,3 +49,7 @@ update-schema:
 	goimports -w ./database/
 	@echo "Code generation completed"
 
+doc-%:
```
ru-fu, line 52: I would hope there is some Makefile magic that makes these targets show up in tab completion, but I couldn't figure it out. Any ideas @simondeziel ?

simondeziel: AFAIK, the bash completion rules shipped by the `bash-completion` package doesn't support that.

ru-fu: Of course we could manually add the different targets ... but I'm not sure that's a good idea.
For me, it's no big deal to not have autocompletion, since I'm familiar with the targets. What do others think?

simondeziel: Like you I don't think that replicating the target would be worth it.

roosterfish: Never used it but could we use something like `include` https://www.gnu.org/software/make/manual/html_node/Include.html to go into the direction of auto completion? Maybe everything that gets included can be namespaced with `doc-`?

ru-fu: We would definitely need a prefix, yes, because otherwise we'll likely have conflicts with existing targets (`install` for example). But I haven't managed to find a way how to do that so far ...

roosterfish (dismissed):
LGTM

roosterfish (approved):

roosterfish:
@ru-fu this can be rebased now, https://github.com/canonical/microcloud/pull/323 is merged.

ru-fu:
> @ru-fu this can be rebased now, #323 is merged.

Thanks - done!

roosterfish (approved):

PR #323, opened 2024-06-17
test/includes/microcloud: Remove `core.trust_password` support

roosterfish:
This is required after https://github.com/canonical/lxd/pull/13567. Instead of the password we now have to use a token when adding the remote.


simondeziel (dismissed):

simondeziel (dismissed):

roosterfish:
Will mark as draft until I have finalized the testing on my end.

roosterfish:
That is now ready for a final review. I have tested forming LXD clusters in version 5.21 (prior to password removal) and edge.

There was an error in the `HasExtension` check as it will always return `true` if the underlying LXD client has never before done a call to `GetServer()` which will fetch and cache the API extensions. See https://github.com/canonical/lxd/blob/main/client/lxd_server.go#L62.
A fix for this is added to this PR.

roosterfish:
The pipeline is now failing due to a regression caused by https://github.com/canonical/lxd-pkg-snap/pull/471.

service/lxd.go
```diff
@@ -309,6 +309,14 @@ func (s *LXDService) HasExtension(ctx context.Context, target string, address st
 		}
 	}
 
+	// Fill the cache of API extensions.
+	// If the client's internal `server` field isn't yet populated
+	// a call to HasExtension will always return true for any extension.
```
masnax, line 314: Is there a corresponding fix for this in LXD?

masnax (approved):
LGTM, thanks @roosterfish

service/lxd.go
```diff
@@ -200,7 +200,7 @@ func (s LXDService) Join(ctx context.Context, joinConfig JoinConfig) error {
 
 	err = op.WaitContext(ctx)
 	if err != nil {
-		return fmt.Errorf("Failed to configure cluster :%w", err)
+		return fmt.Errorf("Failed to configure cluster: %w", err)
```
simondeziel, line 203: It seems to have turn up a `No matching cluster join operation found` error in https://github.com/canonical/microcloud/actions/runs/9611369723/job/26518376077?pr=323#step:13:58498

That's an error I also saw pop-up from time to time in the LXD test suite. Anyway, not relevant for this PR but would be nice to have that one addressed in LXD ;)

simondeziel (approved):

simondeziel:
> The pipeline is now failing due to a regression caused by [canonical/lxd-pkg-snap#471](https://github.com/canonical/lxd-pkg-snap/pull/471).

This was fixed by https://github.com/canonical/lxd-pkg-snap/pull/476 and now tests are working.

service/lxd.go
masnax: Looks like it might be a race between creating the operation and checking for it.

service/lxd.go
roosterfish: As we discussed earlier I'll follow up on this in LXD.

PR #325, opened 2024-06-20
service: Drop scheme from LXD cluster URLs

masnax:
This makes it consistent with the other services.


service/lxd.go
```diff
@@ -256,7 +256,12 @@ func (s LXDService) clusterMembers(client lxd.InstanceServer) (map[string]string
 
 	genericMembers := make(map[string]string, len(members))
 	for _, member := range members {
-		genericMembers[member.ServerName] = member.URL
+		url, err := url.Parse(member.URL)
```
roosterfish, line 259: Does this break some functionality in already deployed MicroClouds? I guess it will then always use `https:` as a default?

masnax: We currently only use this function to check if a cluster member of a particular name appears after requesting it to join, so this shouldn't break any existing functionality. 

This will be useful in detecting if LXD has already been set up and on the same network as MicroCeph/MicroOVN/MicroCloud.

masnax: And yes the microcloud client always uses HTTPS.

roosterfish: Ok thanks.

roosterfish (approved):

PR #326, opened 2024-06-20
microcloud/api: Don't require mDNS secret for cluster members

masnax:
After the cluster has formed, the nodes should be in each others' truststore so we don't need to rely on the mDNS auth secret for communication anymore.


roosterfish (approved):
Looks good!

PR #327, opened 2024-06-20
Explicitly set bootstrap flag

masnax:
We often infer whether or not we are bootstrapping by checking the `systems` map for the local system. However now that we can re-use clusters from existing systems, we can't always rely on the local member being the bootstrapper. 

Instead, this PR explicitly sets whether or not we are bootstrapping based on the command we are running.

Additionally, there is a commit that cleans up the network questions. Before, we used to apply the FAN network first and then overwrite it with the OVN network. This is a bit inefficient and will break with adding services (#260) so this PR splits the network questions into two functions. First we try to create the OVN network (or return nil). Then if weren't able to do that, we try to create the FAN network.

### DIFF
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -21,14 +21,14 @@ import (
 )
 
 // askRetry will print all errors and re-attempt the given function on user input.
-func (c *CmdControl) askRetry(question string, autoSetup bool, f func() error) {
+func (c *initConfig) askRetry(question string, f func() error) {
 	for {
 		retry := false
 		err := f()
 		if err != nil {
 			fmt.Println(err)
 
-			if !autoSetup {
+			if !c.autoSetup {
 				retry, err = c.asker.AskBool(fmt.Sprintf("%s (yes/no) [default=yes]: ", question), "yes")
 				if err != nil {
 					fmt.Println(err)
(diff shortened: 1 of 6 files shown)


cmd/microcloud/ask.go
```diff
@@ -148,8 +148,7 @@ func (c *CmdControl) askAddress(autoSetup bool, listenAddr string) (string, *net
 	return listenAddr, iface, subnet, nil
 }
 
-func (c *CmdControl) askDisks(sh *service.Handler, systems map[string]InitSystem, autoSetup bool, wipeAllDisks bool) error {
-	_, bootstrap := systems[sh.Name]
+func (c *CmdControl) askDisks(sh *service.Handler, systems map[string]InitSystem, autoSetup bool, wipeAllDisks bool, bootstrap bool) error {
```
MggMuggins, line 151: This is totally a nit, but the sum of these two/three (`autoSetup` and `bootstrap` (and `wipeAllDisks`)) seems like it's being passed around a lot here. It would be more readable at the call site if this were factored into some kind of `AskFlags` structure.

masnax: Yeah, that's true. A lot of this config (even the systems map) should probably be part of the receiver of these methods.

masnax: I've added 2 commits here (though it might not be the best place for it) to add a new `initConfig` struct. 

This holds all the arguments that are passed to every helper, so that they can all be receivers of that type and access the fields.

MggMuggins: Thanks, this looks better.

PR #328, opened 2024-06-22
Include 'lookup_interface' in preseed example and Use the correct ovn interface

WizardBit:
The preseed yaml file used as an example in the [docs](https://canonical-microcloud.readthedocs-hosted.com/en/latest/how-to/initialise/#howto-initialise-preseed) does not include the option 'lookup_subnet'. 

As this preseed example closely aligns with the [get started guide](https://canonical-microcloud.readthedocs-hosted.com/en/latest/tutorial/get_started/) `eth1` was changed  to `enp6s0` to avoid confusion with the device name added to the VM.

Not included in this PR, as I cannot find additional information on the behavior, is  `reuse_existing_clusters` which is also missing.

### DIFF
--- a/doc/how-to/preseed.yaml
+++ b/doc/how-to/preseed.yaml
@@ -1,5 +1,7 @@
 # `lookup_subnet` limits the subnet when looking up systems with mDNS.
 lookup_subnet: 10.0.0.1/24
+# `lookup_interface` limits the interface when looking up systems with mDNS.
+lookup_interface: eth0
 
 # `systems` lists the systems we expect to find by their host name.
 #   `name` represents the host name


doc/how-to/preseed.yaml
```diff
@@ -1,15 +1,17 @@
 # `lookup_subnet` limits the subnet when looking up systems with mDNS.
 lookup_subnet: 10.0.0.1/24
+# `lookup_interface` limits the interface when looking up systems with mDNS.
+lookup_interface: enp5s0
```
roosterfish, line 4: Thanks for this! It should be `eth0`. Please see the other comment.

roosterfish (changes requested):

roosterfish:
@WizardBit thanks for your contribution. 

Your mail `82845986+WizardBit@users.noreply.github.com` is failing the CLA check. Could you please use another one? 

In the commit messages we put a prefix which makes it easier to identify the change in the long list of commits. See https://github.com/canonical/microcloud/commit/48695e6 as an example.

WizardBit:
@roosterfish Thank you for the clarification and glad to help add this missing option.

roosterfish:
Thanks for changing the mail, just one more thing. Can you please squash the two commits together? It is so that you don't add stuff and delete it in the next commit again.

roosterfish (approved):
LGTM!

roosterfish:
> LGTM!

@WizardBit it looks that GitHub doesn't recognize the signing key that you have used. Have you added the key also to https://github.com/settings/keys under "Signing keys"?

PR #332, opened 2024-06-25
Collect system information prior to asking any questions

masnax:
It was getting very hard and inefficient to modify or extend the MicroCloud setup logic without devolving into lots of temporary maps and slices, and redundantly calling back over the network to the other systems to fetch their configuration.

This PR attempts to address that by immediately collecting all relevant system information for the cluster setup in a `SystemInformation` for each system. Each question helper can then reference the fields and methods of the `SystemInformation` to check what is and isn't supported, rather than fetching them over the network each time.

In order to pass around `SystemInformation` as well as other common state information that were previously passed directly as arguments to each function, this PR also adds an `initConfig` type, of which all question functions are a method of. 

Sadly, this does result in quite a large refactor, but given how deeply everything here is coupled together, I don't see a way around it. I've tried to at least split the commits by the functions that are being changed.

Due to this refactor, several edge cases around detecting already configured storage pools and networks have been addressed (tests for these will come later, I didn't want to make this PR any bigger). As well, the ceph cluster network questions are now part of the distributed storage set of questions, rather than the OVN network questions.

### DIFF
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -153,99 +153,15 @@ func (c *initConfig) askAddress() error {
 	return nil
 }
 
-func (c *CmdControl) askDisks(sh *service.Handler, systems map[string]InitSystem, autoSetup bool, wipeAllDisks bool, encryptAllDisks bool) error {
-	_, bootstrap := systems[sh.Name]
-	allResources := make(map[string]*api.Resources, len(systems))
-	var err error
-	for peer, system := range systems {
-		allResources[peer], err = sh.Services[types.LXD].(*service.LXDService).GetResources(context.Background(), peer, system.ServerInfo.Address, system.ServerInfo.AuthSecret)
-		if err != nil {
-			return fmt.Errorf("Failed to get system resources of peer %q: %w", peer, err)
-		}
+func (c *initConfig) askDisks(sh *service.Handler) error {
+	err := c.askLocalPool(sh)
+	if err != nil {
+		return err
 	}
 
-	foundDisks := false
-	for peer, r := range allResources {
-		system := systems[peer]
-		system.AvailableDisks = make([]api.ResourcesStorageDisk, 0, len(r.Storage.Disks))
-		for _, disk := range r.Storage.Disks {
-			if len(disk.Partitions) == 0 {
-				system.AvailableDisks = append(system.AvailableDisks, disk)
-			}
-		}
-
-		if len(system.AvailableDisks) > 0 {
-			foundDisks = true
-		}
-
-		systems[peer] = system
-	}
-
-	wantsDisks := true
-	if !autoSetup && foundDisks {
-		wantsDisks, err = c.asker.AskBool("Would you like to set up local storage? (yes/no) [default=yes]: ", "yes")
-		if err != nil {
-			return err
-		}
-	}
-
-	if !foundDisks {
-		wantsDisks = false
-	}
-
-	lxd := sh.Services[types.LXD].(*service.LXDService)
-	if wantsDisks {
-		c.askRetry("Retry selecting disks?", autoSetup, func() error {
-			return askLocalPool(systems, autoSetup, wipeAllDisks, *lxd)
-		})
-	}
-
-	if sh.Services[types.MicroCeph] != nil {
-		availableDisks := map[string][]api.ResourcesStorageDisk{}
-		for peer, system := range systems {
-			if len(system.AvailableDisks) > 0 {
-				availableDisks[peer] = system.AvailableDisks
-			}
-		}
-
-		if bootstrap && len(availableDisks) < 3 {
-			fmt.Println("Insufficient number of disks available to set up distributed storage, skipping at this time")
-		} else {
-			wantsDisks = true
-			if !autoSetup {
-				wantsDisks, err = c.asker.AskBool("Would you like to set up distributed storage? (yes/no) [default=yes]: ", "yes")
-				if err != nil {
-					return err
-				}
-
-				if len(systems) != len(availableDisks) && wantsDisks {
-					wantsDisks, err = c.asker.AskBool("Unable to find disks on some systems. Continue anyway? (yes/no) [default=yes]: ", "yes")
-					if err != nil {
-						return err
-					}
-				}
-			}
-
-			if wantsDisks {
-				c.askRetry("Retry selecting disks?", autoSetup, func() error {
-					return c.askRemotePool(systems, autoSetup, wipeAllDisks, encryptAllDisks, sh)
-				})
-			}
-		}
-	}
-
-	if !bootstrap {
-		for peer, system := range systems {
-			if len(system.MicroCephDisks) > 0 {
-				if system.JoinConfig == nil {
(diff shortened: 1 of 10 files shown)


service/system_information.go
```diff
@@ -0,0 +1,331 @@
...
+	existingServices := map[types.ServiceType]map[string]string{}
+	for service := range sh.Services {
+		var existingCluster map[string]string
+		if localSystem {
+			existingCluster, err = sh.Services[service].ClusterMembers(ctx)
+		} else {
+			existingCluster, err = sh.Services[service].RemoteClusterMembers(ctx, connectInfo.AuthSecret, connectInfo.Address)
+		}
+
+		if err != nil && err.Error() != "Daemon not yet initialized" && err.Error() != "Server is not clustered" {
```
masnax, line 199: The microcluster half of this error check can use `api.StatusErrorCheck` to check for a `503` once the dependencies are updated, but I'm not sure what to do for the LXD side. Since these are remote requests, it's not ideal to send a whole other request to `GetServer` just to check if the system is clustered, so is there a nice way to detect this particular error from LXD?

@tomponline got any input here?

tomponline: Try using Go client's IsClustered function.

roosterfish:
@masnax ready for review? Then I'll focus on this one next so that we can pave the way for putting the explicit trust establishment into place.

masnax:
> @masnax ready for review? Then I'll focus on this one next so that we can pave the way for putting the explicit trust establishment into place.

This is ready for review, but will likely see rebases once #308 is merged.

masnax:
Looks like there is some sort of cloud-init issue with the test VMs
cc @simondeziel

simondeziel:
> Looks like there is some sort of cloud-init issue with the test VMs cc @simondeziel

I think this PR should be rebased now that #335 was merged.

masnax:
@roosterfish Freshly rebased and ready for review, thanks :)

service/system_information.go
```diff
@@ -0,0 +1,330 @@
...
+		microceph := sh.Services[types.MicroCeph].(*CephService)
+
+		if localSystem {
+			s.CephConfig, err = microceph.ClusterConfig(ctx, "", "")
+		} else {
+			s.CephConfig, err = microceph.ClusterConfig(ctx, s.ClusterAddress, s.AuthSecret)
+		}
+
+		if err != nil && !api.StatusErrorCheck(err, http.StatusServiceUnavailable) {
+			return nil, fmt.Errorf("Failed to get ceph configuration on %q: %w", s.ClusterName, err)
```
roosterfish, line 166: Ceph

service/system_information.go
```diff
@@ -0,0 +1,330 @@
...
+			existingServices[service][k] = host
+		}
+	}
+
+	return existingServices, nil
+}
+
+// SupportsLocalPool checks if the SystemInformation supports a MicroCloud configured local storage pool.
+// Additionally returns whether such a pool already exists.
+func (s *SystemInformation) SupportsLocalPool() (hasPool bool, supportsPool bool) {
```
roosterfish, line 223: What is the difference between `hasPool` and `supportsPool`? 
It's hard to tell based on the three return values.

service/lxd.go
```diff
@@ -424,68 +382,55 @@ func defaultNetworkInterfacesFilter(clients map[string]lxd.InstanceServer, peer
...
+		if filtered {
+			continue
+		}
 
-			candidates[peer] = append(candidates[peer], CephDedicatedInterface{
+		if len(addresses) == 0 {
+			uplinkInterfaces[network.Name] = network
+		} else {
+			cephInterfaces[network.Name] = CephDedicatedInterface{
 				Name:      network.Name,
```
roosterfish, line 489: Isn't the name redundant in here now?

cmd/microcloud/main_init_preseed.go
```diff
@@ -463,59 +467,86 @@ func (p *Preseed) Parse(s *service.Handler, bootstrap bool, lookupTimeout time.D
 		}
 	}
 
+	// If an uplink interface was explicitly chosen, we will try to set up an OVN network.
+	explicitOVN := len(ifaceByPeer) > 0
```
roosterfish, line 471: Can `ifaceByPeer` be > 1?

cmd/microcloud/ask.go
```diff
@@ -1153,55 +1003,148 @@ func (c *CmdControl) askNetwork(sh *service.Handler, systems map[string]InitSyst
...
 				return err
 			}
 
-			if ip.To4() != nil {
-				ipv4Gateway = gateway
-				ipv4Ranges = ipRange
-			} else {
-				ipv6Gateway = gateway
+			if !proceedWithNoOverlayNetworking {
+				return fmt.Errorf("cluster bootstrapping aborted due to lack of usable networking")
```
roosterfish, line 1114: Start error message with uppercase.

roosterfish (changes requested):
This looks good overall. Only a few comments but the restructuring looks solid and makes it much easier in the future to extend maintain the questionnaire.

service/system_information.go
masnax: `hasPool` indicates that the storage pool already exists with a configuration that MicroCloud can work with (so for example a `zfs` storage pool named `local` already exists)

`supportsPool` indicates that LXD is incompatible with the storage pool in the way MicroCloud would set it up (for example there is a `dir` storage pool named `local` so we can't create a `zfs` storage pool named `local`)

cmd/microcloud/main_init_preseed.go
masnax: It's the set of interfaces mapped by system name so it can be anything between 0 and `len(systemsBeingConfigured)`

service/lxd.go
masnax: It may often be convenient to pass this around as a slice instead of a map, that's why the name is in both places, so we are not locked into always passing the map around.

cmd/microcloud/main_init_preseed.go
masnax: In particular, these are the uplink interfaces selected to use for the OVN uplink network. Configuring this is optional so the set can be empty.

service/system_information.go
roosterfish: Ah I see `existingLocalPool` gets set at another place if there is a `local` entry in the respective map.

Thanks clear now.

cmd/microcloud/main_init_preseed.go
roosterfish: Ok thanks.

service/lxd.go
roosterfish: This way you don't have to give up the map structure which is already produced by `GetNetworkInterfaces()`.

roosterfish (approved):
Thanks, looks great! I guess the test failures aren't related to any of the changes in this PR?

masnax:
> Thanks, looks great! I guess the test failures aren't related to any of the changes in this PR?

Yeah, the instance tests appear to be broken with either 24.04 or the new runners.

masnax:
As for the interactive tests, those appear broken due to some other unrelated reason, so I'm going to go ahead and merge this PR now.

PR #334, opened 2024-06-26
github: run system-tests on larger runners (8cores+32G)

simondeziel:
Despite the additional HW resources, the test run-times didn't improved drastically but hopefully this should be more robust. This should also open the door to doing heavier tests without running into space or RAM limits. The CPU overcommit ratio is also better thus reducing the steal time.

Test run-times can be compared with https://github.com/canonical/microcloud/actions/runs/9783672548/job/27012801592

Note: the [System (1.22.x, interactive)](https://github.com/canonical/microcloud/actions/runs/9783893557/job/27019237295?pr=334#logs) still takes ~3 hours and sometimes fail, same as with the regular runners :/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -82,7 +82,7 @@ jobs:
       SKIP_VM_LAUNCH: "1"
       SNAPSHOT_RESTORE: "1"
     name: System
-    runs-on: ubuntu-22.04
+    runs-on: GitHubMicrocloud
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -186,6 +186,9 @@ jobs:
           # we will be able to stop worrying about that special case.
           if [ "$(stat -c '%d' /)" = "$(stat -c '%d' /mnt)" ]; then
             echo "FAIL: rootfs and ephemeral part on the same disk, aborting"
+            lsblk
+            blkid
+            sudo fdisk -l
             exit 1
           fi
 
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -81,7 +81,7 @@ jobs:
       SKIP_VM_LAUNCH: "1"
       SNAPSHOT_RESTORE: "1"
     name: System
-    runs-on: ubuntu-22.04
+    runs-on: GitHubMicrocloud
```
roosterfish, line 84: Just wondering where `GitHubMicrocloud` is setup?

simondeziel: Since those are paid an extra per CI minute, we had to ask IS to assign them to us. They said it was a runner group but didn't share the details on the size of said group.

roosterfish: Ok thanks :)

simondeziel:
Things are way more flaky on 24.04 it seems. The wrong/unexpected disk is often picked causing multiple failures. I'm a bit unsure how to improve that so if anyone has an idea, please let me know.

simondeziel:
A new failure reason:

```
10.223.175.0/24
                                                 # workaround for set -e
'
+ sed '/^\s*#/d; s/\s*#.*//; /^$/d'
Error: Failed to bootstrap local MicroCeph: Failed to run post-bootstrap actions: Failed to start monitor: Failed to run: snapctl start microceph.mon --enable: exit status 1 (error: error running snapctl: snap "microceph" has "auto-refresh" change in progress)
+ cleanup
```

simondeziel:
@masnax thanks for #341 it made it reliable again, apparently even with 24.04 as all tests succeeded on their first try.

masnax (approved):

PR #338, opened 2024-07-04
Experiment: allocate more resources to micro0X instances - do not merge

simondeziel:


masnax:
@simondeziel Is it worth testing this with 24.04 since #308 is going to use 24.04 for the test VMs?

simondeziel:
> @simondeziel Is it worth testing this with 24.04 since #308 is going to use 24.04 for the test VMs?

This PR was really meant as a playground. It seems that bumping the vCPU/RAM doesn't help with the test run-times. As such, I'll leave things as-is in terms of allocation.

I've added the switch to 24.04 in #334, thanks for the suggestion.

PR #341, opened 2024-07-06
cmd/microcloud: Force disk ordering by path name

masnax:
This is part of #332 too but since #334 is having issues with 24.04 VMs, I think it's worth merging now.

This makes it so that when disks are loaded for selection either internally during preseed/auto, or visually in a table with interactive setup, that they are sorted by their path, rather than the order that they appear in the list of resources fetched from LXD.


masnax:
This is blocked by #342

masnax:
@roosterfish mind taking a look at this one? Should fix some inconsistencies in the test suite.

roosterfish (approved):
LGTM!

PR #342, opened 2024-07-06
Update dependencies

masnax:
Both MicroCeph & MicroOVN have updated to the most recent microcluster, so MicroCloud should as well.

Additionally, this fixes the issue with MicroCloud no longer working with `latest/edge` MicroOVN and MicroCeph due to error messages changing.


api/types/version.go
```diff
@@ -0,0 +1,4 @@
+package types
+
+// APIVersion is the current API version.
+const APIVersion = "1.0"
```
roosterfish, line 4: Should this be set to the latest MicroCloud release `1.1`?

roosterfish (changes requested):
One comment on the version rest looks good.

It also requires a version update to `1.22.4` in the `Check compatibility with min Go version (1.22.3)` action.

simondeziel:
> It also requires a version update to `1.22.4` in the `Check compatibility with min Go version (1.22.3)` action.

In LXD, the [`Makefile`](https://github.com/canonical/lxd/blob/main/Makefile#L14) defines `GOMIN` which is then used in the [CI job](https://github.com/canonical/lxd/blob/main/.github/workflows/tests.yml#L245-L249)

api/types/version.go
masnax: I don't believe we are making any changes to the MicroCloud API in this release so no.

masnax: This is the API version that is in the path prefix: like `127.0.0.1:9000/1.0`

roosterfish: True, I somehow mixed it with the versioning @MggMuggins is currently working on in https://github.com/canonical/microcluster/pull/165.

roosterfish (dismissed):
LGTM!

simondeziel (dismissed):

masnax:
@simondeziel Did I miss something? Looks like the `Code` workflow isn't returning?

simondeziel:
> @simondeziel Did I miss something? Looks like the `Code` workflow isn't returning?

Just looked at it and it started "moving" and has made progress now. It was stuck in the `setup-go` action for a while.

masnax:
> > @simondeziel Did I miss something? Looks like the `Code` workflow isn't returning?
> 
> Just looked at it and it started "moving" and has made progress now. It was stuck in the `setup-go` action for a while.

Oh, that was me, sorry for the confusion. I re-ran the `Code` block but it seems to have gotten stuck again.

simondeziel (approved):

roosterfish (approved):

PR #346, opened 2024-07-11
test/includes: Add back specific disk validation

masnax:
Now that we are properly sorting disks, we can reliably validate that the disk ID we expected is the one that was selected automatically.

### DIFF
--- a/test/suites/basic.sh
+++ b/test/suites/basic.sh
@@ -218,7 +218,7 @@ test_instances_config() {
 
   # Setup a MicroCloud with 3 systems, ZFS storage, and a FAN network.
   addr=$(lxc ls micro01 -f csv -c4 | grep enp5s0 | cut -d' ' -f1)
-  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud init --preseed << EOF
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud init --preseed --lookup-timeout 10 << EOF
 lookup_subnet: ${addr}/24
 lookup_interface: enp5s0
 systems:
@@ -253,7 +253,7 @@ EOF
 
   # Create a MicroCloud with ceph and ovn setup.
   addr=$(lxc ls micro01 -f csv -c4 | grep enp5s0 | cut -d' ' -f1)
-  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud init --preseed << EOF
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud init --preseed --lookup-timeout 10 << EOF
 lookup_subnet: ${addr}/24
 lookup_interface: enp5s0
 systems:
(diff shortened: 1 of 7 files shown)


roosterfish (dismissed):

simondeziel (approved):

simondeziel:
Is the CI breakage due to an issue with a recent microceph snap rev in phased rollout?

### Broken

Failing with `Error: System "micro02" failed to join the cluster: Failed to update cluster status of services: Failed to join "MicroCeph" cluster: 1 join attempts were unsuccessful. Last error: json: cannot unmarshal string into Go struct field ClusterMember.extensions of type extensions.Extensions`
```
  microceph (edge) 18.2.0+snapb17a96f27a from Canonical** installed
  Restarted.
  microovn (edge) 24.03.2+snapa14a37d362-dirty from Canonical** installed
  lxd (edge) git-36bca7f from Canonical** installed
  + lxc exec micro01 -- mkdir -p /root/snap/lxd/common/config/
  + lxc exec micro01 -- touch /root/snap/lxd/common/config/config.yml
  + '[' -n '' ']'
  + lxc exec micro01 -- snap install microcloud --channel=latest/edge --cohort=+
  microcloud (edge) git-6a443ab from Canonical** installed
```
  
### Apparently working working
```
  microceph (edge) 18.2.0+snapce68235a3c from Canonical** installed
  Restarted.
  microovn (edge) 24.03.2+snapa14a37d362-dirty from Canonical** installed
  lxd (edge) git-36bca7f from Canonical** installed
  + lxc exec micro01 -- mkdir -p /root/snap/lxd/common/config/
  + lxc exec micro01 -- touch /root/snap/lxd/common/config/config.yml
  + '[' -n '' ']'
  + lxc exec micro01 -- snap install microcloud --channel=latest/edge --cohort=+
  microcloud (edge) git-6a443ab from Canonical** installed
```

masnax:
Hmm, looks like the new runners can get a bit flaky when looking up systems, I'm seeing errors during the auto & preseed tests where microcloud is returning too early and hasn't found all the necessary systems yet.

I've added a `lookup-timeout` flag so that the timeout for mDNS lookups can be managed. We'll probably have to somewhat change this for the "session timeout" with secure join, but it should be fine for now.

test/suites/basic.sh
```diff
@@ -218,7 +218,7 @@ test_instances_config() {
 
   # Setup a MicroCloud with 3 systems, ZFS storage, and a FAN network.
   addr=$(lxc ls micro01 -f csv -c4 | grep enp5s0 | cut -d' ' -f1)
-  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud init --preseed << EOF
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud init --preseed --lookup-timeout 10 << EOF
```
simondeziel, line 221: I think I'd rather have the default value bumped to 10s (from the 5s default).

masnax: Up until this case, we have never seen an issue with finding systems so I felt it might be overkill to force every deployment using `--auto` or `--preseed` to sleep for 10s when 5s appears to be good enough for the majority of cases.

simondeziel: Ah sorry I misinterpreted that as being a regular timeout, not an actual wait time. Ignore my point then.

roosterfish (approved):
LGTM!

PR #351, opened 2024-07-18
Add service-wide cluster member remove command

masnax:
`microcloud remove <name>` will now remove the given cluster member from all services. If `--force` is specified, it is also propagated to all services.

Since the clusters may not have all the same members, we do not error out if some services don't have the corresponding cluster member, instead we will log this as a warning and continue attempting to remove the cluster member from other services.

If no service has the cluster member, then we will error out, reporting that we couldn't find the cluster member across any services.

LXD will be attempted first, as it depends on the other ones. Then we concurrently try to remove the system from MicroCloud, MicroOVN, and MicroCeph.

### DIFF
--- a/api/services_cluster.go
+++ b/api/services_cluster.go
@@ -1,0 +1,100 @@
+package api
+
+import (
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/microcluster/rest"
+	"github.com/canonical/microcluster/state"
+	"github.com/gorilla/mux"
+
+	"github.com/canonical/microcloud/microcloud/api/types"
+	"github.com/canonical/microcloud/microcloud/service"
+)
+
+// ServicesClusterCmd represents the /1.0/services/cluster/{name} API on MicroCloud.
+var ServicesClusterCmd = func(sh *service.Handler) rest.Endpoint {
+	return rest.Endpoint{
+		AllowedBeforeInit: true,
+		Name:              "services/cluster/{name}",
+		Path:              "services/cluster/{name}",
+
+		Delete: rest.EndpointAction{Handler: authHandler(sh, removeClusterMember), AllowUntrusted: true},
+	}
+}
+
+// removeClusterMember removes the given cluster member from all services that it exists in.
+func removeClusterMember(state *state.State, r *http.Request) response.Response {
+	force := r.URL.Query().Get("force") == "1"
+	name, err := url.PathUnescape(mux.Vars(r)["name"])
+	if err != nil {
+		return response.BadRequest(err)
+	}
+
+	supportedServices := map[types.ServiceType]string{
+		types.MicroOVN:  MicroOVNDir,
+		types.MicroCeph: MicroCephDir,
+		types.LXD:       LXDDir,
+	}
+
+	existingServices := []types.ServiceType{types.MicroCloud}
+	for serviceType, stateDir := range supportedServices {
+		if service.Exists(serviceType, stateDir) {
+			existingServices = append(existingServices, serviceType)
+		}
+	}
+
+	addr, _, err := net.SplitHostPort(state.Address().URL.Host)
+	if err != nil {
+		return response.SmartError(fmt.Errorf("State address %q is invalid: %w", state.Address().String(), err))
+	}
+
+	sh, err := service.NewHandler(state.Name(), addr, state.OS.StateDir, false, false, existingServices...)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	// Remove the node from services in the following order:
+	// 1. Remove from LXD first as it may have storage & networks that depend on the others for cleanup.
+	// 2. Remove from MicroCeph and MicroOVN next, concurrently.
+	// 3. Remove from MicroCloud last so that if there were any errors causing the other services to fail, MicroCloud will still know about the node.
+	var memberExists bool
+	err = sh.RunConcurrent(types.LXD, types.MicroCloud, func(s service.Service) error {
+		existingMembers, err := s.ClusterMembers(r.Context())
+		if err != nil && !api.StatusErrorCheck(err, http.StatusServiceUnavailable) {
+			return err
+		}
+
+		// If we got a 503 error back, that means the service is installed, but hasn't been set up yet, so there are no cluster members to remove.
+		if err != nil {
+			return nil
+		}
+
+		// The cluster member may not exist for this service if it was removed manually, so skip removal for that service.
+		_, ok := existingMembers[name]
+		if !ok {
+			logger.Warn("Cluster member not found for service", logger.Ctx{"service": s.Type(), "member": name})
+			return nil
+		}
+
+		if !memberExists && ok {
+			memberExists = ok
+		}
+
+		return s.DeleteClusterMember(r.Context(), name, force)
+	})
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if !memberExists {
+		return response.NotFound(fmt.Errorf("Cluster member %q not found on any service", name))
+	}
+
+	return response.EmptySyncResponse
+}
--- a/cmd/microcloud/remove.go
+++ b/cmd/microcloud/remove.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"context"
+
+	"github.com/canonical/microcluster/microcluster"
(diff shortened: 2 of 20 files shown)


api/services_cluster.go
```diff
@@ -0,0 +1,114 @@
...
+		logger.Warn("Cluster member not found for service", logger.Ctx{"service": lxd.Type(), "member": name})
+	} else {
+		memberExists = ok
+		err = lxd.DeleteClusterMember(r.Context(), name, force)
+		if err != nil {
+			return response.SmartError(err)
+		}
+	}
+
+	err = sh.RunConcurrent(false, true, func(s service.Service) error {
```
MggMuggins, line 82: Should we allow a member to be removed from the `microcloudd` cluster if microceph and microovn may fail to remove? I forsee confusion if `microcloud cluster list` reflects that the cluster member was removed but the member is still present in the Micro{Ceph,OVN} clusters (e.g. user only looks at the `cluster list` from microcloud and never checks the Ceph/OVN ones). Ofc this endpoint will still work to remove a member from the other service's clusters even if its not present in the microcloudd cluster, but that may not be clear.

masnax: Yeah that's a good point, we should remove from LXD first and MicroCloud last.

I was considering making the whole thing sequential and removing the node from MicroCeph second since it has the most restrictions, particularly since it requires `--force` if the remaining cluster size will be less than 3.

But thinking about it more, I don't necessarily think MicroOVN should require `--force` just because MicroCeph requires it, so I think we should still try to cleanly remove the node from MicroOVN concurrently with MicroCeph. 

So If removing a node from a 3-node or smaller cluster, we will always require `--force` to be specified both for MicroCeph and necessarily also MicroCloud since it's tried last.

api/services_cluster.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// ServicesClusterCmd represents the /1.0/services/cluster/{name} API on MicroCloud.
+var ServicesClusterCmd = func(sh *service.Handler) rest.Endpoint {
+	return rest.Endpoint{
+		AllowedBeforeInit: true,
+		Name:              "services/cluster/{name}",
+		Path:              "services/cluster/{name}",
+
+		Delete: rest.EndpointAction{Handler: authHandler(sh, removeClusterMember), AllowUntrusted: true},
```
roosterfish, line 27: A general question about `AllowUntrusted: true`. Does the security in MicroCloud come from the fact that only a root user can connect to the Unix socket to trigger this action?

api/services_cluster.go
```diff
@@ -0,0 +1,100 @@
...
+	existingServices := []types.ServiceType{types.MicroCloud}
+	for serviceType, stateDir := range supportedServices {
+		if service.Exists(serviceType, stateDir) {
+			existingServices = append(existingServices, serviceType)
+		}
+	}
+
+	addr, _, err := net.SplitHostPort(state.Address().URL.Host)
+	if err != nil {
+		return response.SmartError(fmt.Errorf("State address %v is invalid: %w", state.Address().String(), err))
```
roosterfish, line 54: `%q` instead of `%v`?

api/services_cluster.go
```diff
@@ -0,0 +1,100 @@
...
+	err = sh.RunConcurrent(types.LXD, types.MicroCloud, func(s service.Service) error {
+		existingMembers, err := s.ClusterMembers(r.Context())
+		if err != nil && !api.StatusErrorCheck(err, http.StatusServiceUnavailable) {
+			return err
+		}
+
+		// If we got a 503 error back, that means the service is installed, but hasn't been set up yet, so there are no cluster members to remove.
+		if err != nil {
+			return nil
+		}
```
roosterfish, line 76: Would this be a little bit easier to read with a nested `if`? Not a blocker though.

```go
if err != nil {
    if api.StatusErrorCheck(err, http.StatusServiceUnavailable) {
        return nil
    }

    return err
}
```

api/services_cluster.go
```diff
@@ -0,0 +1,100 @@
...
+
+		return s.DeleteClusterMember(r.Context(), name, force)
+	})
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if !memberExists {
+		return response.SmartError(api.StatusErrorf(http.StatusNotFound, "Cluster member %q not found on any service", name))
```
roosterfish, line 96: I saw this the other day, there is a `response.NotFound()` you can use instead.

api/services_cluster.go
roosterfish: You can use `reponse.InternalError()` here.

cmd/microcloud/remove.go
```diff
@@ -0,0 +1,48 @@
...
+	common *CmdControl
+
+	flagForce bool
+}
+
+func (c *cmdRemove) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "remove <name>",
+		Aliases: []string{"rm"},
+		Short:   "Remove the specified system from all MicroCloud services",
```
roosterfish, line 22: Should this say "member" instead of "system"?

The other commands like "add" and "cluster" also talk about members not systems.
Also the description of the force flag below says member.

test/suites/basic.sh
```diff
@@ -1173,3 +1173,149 @@ EOF
...
+  export SETUP_OVN="yes"
+  export OVN_FILTER="enp6s0"
+  export IPV4_SUBNET="10.1.123.1/24"
+  export IPV4_START="10.1.123.100"
+  export IPV4_END="10.1.123.254"
+  export DNS_ADDRESSES="10.1.123.1,8.8.8.8"
+  export IPV6_SUBNET="fd42:1:1234:1234::1/64"
+
+  reset_systems 3 3 3
+  echo "Fail to remove member from MicroCeph and LXD until OSDs are removed"
```
roosterfish, line 1204: Where are the OSDs being removed? I am not really following why the first invocation (without force) of `microcloud remove micro02` is failing or "secured" with `|| true`.
Shouldn't the removal succeed?

test/suites/basic.sh
```diff
@@ -1173,3 +1173,149 @@ EOF
...
+    lxc exec micro01 --env "TEST_CONSOLE=0" -- ${s} cluster list | grep -q "micro01"
+    ! lxc exec micro01 --env "TEST_CONSOLE=0" -- ${s} cluster list | grep -q "micro02" || true
+    lxc exec micro01 --env "TEST_CONSOLE=0" -- ${s} cluster list | grep -q "micro03"
+  done
+
+  for s in "lxc" "microceph" ; do
+    lxc exec micro01 --env "TEST_CONSOLE=0" -- ${s} cluster list | grep -q "micro01"
+    lxc exec micro01 --env "TEST_CONSOLE=0" -- ${s} cluster list | grep -q "micro02"
+    lxc exec micro01 --env "TEST_CONSOLE=0" -- ${s} cluster list | grep -q "micro03"
+  done
```
roosterfish, line 1228: What is happening in between the two loops that the second check for `micro02` doesn't require `|| true`?

test/suites/basic.sh
roosterfish: Ah ok so MicroCeph is the limiting factor here?

roosterfish (changes requested):
Looking very good overall, only a few smaller comments.

api/services_cluster.go
masnax: That's the case for now, I believe we can add an auth handler to these with the secure key exchange mechanism.

api/services_cluster.go
masnax: I'm usually not a big fan of nesting in `if err != nil` as it's easy to overlook when you're not familiar with it.

test/suites/basic.sh
masnax: That's a MicroCeph restriction that does not allow the cluster remove API to proceed if there are OSDs remaining. We call `DeleteClusterMember` for MicroCeph, which triggers its `PreRemove` hook where the restrictions are handled.

api/services_cluster.go
roosterfish: Ok thanks!

test/suites/basic.sh
roosterfish: I missed the loop vars... nvm :)

api/services_cluster.go
masnax: `SmartError` falls back to `InternalError`.

roosterfish: True, I was just thinking that makes it more readable as someone who doesn't know what `SmartError` is doing knows what is effectively being returned here. 

But fine for me as this pattern is also heavily used at other places.

roosterfish (approved):
LGTM!

Issue #148, opened 2023-09-08
microcloud init fails with bridge interface "Error: Failed lookup: write udp6 [::]:<SRC_PORT>->[ff02::fb]:5353: sendto: cannot assign requested address"


tregubovav-dev:
`microcloud init` fails to initialize cluster if bridge interface(s) configured like below:

_# microcloud init
Using address "192.168.92.81" for MicroCloud
Limit search for other MicroCloud servers to 192.168.92.81/23? (yes/no) [default=yes]:
Scanning for eligible servers ...
Error: Failed lookup: write udp6 [::]:43320->[ff02::fb]:5353: sendto: cannot assign requested address_

I'm trying to evaluate microcloud deployment before migrating LXD cluster to microcloud cluster. 

Minimal configuration for reproduction is:
3 nodes (I use Ubuntu 22.04 VMs in LXD cluster on top of RPI4 physical nodes) 
Network config as below (netplan config files are attached):

1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: enp5s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq master br-ext0 state UP group default qlen 1000
    link/ether 00:16:3e:2f:f7:d3 brd ff:ff:ff:ff:ff:ff
3: br-ext0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 0e:15:18:3c:86:c1 brd ff:ff:ff:ff:ff:ff
    inet 192.168.92.81/23 brd 192.168.93.255 scope global br-ext0
       valid_lft forever preferred_lft forever
    inet6 fe80::c15:18ff:fe3c:86c1/64 scope link
       valid_lft forever preferred_lft forever

My suspicion that **microclud init** tries to bind socket to unconfigured physical interface instead of bridge interface.

**Steps to reproduce:**

- Deploy 3 (or more) nodes with Ubuntu 22.04 instances
- Configure network bridge interface on each instance 
- Install microcloud, microovn, microceph and lxd snap packages on each 
[...]

PR #352, opened 2024-07-22
Fix lookup subnet

### LINKED ISSUE #331: Preseed `lookup_subnet` does not work as intended and will only work with host IP
### Summary
`lookup_subnet` has to be the IP address of the host node and not the subnet. This is actually how it is tested in [preseed.sh](https://github.com/canonical/microcloud/blob/5992c468850bea3f4507d5b7e34feb70b9760dba/test/suites/preseed.sh#L92). In this preseed test it would be better to use `lxc network get <network> ipv4.address`  instead. 

It also appears that the interactive method is affected and is using the host IP instead of a subnet to search for other machines and or is phrased incorrectly.

### Preseed
`lookup_subnet` defined in a preseed.
```
lookup_subnet: 10.32.16.1/24
```
This is the error associated.
```
cat seed | microcloud init --preseed
Error: Failed to bootstrap local MicroCloud: Failed to listen on https socket: listen tcp 10.32.16.1:9443: bind: cannot assign requested address
```

### Interactive
In the interactive initialization it mentions the use of the host IP address and the phrasing seems to imply limiting the search to the displayed subnet found.
```
microcloud init
Waiting for LXD to start...
Using address "10.32.16.135" for MicroCloud
Limit search for other MicroCloud servers to 10.32.16.135/24? (yes/no) [default=yes]:  
```

MggMuggins:
Presently, `lookup_subnet` in the [MicroCloud preseed](https://canonical-microcloud.readthedocs-hosted.com/en/latest/how-to/initialise/#non-interactive-configuration) is used to determine the listen address of the services running on the local member. 

Instead, this PR uses the first address assigned to `lookup_interface` that falls within the `lookup_subnet` to determine the listen address for the local services. 

Fixes #331

### DIFF
--- a/cmd/microcloud/main_init_preseed.go
+++ b/cmd/microcloud/main_init_preseed.go
@@ -125,9 +125,19 @@ func (c *initConfig) RunPreseed(cmd *cobra.Command) error {
 		return err
 	}
 
-	ip, _, err := net.ParseCIDR(config.LookupSubnet)
+	_, lookupSubnet, err := net.ParseCIDR(config.LookupSubnet)
 	if err != nil {
 		return err
+	}
+
+	lookupIface, err := net.InterfaceByName(config.LookupInterface)
+	if err != nil {
+		return err
+	}
+
+	listenIP, err := addrInSubnet(lookupIface, *lookupSubnet)
+	if err != nil {
+		return fmt.Errorf("Failed to determine MicroCloud listen address: %w", err)
 	}
 
 	// Build the service handler.
@@ -143,7 +153,7 @@ func (c *initConfig) RunPreseed(cmd *cobra.Command) error {
 	}
 
 	c.name = hostname
-	c.address = ip.String()
+	c.address = listenIP.String()
 	s, err := service.NewHandler(c.name, c.address, c.common.FlagMicroCloudDir, c.common.FlagLogDebug, c.common.FlagLogVerbose, services...)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


roosterfish (commented):
The change in general looks good.

But I am wondering if we actually require the `lookup_interface` config setting.
Let's say you set `lookup_subnet: 10.0.1.1/24`. Now what is the benefit of also setting `lookup_interface`? Wouldn't this be an invalid configuration on the host itself (in regards to routing) if the same network is configured on multiple interfaces?

masnax:
> The change in general looks good.
> 
> But I am wondering if we actually require the `lookup_interface` config setting. Let's say you set `lookup_subnet: 10.0.1.1/24`. Now what is the benefit of also setting `lookup_interface`? Wouldn't this be an invalid configuration on the host itself (in regards to routing) if the same network is configured on multiple interfaces?

I don't think there's anything wrong with having two interfaces where the subnet of one fits within the subnet of another. Since the subnet passed to `lookup_subnet` does not have to match precisely with the subnet on the interface, it can be ambiguous which interface ends up getting used. If we require specifying the interface then there is no ambiguity.

masnax:
Looking at the mDNS logic a little closer, I see this:
* All systems broadcast their information over every capable interface and subnet.
* `lookup_subnet` is used for lookup on the init node. It will find all records from every subnet of all potential joiners.
* From the results of that lookup, we filter the output based on whether it fits in `lookup_subnet` manually.

So the issue is that the `lookup_subnet` doesn't actually have anything to do with the `lookup_interface`. So we will still need two keys.

masnax:
Here's the issue that the introduction of `lookup_interface` was trying to solve. So it looks like setups with interfaces that have conflicting subnets is a real use case: https://github.com/canonical/microcloud/issues/148

masnax:
I don't think any of the points here are blocking for this PR, so I'm going to file an issue for investigating the need for `lookup_interface` and merge this one.

masnax (approved):

PR #356, opened 2024-07-23
test/lint: Add a check for error messages starting with lower case

MggMuggins:
Quick and dirty but hopefully will save some time.

### DIFF
--- a/test/lint/capitalize-errors.sh
+++ b/test/lint/capitalize-errors.sh
@@ -1,0 +1,5 @@
+#!/bin/sh -eu
+
+echo "Checking for error messages beginning with lower-case letters..."
+
+! git grep --untracked -P -n 'fmt\.Errorf\("[a-z]' -- '*.go'


test/lint/capitalize-errors.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh -eu
+
+echo "Checking for error messages beginning with lower-case letters..."
+
+! git grep --untracked -P -n 'fmt\.Errorf\("[a-z]' -- '*.go'
```
roosterfish, line 5: I can at least remember one case in LXD where we concatenate error messages together and have to start the message in lower case. See https://github.com/canonical/lxd/blob/main/lxd/config/errors.go#L19

Also a few lines below you can see [an example](https://github.com/canonical/lxd/blob/main/lxd/config/errors.go#L38) for an error message already starting with a variable.

I think as long as we cannot add something like `//nolint:` it will cause more trouble :(

MggMuggins: This wouldn't catch either of those situations since it's just looking for a lowercase `[a-z]` following `fmt.Errorf("`; it's a pretty blunt instrument. I understand not wanting it to be a pain later but I'm still not convinced that `fmt.Errorf("a` is a sequence of chars that should ever be committed.

roosterfish: You are right, it explicitly checks for those cases where it starts with a `a-z` character.

Let's try this rule then :)

roosterfish (approved):

PR #358, opened 2024-07-25
Clean up join tokens and pending storage & networks on failure

masnax:


roosterfish:
Worth doing a rebase after https://github.com/canonical/microcloud/pull/356 got merged. Otherwise I would suspect errors after merging to main :)

masnax:
@roosterfish This is rebased now. The tests are failing because MicroOVN in `latest/edge` is not currently compatible with MicroCloud due to the move to `v2` microcluster.

To get around this, I've put up this draft PR which is just the commits from this PR plus 3 additional commits to use a locally built MicroOVN snap from the last compatible commit. 

https://github.com/canonical/microcloud/pull/370

cmd/microcloud/main_init.go
```diff
@@ -537,6 +541,16 @@ func (c *initConfig) addPeers(sh *service.Handler) error {
 					}
 				}
 
+				mut.Lock()
```
roosterfish, line 544: Do these locks need to be set within the reverter's func?

roosterfish (changes requested):
Looks good to me, only one comment in regards to the token revert.

roosterfish:
> The tests are failing because MicroOVN in `latest/edge` is not currently compatible with MicroCloud due to the move to `v2` microcluster.

Do you still want to merge it in or wait until the dependencies are settled?

cmd/microcloud/main_init.go
masnax: Moving them into the revert hooks would mean they get evaluated when the `defer` executes. The reverts execute serially, so why would we move them into the hooks?

masnax:
> > The tests are failing because MicroOVN in `latest/edge` is not currently compatible with MicroCloud due to the move to `v2` microcluster.
> 
> Do you still want to merge it in or wait until the dependencies are settled?

Might as well merge it, since I'm not sure when MicroCeph will update their dependency.

cmd/microcloud/main_init.go
roosterfish: Looked at it again. I thought you wanted to lock some variable but instead you lock the reverter itself so that you don't concurrently register the reverts.

roosterfish (approved):
Changes are looking good to me! 

It's unfortunate that the test suite is going havoc now :)

PR #361, opened 2024-08-01
feat: specify an ovn underlay network through MicroOVN

gabrielmougard:
closes https://github.com/canonical/microcloud/issues/239

### DIFF
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -481,7 +481,7 @@ func (c *initConfig) askLocalPool(sh *service.Handler) error {
 	return nil
 }
 
-func validateCephInterfacesForSubnet(lxdService *service.LXDService, systems map[string]InitSystem, availableCephNetworkInterfaces map[string]map[string]service.CephDedicatedInterface, askedCephSubnet string) error {
+func validateCephInterfacesForSubnet(lxdService *service.LXDService, systems map[string]InitSystem, availableCephNetworkInterfaces map[string]map[string]service.DedicatedInterface, askedCephSubnet string) error {
 	validatedCephInterfacesData, err := lxdService.ValidateCephInterfaces(askedCephSubnet, availableCephNetworkInterfaces)
 	if err != nil {
 		return err
@@ -1220,7 +1220,7 @@ func (c *initConfig) askCephNetwork(sh *service.Handler) error {
 		return nil
 	}
 
-	availableCephNetworkInterfaces := map[string]map[string]service.CephDedicatedInterface{}
+	availableCephNetworkInterfaces := map[string]map[string]service.DedicatedInterface{}
 	for name, state := range c.state {
 		if len(state.AvailableCephInterfaces) == 0 {
 			fmt.Printf("No network interfaces found with IPs on %q to set a dedicated Ceph network, skipping Ceph network setup\n", name)
@@ -1228,7 +1228,7 @@ func (c *initConfig) askCephNetwork(sh *service.Handler) error {
 			return nil
 		}
 
-		ifaces := make(map[string]service.CephDedicatedInterface, len(state.AvailableCephInterfaces))
+		ifaces := make(map[string]service.DedicatedInterface, len(state.AvailableCephInterfaces))
 		for name, iface := range state.AvailableCephInterfaces {
 			ifaces[name] = iface
 		}
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -1104,6 +1104,76 @@ func (c *initConfig) askOVNNetwork(sh *service.Handler) error {
 		}
 	}
 
+	canOVNUnderlay := true
+	for peer, system := range c.systems {
+		if len(c.state[system.ServerInfo.Name].AvailableOVNInterfaces) == 0 {
+			fmt.Printf("Not enough interfaces available on %s to create an underlay network, skipping\n", peer)
+			canOVNUnderlay = false
+			break
+		}
+	}
+
+	var ovnUnderlaySelectedIPs map[string]string
+	ovnUnderlayData := [][]string{}
+	for peer, system := range c.systems {
+		// skip any systems that have already been clustered, but are available for other configuration.
+		state, ok := c.state[c.name]
+		if ok {
+			if state.ExistingServices[types.MicroOVN][peer] != "" {
+				continue
+			}
+		}
+
+		for _, net := range c.state[system.ServerInfo.Name].AvailableOVNInterfaces {
+			for _, addr := range net.Addresses {
+				ovnUnderlayData = append(ovnUnderlayData, []string{peer, net.Network.Name, net.Network.Type, addr})
+			}
+		}
+	}
+
+	if len(ovnUnderlayData) != 0 && canOVNUnderlay {
+		wantsDedicatedUnderlay, err := c.asker.AskBool("Configure dedicated underlay networking? (yes/no) [default=no]: ", "no")
+		if err != nil {
+			return err
+		}
+
+		if wantsDedicatedUnderlay {
+			header = []string{"LOCATION", "IFACE", "TYPE", "IP ADDRESS (CIDR)"}
+			fmt.Println("Select exactly one network interface from each cluster member:")
+
+			table = NewSelectableTable(header, ovnUnderlayData)
+			ovnUnderlaySelectedIPs = map[string]string{}
+			err = c.askRetry("Retry selecting underlay network interfaces?", func() error {
+				err = table.Render(table.rows)
+				if err != nil {
+					return err
+				}
+
+				answers, err := table.GetSelections()
+				if err != nil {
+					return err
+				}
+
+				ovnUnderlaySelectedIPs = map[string]string{}
+				for _, answer := range answers {
+					target := table.SelectionValue(answer, "LOCATION")
+					ipAddr := table.SelectionValue(answer, "IP ADDRESS (CIDR)")
+
+					if ovnUnderlaySelectedIPs[target] != "" {
+						return fmt.Errorf("Failed to configure OVN underlay traffic: Selected more than one interface for target %q", target)
+					}
+
+					ovnUnderlaySelectedIPs[target] = ipAddr
+				}
+
+				return nil
+			})
+			if err != nil {
+				return err
+			}
+		}
+	}
+
 	for peer, system := range c.systems {
 		if !askSystems[peer] {
 			continue
@@ -1131,6 +1201,19 @@ func (c *initConfig) askOVNNetwork(sh *service.Handler) error {
 
 		if peer == sh.Name {
 			system.Networks = append(system.Networks, finalConfigs...)
+		}
+
+		if ovnUnderlaySelectedIPs != nil {
+			ovnUnderlayIpAddr, ok := ovnUnderlaySelectedIPs[peer]
+			if ok {
+				ip, _, err := net.ParseCIDR(ovnUnderlayIpAddr)
+				if err != nil {
+					return err
+				}
+
+				fmt.Printf("Using %q for OVN underlay traffic on %q\n", ip.String(), peer)
+				system.OVNGeneveAddr = ip.String()
+			}
 		}
 
 		c.systems[peer] = system
--- a/cmd/microcloud/main_init.go
+++ b/cmd/microcloud/main_init.go
@@ -53,6 +53,9 @@ type InitSystem struct {
 	TargetStoragePools []lxdAPI.StoragePoolsPost
 	// Networks is the cluster-wide network configuration.
 	Networks []lxdAPI.NetworksPost
+	// OVNGeneveAddr represents an IP address to use for the OVN (if OVN is supported) Geneve tunnel on this system.
+	// If left empty, the system will choose to route the Geneve traffic through the management network.
+	OVNGeneveAddr string
 	// StoragePools is the cluster-wide storage pool configuration.
 	StoragePools []lxdAPI.StoragePoolsPost
 	// StorageVolumes is the cluster-wide storage volume configuration.
--- a/cmd/microcloud/main_init.go
+++ b/cmd/microcloud/main_init.go
@@ -527,6 +527,12 @@ func (c *initConfig) addPeers(sh *service.Handler) (revert.Hook, error) {
 			LXDConfig:  info.JoinConfig,
 			CephConfig: info.MicroCephDisks,
 		}
+
+		if info.OVNGeneveAddr != "" {
+			p := joinConfig[peer]
+			p.OVNConfig = map[string]string{"ovn-encap-ip": info.OVNGeneveAddr}
+			joinConfig[peer] = p
+		}
 	}
 
 	clusterSize := map[types.ServiceType]int{}
@@ -782,6 +788,17 @@ func (c *initConfig) setupCluster(s *service.Handler) error {
 
 			if len(microCephBootstrapConf) > 0 {
 				s.SetConfig(microCephBootstrapConf)
+			}
+		}
+
+		if s.Type() == types.MicroOVN {
+			microOvnBootstrapConf := make(map[string]string)
+			if bootstrapSystem.OVNGeneveAddr != "" {
+				microOvnBootstrapConf["ovn-encap-ip"] = bootstrapSystem.OVNGeneveAddr
+			}
+
+			if len(microOvnBootstrapConf) > 0 {
+				s.SetConfig(microOvnBootstrapConf)
 			}
 		}
 
--- a/doc/explanation/microcloud.md
+++ b/doc/explanation/microcloud.md
@@ -61,6 +61,19 @@ MicroCloud will still be usable, but you will see some limitations:
   As a result of this, network forwarding works at a basic level only, and external addresses must be forwarded to a specific cluster member and don't fail over.
 - There is no support for hardware acceleration, load balancers, or ACL functionality within the local network.
 
+(microcloud-networking-underlay)=
+
+Network interface to connect to an OVN underlay network
+:  While setting up the distributed networking with OVN, you can also decide to set up a dedicated underlay network for the OVN traffic.
+   A dedicated underlay network for OVN serves as the physical infrastructure over which the virtual networking (overlay network) is constructed. It can be beneficial to set up a dedicated underlay network for the following reasons:
+
+- A dedicated underlay network isolates the overlay network traffic from other types of traffic (such as management traffic).
+- By dedicating a physical network solely for the purpose of the virtual network overlay, network congestion can be reduced.
+- A dedicated underlay network can be tuned to provide predictable latency and bandwidth characteristics, which are crucial for the performance of latency-sensitive applications running over the OVN overlay.
+- A dedicated underlay can be designed to scale independently of other networks, which allows for more efficient scaling of the overlay network as the demand for virtual network resources increases.
+
+See {ref}`howto-ovn-underlay` for how to set up a dedicated OVN underlay network.
+
 ### Dedicated internal network for Ceph
 You can set up a dedicated network for Ceph to separate the internal Ceph traffic from the rest of the MicroCloud cluster traffic.
 
--- a/doc/how-to/initialise.md
+++ b/doc/how-to/initialise.md
@@ -90,6 +90,13 @@ Complete the following steps to initialise MicroCloud:
       You must select one network interface per machine.
    1. If you want to use IPv4, specify the IPv4 gateway on the uplink network (in CIDR notation) and the first and last IPv4 address in the range that you want to use with LXD.
    1. If you want to use IPv6, specify the IPv6 gateway on the uplink network (in CIDR notation).
+   1. If you chose to set up distributed networking, you can choose to setup an underlay network for the distributed networking:
+
+      If you choose ``yes``, configure the underlay network:
+
+      1. Select the network interfaces that you want to use (see {ref}`microcloud-networking-underlay`).
+
+         You must select one network interface with an IP address per machine.
 1. MicroCloud now starts to bootstrap the cluster.
    Monitor the output to see whether all steps complete successfully.
    See {ref}`bootstrapping-process` for more information.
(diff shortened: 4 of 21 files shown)


cmd/microcloud/ask.go
```diff
@@ -1039,6 +1039,81 @@ func (c *initConfig) askOVNNetwork(sh *service.Handler) error {
...
+			for peer, ipAddr := range selected {
+				system := c.systems[peer]
+				ip, _, err := net.ParseCIDR(ipAddr)
+				if err != nil {
+					return err
+				}
+
+				system.OVNGeneveAddr = ip.String()
+				c.systems[peer] = system
+			}
```
masnax, line 1113: This block seems to be where we update the systems list. Could you please move it down to the block starting at line 1117 so that all the assignments are in one place? Just to make extending these functions a bit easier.

service/lxd.go
```diff
@@ -468,11 +468,18 @@ type CephDedicatedInterface struct {
 	Addresses []string
 }
 
+// OVNDedicatedInterface represents a dedicated interface for OVN.
+type OVNDedicatedInterface struct {
+	Network   api.Network
```
masnax, line 473: Could this type get merged with CephDedidatedInterface? Seems the only difference is one includes the whole `api.Network` while the other just includes the `Type`.

masnax (changes requested):
Looks good, hope the refactor wasn't too annoying. I Just have a few nits.

doc/explanation/microcloud.md
```diff
@@ -61,6 +61,19 @@ MicroCloud will still be usable, but you will see some limitations:
...
+Network interface to connect to an OVN underlay network
+:  While setting up the distributed networking with OVN, you can also decide to set up a dedicated underlay network for the OVN traffic.
+   A dedicated underlay network for OVN serves as the physical infrastructure over which the virtual networking (overlay network) is constructed. It can be beneficial to set up a dedicated underlay network for the following reasons:
+
+- A dedicated underlay network isolates the overlay network traffic from other types of traffic (such as management traffic).
+- By dedicating a physical network solely for the purpose of the virtual network overlay, network congestion can be reduced.
+- A dedicated underlay network can be tuned to provide predictable latency and bandwidth characteristics, which are crucial for the performance of latency-sensitive applications running over the OVN overlay.
+- A dedicated underlay can be designed to scale independently of other networks, which allows for more efficient scaling of the overlay network as the demand for virtual network resources increases.
+
+See {ref}`howto-ovn-underlay` for how to set up a dedicated OVN underlay network.
```
ru-fu, line 75: Wrong indentation - this must be indented as much as the "While/A" paragraph.

doc/explanation/microcloud.md
```diff
@@ -61,6 +61,19 @@ MicroCloud will still be usable, but you will see some limitations:
   As a result of this, network forwarding works at a basic level only, and external addresses must be forwarded to a specific cluster member and don't fail over.
 - There is no support for hardware acceleration, load balancers, or ACL functionality within the local network.
 
+(microcloud-networking-underlay)=
+
+Network interface to connect to an OVN underlay network
```
ru-fu, line 66: Are we talking about a network interface here, or about a network? It sounds like this is a network, unlike the two network interfaces mentioned further up.

If that is true, this formatting is confusing, since it implies that we're talking about a third network interface here.
Instead, turn this into a normal subsection instead of a definition list (in the same way as the internal Ceph network).

If we are talking about a network interface, this one should go into the same list as the other two.

doc/explanation/microcloud.md
ru-fu: (Only if we keep the definition list of course.)

doc/how-to/initialise.md
```diff
@@ -90,6 +90,13 @@ Complete the following steps to initialise MicroCloud:
       You must select one network interface per machine.
    1. If you want to use IPv4, specify the IPv4 gateway on the uplink network (in CIDR notation) and the first and last IPv4 address in the range that you want to use with LXD.
    1. If you want to use IPv6, specify the IPv6 gateway on the uplink network (in CIDR notation).
+   1. If you chose to set up distributed networking, you can choose to setup an underlay network for the distributed networking:
```
ru-fu, line 93:
```suggestion
   1. If you chose to set up distributed networking, you can choose to set up an underlay network for the distributed networking:
```

doc/how-to/ovn_underlay.md
```diff
@@ -0,0 +1,83 @@
+(howto-ovn-underlay)=
+# How to configure an OVN underlay network
+
+When running {command}`microcloud init`, if you decided to setup distributed networking and if you have enough at least one unused network interface per cluster member with an IP address, you are asked if you want to configure an underlay network for OVN.
```
ru-fu, line 4:
```suggestion
When running {command}`microcloud init`, if you decided to set up distributed networking and if you have at least one unused network interface with an IP address per cluster member, you are asked if you want to configure an underlay network for OVN.
```

doc/how-to/ovn_underlay.md
```diff
@@ -0,0 +1,83 @@
...
+Here is the question you will be asked:
+
+`Configure dedicated underlay networking? (yes/no) [default=no]: <answer>`
+
+You can choose to skip this question (just hit `Enter`). This will then use the internal MicroCloud network as an OVN 'underlay', which is the same as the OVN management network ('overlay' network).
```
ru-fu, line 9:
```suggestion
If you answer `no` to this question (just hit `Enter`), the internal MicroCloud network is used as an OVN *underlay*, which is the same as the OVN management network (*overlay* network).
```

doc/how-to/ovn_underlay.md
```diff
@@ -0,0 +1,83 @@
...
+
+You could also choose to configure a dedicated underlay network for OVN with typing `yes`. A list of available network interfaces with an IP address will be displayed.
```
ru-fu, line 11:
```suggestion
Enter `yes` to configure a dedicated underlay network for OVN. A list of available network interfaces with an IP address will be displayed.
```

doc/how-to/initialise.md
```diff
@@ -90,6 +90,13 @@ Complete the following steps to initialise MicroCloud:
...
+
+      If you choose ``yes``, configure the underlay network:
+
+      1. Select the network interfaces that you want to use (see {ref}`microcloud-networking-underlay`).
+
+         You must select one network interface with an IP address per machine.
```
ru-fu, line 99:
```suggestion
      If you choose `yes`, select the network interfaces that you want to use for the underlay network (see {ref}`microcloud-networking-underlay`).
      You must select one network interface with an IP address per machine.
```

ru-fu: No need to add substeps if there is only one step.

doc/how-to/ovn_underlay.md
```diff
@@ -0,0 +1,83 @@
...
+           lxc network create ovnbr0
+
+   1. Enter the following commands to find out the assigned IPv4 and IPv6 addresses for the networks and note them down:
+
+           lxc network get ovnbr0 ipv4.address
+           lxc network get ovnbr0 ipv6.address
+
+1. Create the network interfaces that will be used for the OVN underlay setup for each VM:
+
+   1. Add the network device for the `ovnbr0` network:
```
ru-fu, line 29: No need for a substep here - just combine the two intros into one.

doc/how-to/ovn_underlay.md
```diff
@@ -0,0 +1,83 @@
...
+
+           lxc config device add micro1 eth2 nic network=ovnbr0 name=eth2
+           lxc config device add micro2 eth2 nic network=ovnbr0 name=eth2
+           lxc config device add micro3 eth2 nic network=ovnbr0 name=eth2
+           lxc config device add micro4 eth2 nic network=ovnbr0 name=eth2
+
+1. Now, just like in the tutorial, start the VMs.
+1. On each VM, bring the network interfaces up and give them an IP address within their network subnet:
+
+   1. For the `ovnbr0` network, do the following for each VM::
```
ru-fu, line 39: Same comment, combine them.

doc/how-to/ovn_underlay.md
```diff
@@ -0,0 +1,83 @@
...
+             [x]  | micro3   | enp7s0 | physical | 10.0.1.4/24                               |
+             [ ]  | micro3   | enp7s0 | physical | fd42:5782:5902:5b9e:216:3eff:fe36:d29c/64 |
+           > [x]  | micro2   | enp7s0 | physical | 10.0.1.3/24                               |
+             [ ]  | micro2   | enp7s0 | physical | fd42:5782:5902:5b9e:216:3eff:fedb:f04e/64 |
+                  +----------+--------+----------+-------------------------------------------+
+
+1. The MicroCloud initialisation process will now continue as usual and the OVN cluster will be configured with the underlay network you provided.
+1. You can now inspect the OVN underlay setup:
+
+   1. Inspect the OVN southbound encapsulation parameters:
```
ru-fu, line 67: Again, combine.

ru-fu (changes requested):

doc/explanation/microcloud.md
gabrielmougard: fixed

doc/explanation/microcloud.md
gabrielmougard: fixed

doc/how-to/initialise.md
gabrielmougard: fixed

doc/how-to/ovn_underlay.md
gabrielmougard: fixed

doc/how-to/ovn_underlay.md
gabrielmougard: fixed

doc/how-to/ovn_underlay.md
gabrielmougard: fixed

doc/how-to/initialise.md
gabrielmougard: fixed

doc/how-to/ovn_underlay.md
gabrielmougard: fixed

doc/how-to/ovn_underlay.md
gabrielmougard: fixed

doc/how-to/ovn_underlay.md
gabrielmougard: fixed

cmd/microcloud/ask.go
gabrielmougard: fixed

service/lxd.go
gabrielmougard: fixed

cmd/microcloud/main_init.go
```diff
@@ -757,6 +766,20 @@ func (c *initConfig) setupCluster(s *service.Handler) error {
...
 			}
 
+			if s.Type() == types.MicroOVN {
+				microOvnBootstrapConf := make(map[string]string)
+				for _, system := range c.systems {
+					// If the system needs to be bootstrapped with a custom ovn encapsulation ip for geneve, add it to the init config.
+					if system.OVNGeneveAddr != "" {
+						microOvnBootstrapConf["ovn-encap-ip"] = system.OVNGeneveAddr
+					}
+				}
```
masnax, line 776: Maybe the test failures have to do with this?

Looks like you're iterating over the `c.systems` map and selecting an effectively random address for the geneve address. I think you might want to grab the config from `bootstrapSystem` like with `MicroCephInternalNetworksubnet`.

masnax (changes requested):

cmd/microcloud/main_init.go
gabrielmougard: fixed

gabrielmougard:
@masnax it seems to solve the issue. But now the interactive combination tests seem to fail at some point.

masnax:
@gabrielmougard Needs a rebase.

cmd/microcloud/ask.go
```diff
@@ -1096,6 +1096,71 @@ func (c *initConfig) askOVNNetwork(sh *service.Handler) error {
...
+			header = []string{"LOCATION", "IFACE", "TYPE", "IP ADDRESS (CIDR)"}
+			fmt.Println("Select exactly one network interface from each cluster member:")
+			data = [][]string{}
+			for peer, state := range c.state {
+				for _, net := range state.AvailableOVNInterfaces {
+					for _, addr := range net.Addresses {
+						data = append(data, []string{peer, net.Network.Name, net.Network.Type, addr})
+					}
+				}
+			}
```
masnax, line 1128: `c.state` is a list of all existing and prospective cluster members, which means the table will show entries for nodes that may have been already configured during `microcloud init`. Since the config is only applied on bootstrap/join, the new configuration will be ignored.

If you switched this to `c.systems`, you would instead iterate over only the new systems. But that would introduce the opposite problem when running `microcloud add` of what to do if the user initially selected "no" for underlay network setup when they ran `microcloud init`. 

Ideally, we should be iterating over the list of all systems (existing or new) which have not set a geneve address. Problem is I don't see a way to fetch the geneve address from MicroOVN over the API. 

Could you speak to the MicroOVN team about adding a way to query the nb/sb databases over the API? Or at least add the geneve address to `microovn status`, since it's included as a config in `microovn init`?

cmd/microcloud/main_init.go
```diff
@@ -764,6 +773,17 @@ func (c *initConfig) setupCluster(s *service.Handler) error {
...
 		}
 
+		if s.Type() == types.MicroOVN {
+			microOvnBootstrapConf := make(map[string]string)
+			if bootstrapSystem.OVNGeneveAddr != "" {
+				microOvnBootstrapConf["ovn-encap-ip"] = bootstrapSystem.OVNGeneveAddr
+			}
+
+			if len(microOvnBootstrapConf) > 0 {
+				s.SetConfig(microOvnBootstrapConf)
```
masnax, line 783: So we only apply the underlay config during bootstrap/join. When running `microcloud add` or `microcloud service add`, other nodes may not have underlay networking configured, so even though we present a list of every old and new cluster member, no config will change.

I guess this means we also need a way to set the geneve address over the API. 

We could get around this for now if we at least have a way to _get_ the current underlay configuration. We could just skip configuration in the `add` cases if it was not previously configured during `init`.

masnax (dismissed):
Sadly it looks like we're pretty limited by MicroOVN about what we can do during the `microcloud add` and `microcloud service add` case, so I'm going to approve this even though the issues (https://github.com/canonical/microcloud/pull/361#discussion_r1729520490 and https://github.com/canonical/microcloud/pull/361#discussion_r1729529722) exist. 

I'll add an issue to track those, but we can merge this one :)

cmd/microcloud/ask.go
mkalcok: Is there an option for microcloud to query OVN directly? MicroOVN provides [content interfaces](https://github.com/canonical/microovn/blob/0dc300879ab2063d0531811c15325217e8f3bcc6/snap/snapcraft.yaml#L15-L30) that should enable you to talk directly to the OVN/OVS. If you need overall cluster information, you'd need to query equivalent of
```
microovn.ovn-sbctl list encap
```
Example output:
```
_uuid               : a3096d41-e53e-422a-ba50-2fec3542d9ad
chassis_name        : movn1
ip                  : "10.75.224.231"
options             : {csum="true"}
type                : geneve
```

If you are interested only about the local chassis, you could run
```
microovn.ovs-vsctl get open_vswitch . external_ids:ovn-encap-ip
```
Example output:
```
"10.75.224.231"
```

masnax: @gabrielmougard so in the loop above, you should iterate on `c.systems` instead of `c.state`. 

Also, we need to skip underlay configuration during `microcloud service add` because in that case `c.state == c.systems`. So you'll have to check if they are already clustered:
```go
for peer, systems := range c.systems {
   // skip any systems that have already been clustered, but are available for other configuration.
   if c.state[c.name].ExistingClusters[types.MicroOVN][peer] == "" {
        continue
    }
}
```

masnax:
@gabrielmougard it looks like the `basic` tests aren't passing because the `test_add_services` set of tests don't have an expected answer for the `ovn_underlay_network`. 

With the changed behaviour, `microcloud init` should ask about the underlay network, but running `microcloud service add` should skip it, so in each test case you need to set and unset the environment variables accordingly. 

It would be helpful to also add a test to cover the case where you run `microcloud add` in the following situation:

* `microcloud init` was run to set up 2 nodes with MicroOVN uninstalled (so setup for it was skipped)
* MicroOVN was then installed on both systems (but not initialized)
* `microovn cluster bootstrap` was run on a "new" 3rd node
* `microcloud add` was called to add this new 3rd node.

In this case, we should verify that MicroCloud offers `micro01` and `micro02` to add their underlay configuration, but skips `micro03` since it was already set up, even though technically `micro03` is the one joining the overall MicroCloud cluster.

masnax:
@gabrielmougard Actually I noticed the question for underlay networking still gets asked even though there won't be any eligible systems when running `service add` or `add` with preconfigured MicroOVN.

Can you move the iteration over `c.systems` up above from where we ask if the user wants to set up underlay networking, and exit early if the resulting size of `data` is 0?

cmd/microcloud/ask.go
```diff
@@ -1104,6 +1104,83 @@ func (c *initConfig) askOVNNetwork(sh *service.Handler) error {
...
+					ipAddr := table.SelectionValue(answer, "IP ADDRESS (CIDR)")
+
+					if ovnUnderlaySelectedIPs[target] != "" {
+						return fmt.Errorf("Failed to configure OVN underlay traffic: Selected more than one interface for target %q", target)
+					}
+
+					ovnUnderlaySelectedIPs[target] = ipAddr
+				}
+
+				if len(ovnUnderlaySelectedIPs) != len(c.state) {
```
masnax, line 1172: This check is no longer valid, since we can't guarantee all systems will be configured.

cmd/microcloud/ask.go
```diff
@@ -1104,6 +1104,83 @@ func (c *initConfig) askOVNNetwork(sh *service.Handler) error {
...
+			for peer := range c.systems {
+				// skip any systems that have already been clustered, but are available for other configuration.
+				state, ok := c.state[c.name]
+				if ok {
+					if state.ExistingServices[types.MicroOVN][peer] != "" {
+						continue
+					}
+				}
+
+				for _, net := range state.AvailableOVNInterfaces {
```
masnax, line 1140: You're iterating on the variable `state` which you instantiate above to always point to `c.state[c.name]` so all entries in the table are getting the same address.

You need `c.systems[peer].AvailableOVNInterfaces`, for this loop, not `state.AvailableOVNInterfaces`.

cmd/microcloud/ask.go
```diff
@@ -1104,6 +1104,83 @@ func (c *initConfig) askOVNNetwork(sh *service.Handler) error {
...
+		if err != nil {
+			return err
+		}
+
+		if wantsDedicatedUnderlay {
+			header = []string{"LOCATION", "IFACE", "TYPE", "IP ADDRESS (CIDR)"}
+			fmt.Println("Select exactly one network interface from each cluster member:")
+			data = [][]string{}
+
+			for peer := range c.systems {
```
masnax, line 1131: The loop that populates `data` should be moved above the question asking if the user wants to configure the underlay network.

If the resulting set of rows (stored in `data`) is empty, then we can skip the question entirely, since we can't set anything up.

masnax (changes requested):
I've highlighted the blocks that are resulting in test failures. You'll also need to change `test_add_services` a bit because each test case in that set of tests will need the variable to be set/unset, and we should add a case (you can put it in `test_reuse_cluster`) which ensures we _only_ set up the underlay network for not-previously-configured systems.

cmd/microcloud/ask.go
gabrielmougard: > @gabrielmougard so in the loop above, you should iterate on `c.systems` instead of `c.state`.
> 
> Also, we need to skip underlay configuration during `microcloud service add` because in that case `c.state == c.systems`. So you'll have to check if they are already clustered:
> 
> ```go
> for peer, systems := range c.systems {
>    // skip any systems that have already been clustered, but are available for other configuration.
>    state, ok := c.state[c.name] 
>    if ok {
>       if state.ExistingClusters[types.MicroOVN][peer] != "" {
>          continue
>       }
>    }
> }
> ```

should be fixed now @masnax

gabrielmougard:
> @gabrielmougard it looks like the `basic` tests aren't passing because the `test_add_services` set of tests don't have an expected answer for the `ovn_underlay_network` when calling `service add`.
> 
> With the changed behaviour, `microcloud init` should ask about the underlay network, but running `microcloud service add` should skip it, so in each test case you need to set and unset the environment variables accordingly.
> 
> It would be helpful to also add a test to cover the case where you run `microcloud add` in the following situation:
> 
> * `microcloud init` was run to set up 2 nodes with MicroOVN uninstalled (so setup for it was skipped)
> * MicroOVN was then installed on both systems (but not initialized)
> * `microovn cluster bootstrap` was run on a "new" 3rd node
> * `microcloud add` was called to add this new 3rd node.
> 
> In this case, we should verify that MicroCloud offers `micro01` and `micro02` to add their underlay configuration, but skips `micro03` since it was already set up, even though technically `micro03` is the one joining the overall MicroCloud cluster.

I added the test case @masnax Tell me what you think about it ;)

cmd/microcloud/ask.go
gabrielmougard: fixed

cmd/microcloud/ask.go
gabrielmougard: fixed

masnax (approved):
The preseed logic seems to be incorrect, but it's a small fix and since you're going away for a couple weeks, I'll merge this and make a follow-up PR. I'll tag you so you can see how the preseed IPs should be determined.

We also need to update the preseed tests for the new keys, and add the new preseed key to the docs as well.

PR #362, opened 2024-08-01
refactor: update to microcluster v2

gabrielmougard:
Test to see if we can use microcluster/v2 for MicroCloud


masnax:
This will likely not work exactly in the way we expect until microceph & microovn are updated to the new version too.

This is because the API endpoint paths and payloads that the client functions from `v2` use are not the same as the `v1` daemon serves. 

If this is just a test, then you might be able to work around it by keeping both the `v1` and `v2` modules imported, and then using the `v1` module in all the places where we reach out to microceph & microovn.

PR #367, opened 2024-08-07
Fix `microcloud add` storage & network questions

masnax:
If storage & networks have not been set up on existing cluster members, `microcloud add` should try to set them up for both new cluster members and old ones. 

This can be accomplished by simply populating the `InitSystem`s for all existing cluster members.

This also simplifies `setupCluster` because it no longer has different behaviour based on the `bootstrap` flag being set on `initConfig`. 

---
Some other cleanup:
* The profile update question has been moved to before cluster formation because the flow is awkward when the user has to wait a few minutes and then gets asked another question before the setup continues. 
* Standardizes ellipses in messages to be of the form `text ...`

### DIFF
--- a/cmd/microcloud/add.go
+++ b/cmd/microcloud/add.go
@@ -146,6 +146,33 @@ func (c *cmdAdd) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Also populate system information for existing cluster members. This is so we can potentially set up storage and networks if they haven't been set up before.
+	for name, address := range state.ExistingServices[types.MicroCloud] {
+		_, ok := cfg.systems[name]
+		if ok {
+			continue
+		}
+
+		cfg.systems[name] = InitSystem{
+			ServerInfo: mdns.ServerInfo{
+				Name:     name,
+				Address:  address,
+				Services: services,
+			},
+		}
+
+		if name == cfg.name {
+			continue
+		}
+
+		state, err := s.CollectSystemInformation(context.Background(), mdns.ServerInfo{Name: name, Address: address})
+		if err != nil {
+			return err
+		}
+
+		cfg.state[name] = *state
+	}
+
 	err = cfg.askDisks(s)
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


cmd/microcloud/add.go
```diff
@@ -146,6 +146,33 @@ func (c *cmdAdd) Run(cmd *cobra.Command, args []string) error {
...
+
+		cfg.systems[name] = InitSystem{
+			ServerInfo: mdns.ServerInfo{
+				Name:     name,
+				Address:  address,
+				Services: services,
+			},
+		}
+
+		if name == "" || name == cfg.name {
```
roosterfish, line 164: Not sure if I saw this pattern somewhere else in the code. 
In which situation do we have an empty entry in `state.ExistingServices`?

roosterfish: Ah, saw this comment later in another commit: `// There may be existing cluster members that are not a part of MicroCloud, so ignore those.`

Is that the root cause? Can you please add a comment here too?

roosterfish (changes requested):
LGTM, just one question in regards to skipping empty systems.

cmd/microcloud/add.go
masnax: You're right, `name == ""` is unnecessary here because this map is populated with the output of the cluster members fetching API. I've fixed this now.

The comment you're quoting is something else: basically if we encounter "new" MicroCloud cluster members who are already part of one pre-existing MicroOVN or MicroCeph cluster, then we will try to add OSDs for those systems as well, but only if we have `mdns.ServerInfo` existing for them, which means they were included in the questionnaires. 


For example:
```
pre-init:
microcloud: none
microovn:   none
microceph:  m2, m3, m4
lxd:        none

post-init:
microcloud: m1, m2, m3
microovn:   m1, m2, m3
microceph:  m1, m2, m3, m4 -- add OSDs for each of m1, m2, m3, but skip m4
lxd:        m1, m2, m3
```

roosterfish: I see thanks for the explanation!

roosterfish (approved):

PR #370, opened 2024-08-09
Tests for #358

masnax:

PR #371, opened 2024-08-12
Set core https address

### LINKED ISSUE #214: Different `core.https_address` on init member
When deploying a new MicroCloud cluster, the node on which `microcloud init` gets executed has a different setting for `core.https_address` compared to its cluster peers. It is using the wildcard pattern `[::]:8443` unlike its peers which use the primary IP of the machine `{IP}:8443`.  

```bash
julian@thinkpad:~$ lxc exec m1 -- lxc config get core.https_address
[::]:8443
julian@thinkpad:~$ lxc exec m2 -- lxc config get core.https_address
10.146.93.224:8443
julian@thinkpad:~$ lxc exec m3 -- lxc config get core.https_address
10.146.93.54:8443

```

MggMuggins:
Fixes #214 

I'd prefer to mark this ready for review and rebase to main once we've pushed Microcluster v2 so that the tests can appear happy, but if @masnax would like this merged sooner I can make that happen. Let me know.

### DIFF
--- a/service/lxd.go
+++ b/service/lxd.go
@@ -193,6 +193,21 @@ func (s LXDService) Join(ctx context.Context, joinConfig JoinConfig) error {
 		return fmt.Errorf("Failed to configure cluster: %w", err)
 	}
 
+	// Set the local server's core.https_address to be consistent with the
+	// bootstrap member's wildcard
+	currentServer, etag, err := client.GetServer()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


masnax:
Just a heads up, the tests are going to fail without [the commits mentioned here](https://github.com/canonical/microcloud/pull/370#issuecomment-2276984864) because of MicroOVN using `v2` microcluster while MicroCeph & MicroCloud are still at `v1`.

PR #372, opened 2024-08-13
MicroCloud doc integration

ru-fu:
See https://docs.google.com/document/d/1mSGNMG2J58b494bdSuGvN_UeNgTt5MLk_Z8fV2G2K44/edit?usp=sharing for an explanation.


doc/Makefile
```diff
@@ -26,5 +28,61 @@ help:
         "* other possible targets:                    make <TAB twice> \n" \
         "------------------------------------------------------------- \n"
 
+integrate:
+	# Pull the other repositories
+	cd integration/ && \
+	git -C lxd fetch && git -C lxd reset --hard origin/main || git clone https://github.com/canonical/lxd
```
MggMuggins, line 34: As discussed we should pull a version that matches our compatibility matrix (once that gets added), as it may not be immediately obvious to a user that they haven't been redirected to https://documentation.ubuntu.com/lxd.

Not clear yet how we will specify compatible versions in the `microcloud` executable (RE version enforcement), but it would be ideal if those versions were used here to determine the correct branch. Once again, not a blocker.

Could you open an issue to track this concern with any additional context?

doc/.readthedocs.yaml
```diff
@@ -22,8 +29,8 @@ sphinx:
   fail_on_warning: true
 
 # If using Sphinx, optionally build your docs in additional formats such as PDF
-formats:
-   - pdf
+#formats:
+#   - pdf
```
MggMuggins, line 33: Are pdf builds just not being used/exposed?

doc/index.md
```diff
@@ -6,20 +6,52 @@ relatedlinks: https://snapcraft.io/microcloud
...
+: MicroOVN is a snap-based distribution of [OVN](https://www.ovn.org/).
+  MicroCloud uses MicroOVN to set up OVN networking.
+
+```{note}
+The MicroCloud documentation set is targeted specifically at users of MicroCloud.
+
+The other three documentation set describe the full functionality of each component.
+This functionality is available as part of your MicroCloud setup, but not all of it is relevant.
+For example, all documentation sets contain installation information, but the components are already installed as part of MicroCloud.
+Also, while each component documents how to downscale its cluster, you should not remove machines from one of the components only, but remove them from MicroCloud.
```
MggMuggins, line 49: How about "Also, while each component documents how to remove cluster members, you should not remove machines from only one component. Use MicroCloud to remove cluster members."?

Maybe add a link to "How to remove a machine".

MggMuggins (changes requested):
This is really awesome, thanks! Only a few nits from me.

doc/.sphinx/_integration/header.css
```diff
@@ -0,0 +1,179 @@
+.p-navigation {
```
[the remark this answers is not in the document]
ru-fu: I opened https://github.com/canonical/microcloud/issues/373 - opening an issue with the web team now would probably lead to confusion.

doc/.sphinx/_integration/lxd.html
```diff
@@ -0,0 +1,43 @@
+<header id="header" class="p-navigation">
```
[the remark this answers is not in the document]
ru-fu: I agree, but I can't really think of a way that would be easier.
We could either add a bunch of variables to each config file and use them in the template, or add a script that generates the files for each doc set. Both is possible, but I'm not sure it would actually make things easier to understand.

doc/Makefile
ru-fu: I opened https://github.com/canonical/microcloud/issues/374, and updated the Makefile to specify the versions at the top.

doc/.readthedocs.yaml
ru-fu: This disables PDF builds.

Since we are combining the documentation, the default PDF build probably won't give us a useful result.  (I mean, the [existing build](https://canonical-microcloud.readthedocs-hosted.com/_/downloads/en/latest/pdf/) doesn't either ... :wink: )
It should be feasible to get this to work, but it'll need some looking into.

doc/index.md
ru-fu: Updated.

doc/.sphinx/_integration/lxd.html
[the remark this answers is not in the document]
MggMuggins: Given the circumstances I don't think this is a priority atm and you're likely right about the added complexity.

MggMuggins (dismissed):
Looks good, thanks!

doc/Makefile
```diff
@@ -30,12 +34,14 @@ help:
...
 integrate:
 	# Pull the other repositories
-	cd integration/ && \
-	git -C lxd fetch && git -C lxd reset --hard origin/main || git clone https://github.com/canonical/lxd
-	cd integration/ && \
-	git -C microceph fetch && git -C microceph reset --hard origin/main || git clone https://github.com/canonical/microceph
-	cd integration/ && \
-	git -C microovn fetch && git -C microovn reset --hard origin/main || git clone https://github.com/canonical/microovn
+	cd integration/ && ( git -C lxd fetch || git clone https://github.com/canonical/lxd )
+	git -C integration/lxd reset --hard $(LXDVERSION)
```
roosterfish, line 38: Not a blocker as the outcome is the same but I am wondering if we should use `git checkout` instead of `git reset --hard` which alters the history of this local copy.

It gives the impression you are trying to reset something whereas you only want to track the branch/commit specified in the env variables:

`git -C integration/lxd checkout $(LXDVERSION)`

roosterfish (approved):
Looks great, only one small non-blocking remark to the repo cloning.

doc/Makefile
ru-fu: This works when cloning the repo, and makes perfect sense there.
But if we have the repos cloned already and only want to update them, it might fail (because we change some files in the repos - the header and the config files - and if anything changes upstream as well, the checkout will fail).

Open for any suggestions to do this in a cleaner way!

roosterfish: Ah ok, I was missing the fact that the repo is already there and has received some changes. All fine from my side then.

doc/.readthedocs.yaml
```diff
@@ -9,11 +9,18 @@ version: 2
...
-      - cd doc && python3 .sphinx/build_requirements.py
+  commands:
       - git fetch --unshallow || true
+      - cd doc && make integrate
+      - cd doc/integration/lxd/ && go build -ldflags "-s -w" -o trimpath -o lxc.bin ./lxc
+      # Pretend that woke is installed - we don't need it for building
+      # (workaround until https://github.com/canonical/microovn/pull/168 is merged
+      # and https://github.com/canonical/microceph/pull/400 is restored)
+      - ln -s /bin/sh doc/integration/microovn/docs/woke
+      - ln -s /bin/sh doc/integration/microceph/docs/woke
```
simondeziel, line 22: This type of kludge typically use a less feature-full binary like `true`:
```suggestion
      - ln -s /bin/true doc/integration/microovn/docs/woke
      - ln -s /bin/true doc/integration/microceph/docs/woke
```

PR #375, opened 2024-08-16
Build microovn from last compatible version

masnax:
Until MicroCeph updates its microcluster dependency to `v2`, we're stuck using microcluster `v1` in MicroCloud which is incompatible with `v2` because the core endpoint prefixes have changed.

To work around this in the mean time, this adds a block to the github workflows to build MicroOVN from the last compatible commit. The built snap is then supplied to the test suite.

This is actually easier to clean up than trying to fudge both `v1` and `v2` microcluster into MicroCloud, which would require some larger refactors. Once we're ready to move to `v2`, we can just remove the block from the workflow.


test/main.sh
```diff
@@ -132,9 +132,12 @@ export MICROCEPH_SNAP_CHANNEL
 MICROCLOUD_SNAP_CHANNEL="${MICROCLOUD_SNAP_CHANNEL:-latest/edge}"
 export MICROCLOUD_SNAP_CHANNEL
 
-MICROOVN_SNAP_CHANNEL="${MICROOVN_SNAP_CHANNEL:-latest/edge}"
+MICROOVN_SNAP_CHANNEL="${MICROOVN_SNAP_CHANNEL:-22.03/stable}"
```
roosterfish, line 135: Do we still need this env variable as MicroOVN is now built from source?

test/suites/basic.sh
```diff
@@ -826,8 +826,8 @@ test_service_mismatch() {
 
   # Install microceph and microovn on the first machine only.
   for m in micro02 micro03 ; do
-    lxc exec "${m}" -- snap remove microceph --purge
-    lxc exec "${m}" -- snap remove microovn --purge
+    lxc exec "${m}" -- snap disable microceph
```
roosterfish, line 829: Can it cause some leftovers if we just disable/enable the snaps? 
Why not running ` snap install --dangerous "/root/microovn.snap"` here?

roosterfish (changes requested):
I like the idea so we can unblock the test suite for the time being. Just two small comments.

test/main.sh
masnax: Yes, the recommended way to install a built snap is to install the default snap first and then use `--dangerous`. This is to preserve any snap interfaces that won't be set up normally with just `--dangerous` or `--devmode`.

test/suites/basic.sh
masnax: It's overall quicker to simply disable/re-enable the snaps, which is what we do all over the test suite. This one spot was an anomaly.

test/main.sh
roosterfish: I see thanks!

roosterfish (commented):
Might the pipeline errors be related to https://github.com/golangci/golangci-lint/issues/4909?

masnax:
> Might the pipeline errors be related to [golangci/golangci-lint#4909](https://github.com/golangci/golangci-lint/issues/4909)?

Sigh, Added a line to comment out golangci-lint in `snapcraft.yaml`

roosterfish (approved):

PR #378, opened 2024-08-19
Allow OSD selection with fewer than 3 systems

masnax:
Depends on #367 

Allows setting up remote `ceph` storage with fewer than 3 systems supplying disks.

With interactive init, when selecting disks for Ceph, a warning will be displayed indicating the setup is not fault-tolerant:
```
Disk configuration does not meet recommendations for fault tolerance. At least 3 systems must supply disks.
Retry selecting disks? (yes/no) [default=yes]:
```

After all members have joined the cluster, we verify that the spread of OSDs is at least 3, and if it is not, then we update all pool sizes and set the default OSD replication factor. 

For preseed, both OVN and Ceph setup no longer require 3 systems. Some small changes were made to the disk filtering logic, namely that `find_min` will default to 1 for ZFS storage since we can verify that all systems supply a disk. For Ceph, `find_min` must be explicitly set because we can't know if the user actually wants a non-HA setup, or if they are okay with some systems being fully remote for ceph.

### DIFF
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -76,7 +76,7 @@ func (c *initConfig) askUpdateProfile(profile api.ProfilesPost, profiles []strin
 }
 
 // askRetry will print all errors and re-attempt the given function on user input.
-func (c *initConfig) askRetry(question string, f func() error) {
+func (c *initConfig) askRetry(question string, f func() error) error {
 	for {
 		retry := false
 		err := f()
@@ -86,14 +86,13 @@ func (c *initConfig) askRetry(question string, f func() error) {
 			if !c.autoSetup {
 				retry, err = c.asker.AskBool(fmt.Sprintf("%s (yes/no) [default=yes]: ", question), "yes")
 				if err != nil {
-					fmt.Println(err)
-					retry = false
+					return err
 				}
 			}
 		}
 
 		if !retry {
-			break
+			return nil
 		}
 	}
 }
@@ -149,7 +148,7 @@ func (c *initConfig) askAddress() error {
 			}
 
 			table := NewSelectableTable([]string{"ADDRESS", "IFACE"}, data)
-			c.askRetry("Retry selecting an address?", func() error {
+			err := c.askRetry("Retry selecting an address?", func() error {
 				fmt.Println("Select an address for MicroCloud's internal traffic:")
 				err := table.Render(table.rows)
 				if err != nil {
@@ -171,6 +170,9 @@ func (c *initConfig) askAddress() error {
 
 				return nil
 			})
+			if err != nil {
+				return err
+			}
 		} else {
 			fmt.Printf("Using address %q for MicroCloud\n", listenAddr)
 		}
@@ -327,7 +329,7 @@ func (c *initConfig) askLocalPool(sh *service.Handler) error {
 	}
 
 	if !c.autoSetup {
-		c.askRetry("Retry selecting disks?", func() error {
+		err := c.askRetry("Retry selecting disks?", func() error {
 			selected := map[string]string{}
 			sort.Sort(cli.SortColumnsNaturally(data))
 			header := []string{"LOCATION", "MODEL", "CAPACITY", "TYPE", "PATH"}
@@ -386,6 +388,9 @@ func (c *initConfig) askLocalPool(sh *service.Handler) error {
 
 			return nil
 		})
+		if err != nil {
+			return err
+		}
 	}
 
 	if len(selectedDisks) == 0 {
@@ -634,7 +639,7 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
 			return nil
 		}
 
-		c.askRetry("Retry selecting disks?", func() error {
+		err = c.askRetry("Retry selecting disks?", func() error {
 			header := []string{"LOCATION", "MODEL", "CAPACITY", "TYPE", "PATH"}
 			data := [][]string{}
 			for peer, disks := range availableDisks {
@@ -727,6 +732,9 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
 
 			return nil
 		})
+		if err != nil {
+			return err
+		}
 
 		if len(selectedDisks) == 0 {
 			return nil
@@ -946,7 +954,7 @@ func (c *initConfig) askOVNNetwork(sh *service.Handler) error {
 
 	table := NewSelectableTable(header, data)
 	var selectedIfaces map[string]string
-	c.askRetry("Retry selecting uplink interfaces?", func() error {
+	err = c.askRetry("Retry selecting uplink interfaces?", func() error {
 		err := table.Render(table.rows)
 		if err != nil {
 			return err
@@ -977,6 +985,9 @@ func (c *initConfig) askOVNNetwork(sh *service.Handler) error {
 
 		return nil
 	})
+	if err != nil {
+		return err
+	}
 
 	for peer, iface := range selectedIfaces {
 		fmt.Printf(" Using %q on %q for OVN uplink\n", iface, peer)
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -582,7 +582,6 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
 			askSystemsRemoteFS[info.ClusterName] = true
 		}
 	}
-
 	var selectedDisks map[string][]string
 	var wipeDisks map[string]map[string]bool
 	availableDiskCount := 0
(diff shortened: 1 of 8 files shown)


masnax:
Initially, I was setting the OSD replication factor based on the number of systems that supply OSDs. After speaking with the MicroCeph team and reading their [docs](https://canonical-microceph.readthedocs-hosted.com/en/latest/explanation/cluster-scaling/#failure-domains), it seems that the failure domain handling changes internally once the cluster hits 3 members, and so it makes more sense to simply set the replication factor to be the the absolute number of OSDs, up to a maximum of 3.

cmd/microcloud/ask.go
```diff
@@ -715,8 +709,9 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
 				targetDisks[target] = append(targetDisks[target], path)
 			}
 
-			if !useJoinConfigRemote && len(targetDisks) < 3 {
-				return fmt.Errorf("Unable to add remote storage pool: At least 3 peers must have allocated disks")
+			insufficientDisks := !useJoinConfigRemote && len(targetDisks) < recommendedDiskSpread
+			if c.autoSetup && insufficientDisks {
```
roosterfish, line 713: Is this producing a warning (like mentioned in the commit message) or does it re-ask the question if I select fewer 3 disks? Same for the check in interactive mode below.

cmd/microcloud/ask.go
```diff
@@ -732,6 +727,10 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
 
 			selectedDisks = targetDisks
 
+			if insufficientDisks {
+				return fmt.Errorf("Disk configuration does not meet recommendations for fault tolerance. At least %d systems must supply disks.", recommendedDiskSpread)
```
roosterfish, line 731: In case of non-interactive I think the error message should also mention that this it not recommended?

cmd/microcloud/main_init.go
```diff
@@ -805,23 +805,47 @@ func (c *initConfig) setupCluster(s *service.Handler) error {
...
+
+		allDisks, err := cephClient.GetDisks(context.Background(), c)
+		if err != nil {
+			return err
+		}
+
+		if len(allDisks) > 0 {
+			defaultPoolSize := len(allDisks)
+			if defaultPoolSize > 3 {
+				defaultPoolSize = 3
```
roosterfish, line 845: If somebody wants to have a higher replication factor than 3, should we allow configuring this too?
Or would you see this to be an action performed on MicroCeph directly if required?

cmd/microcloud/main_init_preseed.go
```diff
@@ -311,11 +301,6 @@ func (p *Preseed) validate(name string, bootstrap bool) error {
 		}
 	}
 
-	usingOVN := p.OVN.IPv4Gateway != "" || p.OVN.IPv6Gateway != "" || containsUplinks
```
roosterfish, line 314: Is the OVN check intended to be removed here too?

roosterfish: There is also another OVN check removed down below.

roosterfish (changes requested):
Looks good. 

From the changes I understood you now don't require Ceph disks on three different systems but for example if I select three disks on `m1` or one on `m1` and two on `m3` it still looks to be a hard requirement to have at least one disk on every member?

cmd/microcloud/ask.go
```diff
@@ -76,24 +76,25 @@ func (c *initConfig) askUpdateProfile(profile api.ProfilesPost, profiles []strin
...
 
 // askRetry will print all errors and re-attempt the given function on user input.
-func (c *initConfig) askRetry(question string, f func() error) {
+func (c *initConfig) askRetry(question string, f func() error) error {
 	for {
 		retry := false
 		err := f()
 		if err != nil {
-			fmt.Println(err)
+			if err.Error() != "" {
```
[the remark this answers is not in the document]
masnax: Actually, I got rid of the `fmt.Errorf("")` pattern I mentioned to you.

I left this check in because if the error is ever empty for some reason, we'll print a new line with `fmt.Println`.

cmd/microcloud/ask.go
masnax: `askRetry` skips the question and exits if `c.autoSetup` is true.

With `--auto`, if your systems don't have enough resources to set up at least 3 disks across 3 systems, then `remote` storage pool setup will be skipped entirely with that message.

cmd/microcloud/ask.go
masnax: Good point, I'll add a log to the preseed path.

cmd/microcloud/main_init.go
masnax: After speaking to the MicroCeph team, I think this should be a very purposeful change, and is only for specific use-cases, so I don't think it belongs in the MicroCloud questionnaire.

cmd/microcloud/main_init_preseed.go
masnax: Yes, there was aslso a 3 system restriction in preseed mode for OVN setup, so these were removed for that purpose. `usingOVN` is redefined elsewhere for other use cases.

masnax:
> From the changes I understood you now don't require Ceph disks on three different systems but for example if I select three disks on `m1` or one on `m1` and two on `m3` it still looks to be a hard requirement to have at least one disk on every member?

I wouldn't describe it as a "hard requirement", it's just a warning.

Our recommended setup will be 3 OSD hosts minimum. You can still set up MicroCloud with 0-2 OSD hosts with any number of disks. Any less than 3 OSD hosts and we will report that the fault tolerance recommendations have not been met, and ask the user if they want to retry. In the example you gave, there would be 3 disks selected, but only 2 OSD hosts, so the user would have to confirm their selection after the warning.

That actually makes me wonder if we should have a second, scarier message that mentiones data loss risk if they only select 1 disk (so no replication at all).

cmd/microcloud/ask.go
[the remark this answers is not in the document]
roosterfish: But wouldn't it be better to print the newline instead of vanishing the error? Then you have at least an indicator (the newline) that something is off. 

To me it sounds like an empty error would be something you don't want to have so we should not silently suppress it.

cmd/microcloud/ask.go
roosterfish: Mh but in interactive mode when I select three disks on fewer than three systems I'll get the `Disk configuration does not meet recommendations for fault tolerance` message as it returns the error to `askRetry()`. 

To be sure I just checked out your branch and tried myself. Wasn't aware of the fact that you have to say `no` after getting the message and it will then continue on with your selection of fewer than three systems.

Would it make sense to add another sentence saying you can continue on but this wouldn't protect you from potential data loss?
At least I got confused by it :)

roosterfish (changes requested):
One more comment to the warning and error message.

cmd/microcloud/ask.go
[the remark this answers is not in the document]
masnax: Sure, I'll remove the check.

cmd/microcloud/ask.go
masnax: Sure, I'll add an additional line. It'll be weird since the error will have a `\n` in it but that should be OK.

cmd/microcloud/ask.go
```diff
@@ -725,8 +724,15 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
 
 			selectedDisks = targetDisks
 
+			if insufficientDisks {
+				return fmt.Errorf("Disk configuration does not meet recommendations for fault tolerance. At least %d systems must supply disks.", RecommendedOSDHosts)
```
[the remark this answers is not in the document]
masnax: Fixed.

cmd/microcloud/ask.go
masnax: The error is now printed like this:
```
Disk configuration does not meet recommendations for fault tolerance. At least 3 systems must supply disks.
Continuing with this configuration will leave MicroCloud susceptible to data loss
Change disk selection? (yes/no) [default=yes]:  
```

roosterfish: Thanks, looks good to me!

roosterfish (approved):

cmd/microcloud/main_init.go
```diff
@@ -805,23 +808,47 @@ func (c *initConfig) setupCluster(s *service.Handler) error {
...
+			}
 
-			logger.Debug("Adding disk to MicroCeph", logger.Ctx{"name": name, "disk": disk.Path})
-			_, err = cephClient.AddDisk(context.Background(), client, &disk)
+			err = cephClient.PoolSetReplicationFactor(context.Background(), c, &cephTypes.PoolPut{Pools: []string{"*"}, Size: int64(defaultPoolSize)})
```
roosterfish, line 851: @masnax looks like this comment didn't make it (I thought I have saved it), but here I was wondering if we should limit this to the `lxd_remote` and the other pool used for CephFS.

In case you are using an already existing MicroCeph cluster, there might be already other pools configured besides the ones used by LXD/MicroCloud?

PR #383, opened 2024-08-26
Add explicit trust establishment

roosterfish:
This PR adds the explicit trust establishment to MicroCloud according to the spec in https://discourse.ubuntu.com/t/explicit-trust-establishment-mechanism-for-microcloud/44261. 

The PR adds the EFF wordlist with over 1300 lines so please consider this for the review.
The license information is based on https://www.eff.org/copyright.

It has the following dependencies which have to be met before it can be marked as ready:
* LXD: https://github.com/canonical/lxd/pull/13969
* MicroCluster: https://github.com/canonical/microcluster/pull/239

In addition we are currently waiting on MicroCeph to merge the `v2` of MicroCluster which is also are requirement for this PR.

Open TODO's:

- [x] Align test suite to work with updated questions
- [x] Release minor version of MicroCluster to allow pulling the new websocket client functions in (we will track the commit for now)
- [x] Update this PR's `go.mod` file if the dependencies are met
- [x] Check for additional unit testing capabilities for the changes in this PR
- [x] Remove the TODO (regarding https://github.com/canonical/microcluster/blob/a4c5273899ac294750d63089f81f78fd1894df67/internal/rest/resources/api_1.0.go#L21-L25) after importing MicroCluster `v2`
- [x] Update the docs (e.g. preseed)


cmd/microcloud/session.go
```diff
@@ -0,0 +1,149 @@
...
+		return fmt.Errorf("Failed to send session start: %w", err)
+	}
+
+	err = gw.ReceiveCombined(&session)
+	if err != nil {
+		return fmt.Errorf("Failed to read session reply: %w", err)
+	}
+
+	if !c.autoSetup {
+		fmt.Printf("Enter the passphrase on systems that you want to join the cluster:\n\n %s\n\n", session.Passphrase)
```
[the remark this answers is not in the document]
roosterfish: This is probably one of the only times we can accept the warning as we WANT the passphrase to be displayed.

roosterfish:
Added some changes as `microcloud service add` was broken due to the wrong certificate being used.
Functions like `RemoteClusterMembers` are used both during pre-init and after bootstrap/join but the latter has to use the cluster certificate. 

I also made a mistake whilst rebasing https://github.com/canonical/microcloud/pull/358. As the remote token creation/deletion always happens after the MicroCloud's microcluster is formed, we neither need a secret nor a custom certificate as we can already use full mTLS.

cmd/microcloud/session.go
```diff
@@ -0,0 +1,166 @@
...
+		if err != nil {
+			return err
+		}
+
+		fingerprint, err := c.shortFingerprint(cert.Fingerprint())
+		if err != nil {
+			return fmt.Errorf("Failed to shorten fingerprint: %w", err)
+		}
+
+		fmt.Printf(" Using fingerprint %q to confirm the join intents\n", fingerprint)
```
masnax, line 77: How about changing this to `Please verify this fingerprint is displayed on all joining systems`. "join intent" should really be an internal concept.

cmd/microcloud/session.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	// The server confirms the target address if none was provided by the client.
+	err = gw.ReceiveCombined(&session)
+	if err != nil {
+		return fmt.Errorf("Failed to find an eligible system: %w", err)
+	}
+
+	if !c.autoSetup {
+		fmt.Printf("Waiting for system %q to confirm our intent to join ...\n", session.TargetAddress)
```
masnax, line 162: Is there a way to get the name of the initiator at this stage? On some lines we refer to the system by name, and on this line we do so by address. It would be nice to keep it consistent.

Also, how about changing this so that the next step is clear to the user:
```
Complete cluster member selection on "m1" to continue ...
```

cmd/microcloud/session.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	err := gw.Write(session)
+	if err != nil {
+		return fmt.Errorf("Failed to send session start: %w", err)
+	}
+
+	if !c.autoSetup && target == "" {
+		fmt.Println("Searching an eligible system ...")
+	}
```
masnax, line 153: Is this line necessary?

cmd/microcloud/ask.go
```diff
@@ -1333,3 +1337,243 @@ func (c *initConfig) askClustered(s *service.Handler, expectedServices []types.S
...
+		}
+
+		fingerprint, err := c.shortFingerprint(shared.CertFingerprint(remoteCert))
+		if err != nil {
+			return fmt.Errorf("Failed to shorten fingerprint: %w", err)
+		}
+
+		fmt.Printf("\n Received confirmation from the system %q at %q using fingerprint %q\n\n", session.Intent.Name, session.Intent.Address, fingerprint)
+		fmt.Println("Do not exit out to keep the session alive")
+		fmt.Printf("Waiting for system %q to let us join the cluster ...\n", session.Intent.Name)
```
masnax, line 1565: How about changing this so that the user knows to go back to the initiator to finish configuration:
```
Do not exit out to keep the session alive
Complete remaining configuration on "m1" to end the session
```

cmd/microcloud/ask.go
```diff
@@ -1333,3 +1337,243 @@ func (c *initConfig) askClustered(s *service.Handler, expectedServices []types.S
...
+		if err != nil {
+			return fmt.Errorf("Failed to parse certificate: %w", err)
+		}
+
+		fingerprint, err := c.shortFingerprint(shared.CertFingerprint(remoteCert))
+		if err != nil {
+			return fmt.Errorf("Failed to shorten fingerprint: %w", err)
+		}
+
+		fmt.Printf("\n Received confirmation from the system %q at %q using fingerprint %q\n\n", session.Intent.Name, session.Intent.Address, fingerprint)
```
masnax, line 1563: At the point that we display this fingerprint, the mTLS has already been established, so verifying that it matches the initiator's fingerprint is a bit moot, no?

Can we move this to right after the joiner supplies the password?

cmd/microcloud/session.go
masnax: Also, can this be moved up where we first display the passphrase? It would be useful to know the fingerprint of the initiator beforehand so that it can be verified on the joiner right after entering the passphrase.

service/lxd.go
```diff
@@ -73,14 +75,32 @@ func (s LXDService) remoteClient(secret string, address string, port int64) (lxd
...
+		cert, err = clusterCert.PublicKeyX509()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	publicKeyBlock := pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: cert.Raw,
+	}
```
masnax, line 94: you can use the microcluster `types.X509Certificate{Certificate: cert}.String()` so that you don't have to hard-code this.

cmd/microcloud/preseed.go
```diff
@@ -100,8 +107,40 @@ func DiskOperatorSet() filter.OperatorSet {
...
+}
+
+func (c *cmdPreseed) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "preseed",
+		Short: "Initialize and extend a MicroCloud cluster unattended",
+		RunE:  c.Run,
+	}
+
+	cmd.Flags().BoolVar(&c.flagAdd, "add", false, "Add new systems to an existing MicroCloud cluster")
```
masnax, line 123: Can you not infer `add` without the flag if the following is true?
* `initiator` is set
* The `initiator` system is not in the `systems` list
* The local system is bootstrapped, and the `initiator` name matches its cluster name.

cmd/microcloud/preseed.go
```diff
@@ -37,6 +42,8 @@ type Preseed struct {
 // System represents the structure of the systems we expect to find in the preseed yaml.
 type System struct {
 	Name            string      `yaml:"name"`
+	ListenAddress   string      `yaml:"listen_address"`
```
masnax, line 45: This isn't strictly the "listen_address" right? it's also the actual cluster address that will be used. Maybe just call it `address`?

cmd/microcloud/preseed.go
```diff
@@ -232,11 +305,20 @@ func (p *Preseed) validate(name string, bootstrap bool) error {
...
+
+	systemNames := make([]string, 0, len(p.Systems))
 	for _, system := range p.Systems {
 		if system.Name == "" {
 			return fmt.Errorf("Missing system name")
 		}
 
+		if system.ListenAddress == "" {
+			return fmt.Errorf("Missing listen address for system %q", system.Name)
+		}
```
masnax, line 320: This means ListenAddress must be specified for all systems, which defeats the purpose of mDNS, right?

cmd/microcloud/preseed.go
```diff
@@ -19,14 +22,16 @@ import (
...
 	"github.com/canonical/microcloud/microcloud/api"
 	"github.com/canonical/microcloud/microcloud/api/types"
+	cloudClient "github.com/canonical/microcloud/microcloud/client"
 	"github.com/canonical/microcloud/microcloud/mdns"
 	"github.com/canonical/microcloud/microcloud/service"
 )
 
 // Preseed represents the structure of the supported preseed yaml.
 type Preseed struct {
-	LookupSubnet          string        `yaml:"lookup_subnet"`
```
masnax, line 28: I thought we still needed a LookupSubnet for the multicast case where we don't know the joiners' addresses precisely. 

Can't we set a subnet and pick the address of the first interface that fits in the subnet as the initiator's listen address, and then use mDNS to pick up the remaining systems, filtering by this subnet?

multicast should really be a best effort so if it's not precise, that's fine I think. If you need precision you'd use unicast.

masnax (changes requested):
This is awesome, and intuitive to follow, nice job @roosterfish! :)

I mostly have nits to report, primarily structuring of the messages that are printed to the user.

Couple things that stand out:
* During a snap mismatch, the error seems incorrect now because the multicast lookup behaviour is different: `Error: Failed to read join confirmation: Failed to send join intent: Skipping peer "micro01" due to missing services (MicroOVN)`
* During preseed, it seems we require `listen_address` to be set, but that defeats the purpose of multicast lookup
* the `--add` flag for preseed ~~seems redundant~~ can be impoved

* The biggest one (but also not a blocker) is the behaviour if a joiner's session ends. On the initiator there's no feedback, so as you are flipping through your joiners and typing the password, if one of them dies in the background, by the time you get back to the initiator and confirm the systems in the table, you have no way of knowing that a problem occurred, and the initiation fails. It would be nice if we could have a column for "session status" in the table so that we know when a joiner has closed its session, and can work around it. I don't believe our current table package allows for easy updates though, so we can deal with this later.

cmd/microcloud/preseed.go
masnax: Ah, I forgot about the unicast case when adding nodes which requires the joiners to know about the initiator's address. 

I guess you're currently handling this by adding an entry to the `systems` list including an initiator node that is already clustered.

I'm worried that might be a bit confusing because there are other keys that are valid for a `systems` entry. Either we ignore these keys (which would be confusing to the user), or we would have to add special-case validation for this entry so that the user can't specify an uplink network or storage configuration for the initiator when adding nodes.

Instead, I think it might be simpler to just add another key:
`initiator_address` that can be used instead of (or in addition to) `initiator`. If both are specified, then we must verify the address belongs to the node with that name.

Then we don't also need a `--add` flag, since the cases would look like this:
### init
* `initiator: m1`
* `initiator_address: 10.0.0.10` (no effect except validation with `initiator`)
* `systems: m1, m2, m3` (with addresses if unicast)

### add
* `initiator: m1`
* `initiator_address: 10.0.0.10` (required for unicast, no effect except validation with `initiator` for multicast)
* `systems: m2, m3` (with addresses if unicast)

api/session.go
```diff
@@ -0,0 +1,341 @@
...
+			defer func() {
+				err := conn.Close()
+				if err != nil {
+					logger.Error("Failed to close the websocket connection", logger.Ctx{"err": err})
+				}
+			}()
+
+			sessionCtx, cancel := context.WithTimeoutCause(r.Context(), sessionTimeout, fmt.Errorf("Session timeout exceeded"))
+			defer cancel()
+
```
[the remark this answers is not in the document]
roosterfish: As it's just a query parameter I agree it would be nice to have a default. But if `timeout` has a default, should `role` also have one?

Maybe to overcome this, we can have a `/session/joining` and `/session/initiating` which optionally accepts a `?timeount` and defaults to `10m`. 
Then the `sessionGet` handler can accept both the `service.Handler` and the session type (joining, initiating) as argument.

client/websocket.go
```diff
@@ -0,0 +1,147 @@
...
+// If there aren't other channels that need to be consumed in parallel use ReceiveCombined instead.
+func (w *WebsocketGateway) Receive() <-chan []byte {
+	return w.reader
+}
+
+// ReceiveCombined tries to read from the websocket connection and unmarshals
+// the received data into v.
+// It's waiting on both the websocket connection and the context and returns
+// whatever is returning/cancelled first.
+func (w *WebsocketGateway) ReceiveCombined(v any) error {
```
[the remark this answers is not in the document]
roosterfish: Happy to change the name but I don't think another context should be passed here.

The overall problem comes from the fact that the underlying websocket `*websocket.Conn` doesn't know the concept of a context (remember [this](https://github.com/canonical/microcluster/pull/237#discussion_r1722949800) comment).
That was the main reason for adding the `WebsocketGateway` to be able to couple an active websocket connection with an outer context.

Therefore when you create the `WebsocketGateway` you pass the outer context which will be cancelled in case we observe any errors reading from the websocket connection. This is the intended approach according to [the docs](https://pkg.go.dev/github.com/gorilla/websocket#Conn.NextReader):

> Applications must break out of the application's read loop when this method returns a non-nil error value. Errors returned from this method are permanent. Once this method returns a non-nil error, all subsequent calls to this method return the same error.

To be able to react on errors from the underlying websocket quickly, a go routine is spawned that continuously reads from it and cancels the context in case of error.

So you can reuse the context which was already passed when creating the `WebsocketGateway`.
In case you want to timeout waiting for a message (or error of websocket connection), you can use this pattern:

```go
gw := NewWebsocketGateway(outerCtx, conn)

select {
    case <-gw.Context().Done():
       // outer context cancelled or error in websocket conn
    case <-gw.Receive():
       // actual message received
    case <-time.After(duration):
       // timeout waiting for message
}
``` 

But one use case I would see is wrapping the timeout concept in `ReceiveWithContext(ctx, v)` so you can do something like "I want to wait 10s for a message".
So within the func we will wait on both the `gw.Context().Done()` and the passed `ctx` as well as the `w.Receive()` chan.

cmd/microcloud/ask.go
```diff
@@ -1325,3 +1329,238 @@ func (c *initConfig) askClustered(s *service.Handler, expectedServices []types.S
...
+	err = gw.ReceiveCombined(&session)
+	if err != nil {
+		return fmt.Errorf("Failed waiting during join: %w", err)
+	}
+
+	if session.Error != "" {
+		return fmt.Errorf("Failed to join system: %s", session.Error)
+	}
+
+	fmt.Println("Successfully joined")
```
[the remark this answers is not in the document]
roosterfish: I have picked `cluster` as the other might imply we have only joined the MicroCloud's microcluster.

service/session.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/canonical/microcloud/microcloud/api/types"
+	cloudMDNS "github.com/canonical/microcloud/microcloud/mdns"
+)
+
+// Session represents a local trust establishment session.
+type Session struct {
+	lock       sync.RWMutex
+	passphrase string
+	server     *mdns.Server
+	trustStore map[string]x509.Certificate
```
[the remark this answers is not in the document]
roosterfish: Good idea. I have also added a new `StopSession` func to the handler which resets the underlying `session` field to `nil` after stopping the session.

cmd/microcloud/session.go
```diff
@@ -0,0 +1,166 @@
...
+
+	err = gw.ReceiveCombined(&session)
+	if err != nil {
+		return fmt.Errorf("Failed to read session reply: %w", err)
+	}
+
+	if !c.autoSetup {
+		fmt.Printf("Enter the passphrase on systems that you want to join the cluster:\n\n %s\n\n", session.Passphrase)
+		fmt.Println("Waiting for systems to send their intent to join ...")
```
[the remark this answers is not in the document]
roosterfish: I have changed the dialog to this:

```
Select an address for MicroCloud's internal traffic:

 Using address "10.165.133.71" for MicroCloud

Use the following command on systems that you want to join the cluster:

 microcloud join

When requested enter the passphrase:

 juice imaging exterior nevermore

Waiting for systems to send their intent to join ...
```

cmd/microcloud/session.go
roosterfish: In case you rely on mDNS and haven't specified `--target` we have to wait for any mDNS server to respond to our query. If this is taking a bit longer or if there is no `microcloud init` running on another node you will get something like this:

```
microcloud join --lookup-timeout 5
...
Searching an eligible system ...
Error: Failed to find an eligible system: Failed lookup: context deadline exceeded
```

Or the session timeout kicks in whatever happens first.

cmd/microcloud/ask.go
roosterfish: Will use `Complete the remaining configuration on "m1" ...` as we don't need to mention the session twice.

cmd/microcloud/ask.go
roosterfish: We only have the other sides fingerprint in case of mDNS.
When doing unicast the information is available as soon as the initiator talks back to us by confirming the intent to join.

The other option is we retrieve it from the underlying TLS conn when first sending the join intent.

service/lxd.go
roosterfish: Thanks, wasn't aware of this utility in microcluster :)

cmd/microcloud/preseed.go
roosterfish: That is right.
As you anyway have to specify the `address` (formerly `listen_address`) for each system, there is no point in still using mDNS as we can reach out to the nodes directly based on their addresses.
The same happens in the interactive setup where MicroCloud asks you to set the address.

This is the replacement for the two top level keys `lookup_iface` and `lookup_subnet`.

cmd/microcloud/preseed.go
roosterfish: Please see the comment here https://github.com/canonical/microcloud/pull/383#discussion_r1736694656.

cmd/microcloud/preseed.go
roosterfish: I like this idea.

If I don't forgot any scenario, we would not even need `initiator`. By having only `initiator_address`, we can check both scenarios too:

* **init**: Don't set `initiator_address`, pick the first one in the list of systems. If it's set, there has to be a system in the list matching this address.
* **add**: Imply we are in this mode if `initiator_address` is set and if it's not in the list of already specified system addresses.

cmd/microcloud/preseed.go
masnax: If the choice is between having 1 top-level `lookup_subnet` key that enables multicast over a subnet without directly specifying addresses, and outright disabling multicast for preseed, is there a reason you decided to go with the latter rather than the former?

cmd/microcloud/session.go
masnax: Ah I see so in the unicast case this is just the user-specified target address?

roosterfish: Yep, the server checks if the target address is set, if not use mDNS to lookup a target, but in any case return the final target back to the client.

cmd/microcloud/preseed.go
roosterfish: Actually you were right with having both the `initiator` and `initiator_address`. If there is just `initiator_address` you cannot grow the cluster using multicast as you would need to specify the address of an existing cluster member which again defeats the purpose of multicast for preseed.

By having `initiator` and `initiator_address` we can also get rid of the concept of picking the "first" system as we can enforce you specify either `initiator` (when doing multicast) or `initiator_address` (when doing unicast).

client/websocket.go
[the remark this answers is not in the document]
roosterfish: I have now implemented `ReceiveWithContext` which allows having an additional context for timeout scenarios.

But for this PR it's not used as we solely rely on the outer context from the gateway itself.

cmd/microcloud/preseed.go
roosterfish: Okay, as discussed the `lookup_subnet` was added back in together with the `addrInSubnet` func that picks the first configured addressed out of the given subnet.

service/lxd.go
roosterfish: Fixed.

cmd/microcloud/preseed.go
roosterfish: Fixed.

cmd/microcloud/preseed.go
roosterfish: Fixed.

cmd/microcloud/join.go
```diff
@@ -0,0 +1,106 @@
...
+func (c *cmdJoin) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "join",
+		Short: "Join an existing MicroCloud cluster",
+		RunE:  c.Run,
+	}
+
+	cmd.Flags().BoolVar(&c.flagAutoSetup, "auto", false, "Automatic setup with default configuration")
+	cmd.Flags().BoolVar(&c.flagWipe, "wipe", false, "Wipe disks to add to MicroCeph")
+	cmd.Flags().Int64Var(&c.flagLookupTimeout, "lookup-timeout", 0, "Amount of seconds to wait for systems to show up. Defaults: 60s")
```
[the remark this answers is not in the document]
roosterfish: Fixed.

service/session.go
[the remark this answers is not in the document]
roosterfish: Fixed.

cmd/microcloud/session.go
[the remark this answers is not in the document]
roosterfish: Fixed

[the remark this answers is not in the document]
roosterfish: But I am still wondering if we should maybe put the used fingerprint in another line.

cmd/microcloud/ask.go

client/websocket.go
```diff
@@ -0,0 +1,147 @@
...
+// Write writes the given data onto the websocket connection.
+func (w *WebsocketGateway) Write(v any) error {
+	w.writeLock.Lock()
+	defer w.writeLock.Unlock()
+
+	return w.conn.WriteJSON(v)
+}
+
+// WriteClose sends a websocket control close message
+// I waits for the other side to hang up or the gateway's context being cancelled.
```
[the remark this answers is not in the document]
roosterfish: Fixed.

api/session_join.go
```diff
@@ -0,0 +1,78 @@
...
+		select {
+		case sh.Session.IntentCh() <- req:
+			return response.EmptySyncResponse
+		default:
+			return response.InternalError(fmt.Errorf("No active consumer for join intent"))
+		}
+	}
+}
+
+func validateIntent(sh *service.Handler, intent types.SessionJoinPost) error {
```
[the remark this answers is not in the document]
roosterfish: Fixed.

api/session.go
```diff
@@ -0,0 +1,341 @@
...
+// sessionGet returns a MicroCloud join session.
+func sessionGet(sh *service.Handler) func(state *state.State, r *http.Request) response.Response {
+	return func(state *state.State, r *http.Request) response.Response {
+		if sh.ActiveSession() {
+			return response.BadRequest(fmt.Errorf("There already is an active session"))
+		}
+
+		sessionRole := r.URL.Query().Get("role")
+		if !shared.ValueInSlice(sessionRole, []string{"initiating", "joining"}) {
```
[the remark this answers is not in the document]
roosterfish: Fixed.

cmd/microcloud/session.go
roosterfish: Reworked as discussed.

First we lookup the address in case of multicast or continue if the unicast address was already provided.
As the mDNS payload was reduced down to only a minimal set of information we query the status endpoint in both the multicast and unicast scenario to retrieve the required information equally in both cases.

We now query the status endpoint on the target to retrieve the remote's name.
The fingerprint is resolved by using the `shared.GetRemoteCertificate` func.

cmd/microcloud/session.go
roosterfish: Fixed. The used fingerprint is now displayed right after the chosen address for MicroCloud.

client/websocket.go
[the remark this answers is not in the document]
masnax: So I think part of my confusion here is that we had `ReceiveCombined` and `Receive`, but the context is internal, so why do we have the distinction between the two functions?`ReceiveWithContext` is also a victim of this distinction because it's now using the same context for two different `select` cases in all its usages.

So I'm wondering why we even need to call `Receive` apart from `ReceiveWithContext(gw.Context()...` from anywhere outside the `client` package? 

Could we just move the internal context handling directly into `Receive`? Then we can do away with `ReceiveWithContext`, and if we ever need to include it, we can add it back and just have it `select` over the context and `gw.Receive`?

api/session_join.go
```diff
@@ -0,0 +1,80 @@
...
+
+// validateIntent validates the given join intent.
+// It checks whether or not the peer is missing any of our services and returns an error if one is missing.
+func validateIntent(sh *service.Handler, intent types.SessionJoinPost) error {
+	// Skip any peers that are missing our services.
+	for service := range sh.Services {
+		if !shared.ValueInSlice(service, intent.Services) {
+			return fmt.Errorf("Skipping peer %q due to missing services (%s)", intent.Name, string(service))
+		}
+	}
```
masnax, line 77: This error message isn't quite right, I don't believe we "skip" peers anymore, we just error out.

cmd/microcloud/session.go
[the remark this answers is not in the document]
masnax: Yeah I think maybe this can be
```
Use the following command on systems that you want to join the cluster:
 
  microcloud join

When requested enter the passphrase:

  urgent crumble emperor governor

Verify the fingerprint is displayed on joining systems: f211051397df
Waiting to detect systems (because they won't be joining quite yet :)) ...
```

cmd/microcloud/session.go
masnax: Should be "Searching _for_ an eligible system ..." right?

Also maybe `s/system/initiator/`?

api/session.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	session.TargetFingerprint = shared.CertFingerprint(peerCert)
+
+	// Get the remotes name.
+	cloud := sh.Services[types.MicroCloud].(*service.CloudService)
+	peerStatus, err := cloud.RemoteStatus(gw.Context(), peerCert, session.TargetAddress)
+	if err != nil {
+		return fmt.Errorf("Failed to retrieve cluster status: %w", err)
+	}
```
masnax, line 279: Would it be worth it to make this a bit more efficient by returning all the information we need from `RequestJoinIntent`? It would reduce the network back/forth by 1, but it would also push the client notification back a bit.

cmd/microcloud/preseed.go
```diff
@@ -159,11 +230,30 @@ func (c *initConfig) RunPreseed(cmd *cobra.Command) error {
 		return err
 	}
 
+	initiator := config.isInitiator(c.name, c.address)
+
+	fmt.Println("initiator", initiator)
+	fmt.Println("bootstrap", c.bootstrap)
```
masnax, line 236: Leftover debugging? :)

cmd/microcloud/preseed.go
```diff
@@ -159,11 +230,30 @@ func (c *initConfig) RunPreseed(cmd *cobra.Command) error {
...
+
+	if !ready && !c.bootstrap && initiator {
+		return fmt.Errorf("MicroCloud isn't yet initialized, run 'microcloud preseed' instead")
```
masnax, line 239: This error's a bit odd, since there's only one command now. Maybe just mention that `c.name` or `c.address` cannot be the initiator?

cmd/microcloud/preseed.go
```diff
@@ -364,6 +460,77 @@ func (p *Preseed) validate(name string, bootstrap bool) error {
...
+
+// initiatorAddress returns the address of the trust establishment session's initiator.
+// This is either the specified address of the initiator or an empty string.
+func (p *Preseed) initiatorAddress() string {
+	if p.InitiatorAddress != "" {
+		return p.InitiatorAddress
+	}
+
+	return ""
+}
```
masnax, line 480: So looks like we don't infer the first system in the list is the initiator anymore for bootstrap? It's a required key?

I think we should error out if there are ambiguities like `name != "" && name != p.Initiator && address == p.InitiatorAddress`

Also that `initiatorAddress` helper doesn't seem to be doing much.

masnax (changes requested):
Just a handful of more nits, and some validation suggestions :)

client/websocket.go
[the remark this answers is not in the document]
roosterfish: The reason for `Receive` as a standalone is to get a channel from the websocket that allows you to either wait on an ingress message from the websocket or on a different channel using a `select` statement.

This is required in the initiating session.
We can at the same time receive:
 * a new join intent from any joiner which has to be relayed to the client to be displayed in the table
 * the message from the client confirming the join intents

See the section of the code [here](https://github.com/canonical/microcloud/pull/383/files#diff-27dcd3e94dd8f08cf82109cf63ca764783c103a170caa9c30738b1ba87f5c6f8R101).

Now all `ReceiveWithContext` usages use the same context, but I found it might be useful in the future to supply an additional context with a timeout if you want to wait for a message to arrive in X seconds.

cmd/microcloud/session.go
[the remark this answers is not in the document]
roosterfish: Mh yeah this looks better!

I wen with `Verify the fingerprint "30a09aeae52e" is displayed on joining systems.` as the remainder of the dialog also tends to provide the quoted info inline and not after a colon.

cmd/microcloud/session.go
roosterfish: Good point, updated the `Searching ` line.

In case of interactive mode should we introduce terms like initiator/joiner? I think it makes it easier to follow if we just assume those things under the hood.

Whereas in case of preseed I think it's totally fine to introduce these terms as you need to understand when to set either `initiator` or `initiator_address` in the preseed file.

api/session_join.go
roosterfish: Right, thanks! Replaced `Skip` with `Reject`.

cmd/microcloud/preseed.go
roosterfish: Oh yeah, removed :)

cmd/microcloud/preseed.go
roosterfish: Yeah I think I have removed the `--add` flag but left the message.

cmd/microcloud/preseed.go
roosterfish: Yeah this way in all scenarios (interactive, preseed) and modes (unicast, multicast) you have to either provide `initiator` or `initiator_address`. So it looks a bit more consistent.

The preseed's `validate()` already has a check in case either `initiator` or `initiator_address` are empty. 
I have added an additional check that doesn't allow setting both `initiator` and `initiator_address` at the same time.
Therefore things like `name != "" && name != p.Initiator && address == p.InitiatorAddress` shouldn't be anymore possible.

api/session.go
roosterfish: You mean the extra call for getting the cert? 

I was deciding for this in favor of drilling down the microcluster query methods in order to grep something from the HTTP's response `TLS` field.

Do you think it would be worth exposing this in another microcluster query func or should we use the `http` request tooling instead of the microcluster query func inside `RequestJoinIntent`?

masnax: No I meant the additional query to `/1.0` to fetch the cluster name. I wonder if `RequestJoinIntent` should return a response body that contains it so we don't need the additional request over the network.

api/session.go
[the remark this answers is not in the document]
roosterfish: Implemented as discussed.

api/session.go
roosterfish: Fixed by getting the fingerprint from the actual TLS connection when `RequestJoinIntent` is sent.
The name is fetched from the already existing microcluster status endpoint.

cmd/microcloud/ask.go
```diff
@@ -1420,3 +1423,227 @@ func (c *initConfig) askClustered(s *service.Handler, expectedServices []types.S
...
+	err = gw.ReceiveWithContext(gw.Context(), &session)
+	if err != nil {
+		return fmt.Errorf("Failed waiting during join: %w", err)
+	}
+
+	if session.Error != "" {
+		return fmt.Errorf("Failed to join system: %s", session.Error)
+	}
+
+	fmt.Println("Successfully joined the cluster")
```
masnax, line 1647: This prints at the wrong time, or it needs to change its message.

Currently it prints around when the initiator is bootstrapped, but well before the joiner has joined the cluster.

masnax: It's still a rather arbitrary distinction. I think the session should either wait until the node has joined all cluster members (since there would be dangling join tokens), or it should end with `Commencing cluster join. Session closed`. or something.

masnax: I am not following on the dangling tokens though. In which situation would you see this happening?

I meant in case of error. We do try to clean up, though, so it's not a big issue.

What I'm most concerned with is somebody seeing "oh the joiner says it's done" and then try to interact with MicroCloud while the session is still active. This might be a particular problem for preseed because there might be other code that begins to execute as soon as the joiner stops blocking. 

That's why I think the command should be blocking until the initiator reports back to the joiners that the cluster formation is actually fully complete.

roosterfish: The preseed blocking is a good point. Even though you can potentially run the operation also on the initiator to work around this (in case it's not specific to this member).

This feels closely related with the topic we have discussed about letting the joiner know in case the initiator observed an error during forming of the cluster.
I'll log this as a separate issue so we can address all of this together in one shot either using a new endpoint `session/exit` or by having a websocket between the systems.

For know I have changed the exit message to the following:

```
Successfully joined the MicroCloud cluster and closing the session.
Commencing cluster join of the remaining services (LXD, MicroCeph, MicroOVN)
```

It's using the local systems list of services and filters out `MicroCloud` as this cluster is already formed.

masnax:
@roosterfish Have you tried this with IPv6? 

Setting `ipv4.address=none` and using IPv6 for the management network seems to make the joiners get `connection refused` when they enter the passphrase:

```
Error: Failed to find an eligible system: Failed to send our intent to join "fd42:b693:902c:78f:216:3eff:fe50:332c": Failed to send join intent: Post "https://[fd42:b693:902c:78f:216:3eff:fe50:332c]:9443/1.0/session/join": Unable to connect to "[fd42:b693:902c:78f:216:3eff:fe50:332c]:9443": dial tcp [fd42:b693:902c:78f:216:3eff:fe50:332c]:9443: connect: connection refused
```

Fedqys (approved):
The security team has reviewed and validated this Pull Request. Thanks for your hard work on this!

masnax:
@roosterfish Got one other regression for you:

The previous behaviour in the following case allowed a MicroCloud to form, but now it fails:
* initiator: MicroCloud, LXD, MicroOVN
* joiner: MicroCloud, LXD, MicroOVN, MicroCeph

When running `microcloud init`, the initiator is prompted with a warning that they don't have MicroCeph installed, and if they want to continue, MicroCeph will not be configured. When looking up systems, the initiator ignores MicroCeph, so if any joiners have installed the snap already, they can still join the overall MicroCloud, but their MicroCeph will be ignored.

In the end, you would end up with a MicroCloud, LXD, and MicroOVN cluster. MicroCeph would be skipped entirely.

However, with this PR, the above configuration will fail because the joiner must also have MicroCeph uninstalled. I don't think we need this restriction since it will be ignored by the initiator anyway, and they have already consented to proceed without MicroCeph.

masnax:
> @roosterfish Have you tried this with IPv6?
> 
> Setting `ipv4.address=none` and using IPv6 for the management network seems to make the joiners get `connection refused` when they enter the passphrase:
> 
> ```
> Error: Failed to find an eligible system: Failed to send our intent to join "fd42:b693:902c:78f:216:3eff:fe50:332c": Failed to send join intent: Post "https://[fd42:b693:902c:78f:216:3eff:fe50:332c]:9443/1.0/session/join": Unable to connect to "[fd42:b693:902c:78f:216:3eff:fe50:332c]:9443": dial tcp [fd42:b693:902c:78f:216:3eff:fe50:332c]:9443: connect: connection refused
> ```

Nevermind about this one, it was just the `PreInitListenAddress` being assigned `0.0.0.0` as its address. After some testing `[::]` seems to work even with IPv4 disabled, so I've switched to that.

roosterfish:
> @roosterfish Got one other regression for you:

Thanks! I think we have talked about this briefly two weeks ago shortly before I left. 
I have already put in a fix for this together with not prompting for missing services on the `microcloud join` side.

roosterfish:
@masnax preseed test succeeded after resolving the conflict, so I would carefully say it's ready for review again :)

masnax (approved):
Other than that 1 nit, this looks great! Thanks for the hard work @roosterfish!

PR #384, opened 2024-08-28
Status

masnax:
Adds a status command usable with `microcloud status`.

A minimal example output is shown below:

```
 Status: ERROR

 ┃ ⨯ Reliability risk: 3 systems are required for effective fault tolerance
 ┃ ! MicroOVN is not installed on micro02, micro01
 ┃ ! MicroCeph is not installed on micro02, micro01

┌─────────┬────────────────┬───────────┬────────────────────┬───────────────────┬────────┐
│ Name    │ Address        │ OSD Disks │ MicroCeph Services │ MicroOVN Services │ Status │
├─────────┼────────────────┼───────────┼────────────────────┼───────────────────┼────────┤
│ micro02 │ 10.148.181.110 │ -         │ -                  │ -                 │ ONLINE │
│ micro01 │ 10.148.181.237 │ -         │ -                  │ -                 │ ONLINE │
└─────────┴────────────────┴───────────┴────────────────────┴───────────────────┴────────┘
```

The output is split into 3 parts:
* Status: ERROR 
  * Reports the overall summary of the cluster state. This depends on the list of warnings just below it.
* Warnings
  * These are a series of messages that appear with 2 levels of severity (warn/error) that dictate the `Status` value. These will report any issues with the cluster, like missing snaps, mismatched cluster members, schema upgrades, and incorrectly configured OSDs.
* Table
  * The table reports the current state of the cluster. OSDs and ceph/ovn services will only be populated if those services are installed. The overall status depends on how other cluster members reach out to the local system. The status will only report ONLINE if all snaps report ONLINE for that cluster member.

### DIFF
--- a/api/status.go
+++ b/api/status.go
@@ -1,0 +1,288 @@
+package api
+
+import (
+	"context"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/lxd/lxd/util"
+	lxdAPI "github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/logger"
+	cephTypes "github.com/canonical/microceph/microceph/api/types"
+	cephClient "github.com/canonical/microceph/microceph/client"
+	microClient "github.com/canonical/microcluster/v2/client"
+	"github.com/canonical/microcluster/v2/rest"
+	microTypes "github.com/canonical/microcluster/v2/rest/types"
+	"github.com/canonical/microcluster/v2/state"
+	ovnTypes "github.com/canonical/microovn/microovn/api/types"
+	ovnClient "github.com/canonical/microovn/microovn/client"
+
+	"github.com/canonical/microcloud/microcloud/api/types"
+	"github.com/canonical/microcloud/microcloud/client"
+	"github.com/canonical/microcloud/microcloud/service"
+)
+
+// StatusCmd represents the /1.0/status API on MicroCloud.
+var StatusCmd = func(sh *service.Handler) rest.Endpoint {
+	return rest.Endpoint{
+		Name: "status",
+		Path: "status",
+
+		Get: rest.EndpointAction{Handler: statusGet(sh), ProxyTarget: true},
+	}
+}
+
+func statusGet(sh *service.Handler) endpointHandler {
+	// statusMu is used to synchronize map writes to the returned status information, as we populate cluster members for each service concurrently.
+	var statusMu sync.Mutex
+
+	return func(s state.State, r *http.Request) response.Response {
+		statuses := []types.Status{}
+
+		if !microClient.IsNotification(r) {
+			cluster, err := s.Cluster(true)
+			if err != nil {
+				return response.SmartError(err)
+			}
+
+			err = cluster.Query(r.Context(), true, func(ctx context.Context, c *microClient.Client) error {
+				memberStatuses, err := client.GetStatus(ctx, c)
+				if err != nil {
+					logger.Error("Failed to get status for cluster member", logger.Ctx{"error": err, "address": c.URL()})
+
+					return nil
+				}
+
+				statuses = append(statuses, memberStatuses...)
+
+				return nil
+			})
+			if err != nil {
+				return response.SmartError(err)
+			}
+		}
+
+		status := &types.Status{
+			Name:         sh.Name,
+			Address:      sh.Address,
+			Clusters:     make(map[types.ServiceType][]microTypes.ClusterMember, len(sh.Services)),
+			OSDs:         []cephTypes.Disk{},
+			CephServices: []cephTypes.Service{},
+			OVNServices:  []ovnTypes.Service{},
+		}
+
+		err := sh.RunConcurrent("", "", func(s service.Service) error {
+			switch s.Type() {
+			case types.LXD:
+				clusterMembers, err := lxdStatus(r.Context(), s)
+				if err != nil {
+					logger.Error("Failed to get service status", logger.Ctx{"type": s.Type(), "name": sh.Name})
+				}
+
+				statusMu.Lock()
+				status.Clusters[s.Type()] = clusterMembers
+				statusMu.Unlock()
+			case types.MicroCeph:
+				clusterMembers, osds, cephServices, err := cephStatus(r.Context(), s)
+				if err != nil {
+					logger.Error("Failed to get service status", logger.Ctx{"type": s.Type(), "name": sh.Name})
+				}
+
+				status.OSDs = osds
+				status.CephServices = cephServices
+
+				statusMu.Lock()
+				status.Clusters[s.Type()] = clusterMembers
+				statusMu.Unlock()
+			case types.MicroOVN:
+				clusterMembers, ovnServices, err := ovnStatus(r.Context(), s)
+				if err != nil {
+					logger.Error("Failed to get service status", logger.Ctx{"type": s.Type(), "name": sh.Name})
+				}
+
+				status.OVNServices = ovnServices
+
+				statusMu.Lock()
+				status.Clusters[s.Type()] = clusterMembers
+				statusMu.Unlock()
+			case types.MicroCloud:
+				microClient, err := s.(*service.CloudService).Client()
+				if err != nil {
+					return err
+				}
+
+				clusterMembers, err := microStatus(r.Context(), microClient, s)
+				if err != nil {
+					logger.Error("Failed to get service status", logger.Ctx{"type": s.Type(), "name": sh.Name})
+				}
+
+				statusMu.Lock()
+				status.Clusters[s.Type()] = clusterMembers
+				statusMu.Unlock()
+			}
+
+			return nil
+		})
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		statuses = append(statuses, *status)
+
+		return response.SyncResponse(true, statuses)
+	}
+}
+
+func cephStatus(ctx context.Context, s service.Service) (clusterMembers []microTypes.ClusterMember, osds []cephTypes.Disk, cephServices []cephTypes.Service, err error) {
+	microClient, err := s.(*service.CephService).Client("")
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	clusterMembers, err = microStatus(ctx, microClient, s)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	disks, err := cephClient.GetDisks(ctx, microClient)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	for _, disk := range disks {
+		if disk.Location == s.Name() {
+			if osds == nil {
+				osds = []cephTypes.Disk{}
+			}
+
+			osds = append(osds, disk)
+		}
+	}
+
+	services, err := cephClient.GetServices(ctx, microClient)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	for _, service := range services {
+		if service.Location == s.Name() {
+			if cephServices == nil {
+				cephServices = []cephTypes.Service{}
+			}
+
+			cephServices = append(cephServices, service)
+		}
+	}
+
+	return clusterMembers, osds, cephServices, nil
+}
+
+func ovnStatus(ctx context.Context, s service.Service) (clusterMembers []microTypes.ClusterMember, ovnServices []ovnTypes.Service, err error) {
+	microClient, err := s.(*service.OVNService).Client()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	clusterMembers, err = microStatus(ctx, microClient, s)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	services, err := ovnClient.GetServices(ctx, microClient)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	for _, service := range services {
+		if service.Location == s.Name() {
+			if ovnServices == nil {
+				ovnServices = []ovnTypes.Service{}
+			}
+
+			ovnServices = append(ovnServices, service)
+		}
+	}
+
+	return clusterMembers, ovnServices, nil
+}
+
+func microStatus(ctx context.Context, microClient *microClient.Client, s service.Service) ([]microTypes.ClusterMember, error) {
+	clusterMembers, err := microClient.GetClusterMembers(context.Background())
+	if err != nil && !lxdAPI.StatusErrorCheck(err, http.StatusServiceUnavailable) {
+		return nil, err
+	}
+
+	return clusterMembers, nil
+}
+
+func lxdStatus(ctx context.Context, s service.Service) ([]microTypes.ClusterMember, error) {
+	lxdClient, err := s.(*service.LXDService).Client(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	server, _, err := lxdClient.GetServer()
+	if err != nil {
+		return nil, err
+	}
+
+	var microMembers []microTypes.ClusterMember
+	if server.Environment.ServerClustered {
+		clusterMembers, err := lxdClient.GetClusterMembers()
+		if err != nil {
+			return nil, err
+		}
+
+		certs, err := lxdClient.GetCertificates()
+		if err != nil {
+			return nil, err
+		}
+
+		microMembers = make([]microTypes.ClusterMember, 0, len(clusterMembers))
+		for _, member := range clusterMembers {
+			url, err := url.Parse(member.URL)
+			if err != nil {
+				return nil, err
+			}
+
+			addrPort, err := microTypes.ParseAddrPort(util.CanonicalNetworkAddress(url.Host, service.LXDPort))
+			if err != nil {
+				return nil, err
+			}
+
+			// Microcluster requires a certificate to be specified in types.ClusterMemberLocal.
+			var serverCert *microTypes.X509Certificate
+			for _, cert := range certs {
+				if cert.Type == "server" && cert.Name == member.ServerName {
+					serverCert, err = microTypes.ParseX509Certificate(cert.Certificate)
+					if err != nil {
+						return nil, err
+					}
+				}
+			}
+
+			microMember := microTypes.ClusterMember{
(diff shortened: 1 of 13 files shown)


api/status.go
```diff
@@ -0,0 +1,209 @@
...
+					}
+
+					microMembers := make([]microTypes.ClusterMember, 0, len(clusterMembers))
+					for _, member := range clusterMembers {
+						url, err := url.Parse(member.URL)
+						if err != nil {
+							return err
+						}
+
+						addrPort, err := microTypes.ParseAddrPort(util.CanonicalNetworkAddress(url.Host, 8443))
```
roosterfish, line 159: Should we expect that the user has never changed the LXD port or dynamically load the right `core.https_address` from the server config? Maybe it's even contained in the `server` var above.

Aside from this, why does the address have to be populated here? Is it used later to access LXD when parsing the reponse from this endpoint?

api/status.go
```diff
@@ -0,0 +1,209 @@
...
+							}
+						}
+
+						microMember := microTypes.ClusterMember{
+							ClusterMemberLocal: microTypes.ClusterMemberLocal{
+								Name:        member.ServerName,
+								Address:     addrPort,
+								Certificate: *serverCert,
+							},
+							Role:       strings.Join(member.Roles, ","),
```
roosterfish, line 181: Don't we display those roles in the final table as these are the internal Raft ones and we don't have a concept like `mds` or `central` in LXD like for the other Micro* snaps?

api/types/status.go
```diff
@@ -0,0 +1,28 @@
...
+	Address string `json:"address" yaml:"address"`
+
+	// Clusters is a list of cluster members for each service installed on the member.
+	Clusters map[ServiceType][]microTypes.ClusterMember `json:"clusters" yaml:"clusters"`
+
+	// OSDs is a list of all OSDs local to the member.
+	OSDs cephTypes.Disks `json:"osds" yaml:"osds"`
+
+	// CephServices is a list of all ceph services running on this member.
+	CephServices cephTypes.Services `json:"ceph_services" yaml:"ceph_services"`
```
roosterfish, line 24: From what I have seen the code generally talks about services when it comes to the availability of LXD, MicroCeph and MicroOVN. 

Should we pick another name or simply say `CephRoles`? This would maybe also make it easier to have another column in the table that can also show LXD's roles.

cmd/style/style.go
```diff
@@ -0,0 +1,113 @@
...
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/charmbracelet/lipgloss"
+)
+
+const (
+	// White represents the common white color used throughout the CLI.
+	White = "252"
```
roosterfish, line 13: I assume this is RGB. Why not `255` which would be a clean bright white? Is it to highlight the color?

cmd/microcloud/status.go
```diff
@@ -0,0 +1,444 @@
...
+		return err
+	}
+
+	cloudClient, err := sh.Services[types.MicroCloud].(*service.CloudService).Client()
+	if err != nil {
+		return err
+	}
+
+	// Query the status API for the cluster.
+	statuses, err := client.GetStatus(context.Background(), cloudClient)
```
roosterfish, line 159: I have deployed a MicroCloud with three members but shutdown one of them. 
Now when I run `microcloud cluster list` I see one of them being `UNREACHABLE` whilst `microcloud status` gives me `Error: Get "https://10.134.197.4:9443/1.0/status": Unable to connect to "10.134.197.4:9443": dial tcp 10.134.197.4:9443: connect: no route to host`. 

I would have expected to still see the status table but with one member giving the status `OFFLINE`.
It's my assumption that the error is raised at this point :)

roosterfish: After starting up the member I shortly saw this:

```
root@m1:/var/snap/microcloud/common/state# microcloud status

 Status: WARNING

 ┃ ! m3 is not available

┌──────┬────────────────┬───────────┬────────────────────┬────────────────────────┬─────────┐
│ Name │ Address        │ OSD Disks │ MicroCeph Services │ MicroOVN Services      │ Status  │
├──────┼────────────────┼───────────┼────────────────────┼────────────────────────┼─────────┤
│ m2   │ 10.134.197.5   │ 1         │ mds,mgr,mon        │ central,chassis,switch │ ONLINE  │
│ m3   │ 10.134.197.4   │ 1         │ mds,mgr,mon        │ central,chassis,switch │ Offline │
│ m1   │ 10.134.197.164 │ 1         │ mds,mgr,mon        │ central,chassis,switch │ ONLINE  │
└──────┴────────────────┴───────────┴────────────────────┴────────────────────────┴─────────┘
```

So I guess there is a check for this but it looks to be failing currently if the entire node is offline.

cmd/style/style.go
```diff
@@ -0,0 +1,113 @@
...
+	Red = "125"
+
+	// Green represents the common green color used throughout the CLI.
+	Green = "82"
+
+	// Yellow represents the common yellow color used throughout the CLI.
+	Yellow = "214"
+
+	// Purple represnts the common purple color used throughout the CLI.
+	Purple = "139"
```
roosterfish, line 31: If you use the std Ubuntu terminal like I do, then the purple doesn't look that nicely in contrast to the background color:

Screenshot from 2024-09-05 14-32-35

cmd/microcloud/status.go
```diff
@@ -0,0 +1,444 @@
...
+		return err
+	}
+
+	status, err := cloudApp.Status(context.Background())
+	if err != nil {
+		return fmt.Errorf("Failed to get MicroCloud status: %w", err)
+	}
+
+	if !status.Ready {
+		return fmt.Errorf("MicroCloud is uninitialized, run 'microcloud init' first")
```
roosterfish, line 121: In the [first commit](https://github.com/canonical/microcloud/pull/384/commits/71ffd29b8cbdbbca1f67d889e8f4477b13f3485b#diff-1f9d01a7137adb7eda731cd49f4e87908c5f90f56beac88c93792ea7cf504ad3R30) I saw the status endpoint is enabled in pre-init mode. Should we either remove this check or don't let the endpoint be available before init?

cmd/microcloud/status.go
```diff
@@ -0,0 +1,444 @@
...
+		}
+	}
+
+	// Format and colorize cells of the table.
+	rows := make([][]string, 0, len(statuses))
+	for _, s := range statuses {
+		rows = append(rows, formatStatusRow(localStatus, s))
+	}
+
+	t := table.New()
```
roosterfish, line 192: Worth having all of these calls in a helper? I wonder when we create more tables that we can save some repetitive tasks.

roosterfish (changes requested):
Looks great overall.

I have added a few comments in regards to colors, and errors in case some parts of the cluster are offline.

I saw when e.g. stoppping `microcloud` or `microceph` on any of the members and running `microcloud status` on anther member it prints the `Unable to connect...` error message instead of showing the offline status in the table.

cmd/style/style.go
masnax: Good point, I should test out the colors in the default ubuntu terminal theme.

api/types/status.go
```diff
@@ -0,0 +1,28 @@
+package types
+
+import (
+	cephTypes "github.com/canonical/microceph/microceph/api/types"
+	microTypes "github.com/canonical/microcluster/v2/rest/types"
+	ovnTypes "github.com/canonical/microovn/microovn/api/types"
+)
+
+// Status is a set of status information from a cluster member.
+type Status struct {
```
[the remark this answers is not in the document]
masnax: Yeah, I don't think dqlite role per project is very useful information here, as it shouldn't affect the day-to-day usage of MicroCloud.

api/status.go
masnax: Actually we have a constant for each services' port which we should be using here.

The main reason address is populated here is because it's populated for the `microcluster` projects, so I wanted the returned struct from LXD to be as close as possible to those in returned information, even if we're not using it all right now.

api/status.go
masnax: Same here, the role is supplied for consistency because it's automatically supplied for the `microcluster` projects.

cmd/style/style.go
masnax: Reducing contrast, mainly. I haven't checked it on an ubuntu themed terminal with a purple background yet though.

api/status.go
roosterfish: I see thanks.

cmd/style/style.go
roosterfish: Yeah the white on the default terminal isn't too bright. But good question if it's even a 255.

cmd/microcloud/status.go
masnax: It shouldn't be available prior to init, so I've removed the flag on the handler.

cmd/microcloud/status.go
masnax: Yeah this was a bug. If a cluster member can't be reached, we now populate it with empty node-specific information instead, and list it as unreachable.

api/types/status.go
masnax: Roles might get mixed up with dqlite so I used `units`. If you still think roles are better, I can switch it.

api/status.go
```diff
@@ -0,0 +1,209 @@
...
+
+			}
+
+			return nil
+		})
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		statuses = append(statuses, status)
```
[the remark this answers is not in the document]
masnax: Fixed. The slice is now sorted.

masnax:
@roosterfish This one's ready for review now. The styling package is a bit different now as we support `NO_COLOR` and `--no-color` as options to disable colouring. All errors are now also printed in red.

As well, the colours are adaptive based on your terminal background, and there's fewer of them, just red/yellow/green/white for text, and a few shades of grey for table borders.

I've also addressed your earlier comments about refactoring the `/status` API handler into separate functions for each service. 

Test failures are unrelated (see: #415)

cmd/microcloud/status.go
roosterfish: Great thanks, I'll give it another shot :)

cmd/microcloud/status.go
```diff
@@ -0,0 +1,451 @@
...
+	// Format the actual warnings based on the collected data.
+	warnings := Warnings{}
+	if clusterSize < 3 {
+		tmpl := tui.Fmt{Arg: "%s: %d systems are required for effective fault tolerance"}
+		msg := tui.Printf(tmpl,
+			tui.Fmt{Color: tui.Red, Arg: "Reliability risk", Bold: true},
+			tui.Fmt{Color: tui.Bright, Arg: 3, Bold: true},
+		)
+
+		warnings = append(warnings, Warning{Level: Error, Message: msg})
```
roosterfish, line 317: Should this be `Level: Warning` instead?

cmd/microcloud/status_test.go
```diff
@@ -0,0 +1,492 @@
...
+						types.MicroCloud: {genMember("micro01", microTypes.MemberOnline), genMember("micro02", microTypes.MemberOnline)},
+						types.LXD:        {genMember("micro01", microTypes.MemberOnline), genMember("micro02", microTypes.MemberOnline)},
+					},
+				},
+			},
+			expectedWarnings: []Warning{
+				{Level: Error, Message: "Reliability risk: 3 systems are required for effective fault tolerance"},
+				{Level: Warn, Message: "MicroOVN is not installed on micro01, micro02"},
+				{Level: Warn, Message: "MicroCeph is not installed on micro01, micro02"},
+				{Level: Warn, Message: "micro02 is not available"},
```
roosterfish, line 105: If a cluster member is offline, shouldn't this be an error?

roosterfish (changes requested):
Still looking very good, I especially like the custom format pattern with colors!

Whilst trying it out a second time I just felt that things reported as Error could be a Warning instead to not give the impression something is really broken.

cmd/microcloud/status.go
```diff
@@ -0,0 +1,451 @@
...
+}
+
+// Warnings is a list of warnings.
+type Warnings []Warning
+
+// Status returns the overall status of the warning list.
+// If there are any Error level warnings, the status will be error.
+// Otherwise, if there are any Warn level warnings, the status will be warn.
+// Finally, the status will be Success, implying no warnings.
+func (w Warnings) Status() StatusLevel {
```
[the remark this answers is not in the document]
masnax: I guess it's a larger design question about what we want to say MicroCloud deployments should look like. 

So far we have said that a "fully functioning" MicroCloud requires 3 systems for HA, and 3 systems providing OSDs for data resilience. 

I'd say the difference between an error and a warning is that the former is something you _should_ fix for safety/correctness and the latter is something that won't affect any running deployments, but doesn't cover all features of MicroCloud, or an issue that is likely to fix itself (like upgrades).

cmd/microcloud/status.go
masnax: My assumption so far was that we consider falling short of the HA targets to be basically "invalid". We should discuss this further.

cmd/microcloud/status_test.go
```diff
@@ -0,0 +1,492 @@
...
+						types.MicroCloud: {genMember("micro01", microTypes.MemberOnline), genMember("micro02", microTypes.MemberOnline)},
+						types.LXD:        {genMember("micro01", microTypes.MemberOnline), genMember("micro02", microTypes.MemberOnline)},
+					},
+				},
+			},
+			expectedWarnings: []Warning{
+				{Level: Error, Message: "Reliability risk: 3 systems are required for effective fault tolerance"},
+				{Level: Warn, Message: "MicroOVN is not installed on micro01, micro02"},
+				{Level: Warn, Message: "MicroCeph is not installed on micro01, micro02"},
+				{Level: Warn, Message: "LXD upgrade in progress"},
```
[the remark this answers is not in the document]
masnax: This is in tests because we control the input, but in reality LXD doesn't report upgrading status, only the micro* projects do, so unless we make changes to cluster status in LXD, this wouldn't be consistent.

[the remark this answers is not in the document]
roosterfish: Ok thanks!

cmd/microcloud/status.go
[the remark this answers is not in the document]
roosterfish: Sound good, but maybe the top level indicators could be HEALTHY, WARNING and FAULT? 
And if FAULT, you will then see some errors following your definition. 

This gives you at least some reassurance that the overall cluster status is not in an error state but something is off :)

cmd/microcloud/status.go
roosterfish: See https://github.com/canonical/microcloud/pull/384#discussion_r1795714711. I am happy if this stays as error then.

roosterfish:
One more I just saw. If I have a cluster of three nodes, and run `snap stop lxd` on the second one, I am getting this output warning me about Ceph even though it's just LXD which is offline on the second node:

```
root@micro01:~# microcloud status

 Status: ERROR

 ┃ ⨯ Data loss risk: MicroCeph OSD replication recommends at least 3 disks across 3 systems
 ┃ ! micro02 is not available

┌─────────┬─────────────┬──────┬─────────────────┬────────────────────────┬─────────┐
│  Name   │   Address   │ OSDs │ MicroCeph Units │     MicroOVN Units     │ Status  │
├─────────┼─────────────┼──────┼─────────────────┼────────────────────────┼─────────┤
│ micro01 │ 10.83.11.67 │  1   │   mds,mgr,mon   │ central,chassis,switch │ ONLINE  │
│ micro02 │ 10.83.11.97 │  0   │        -        │           -            │ Offline │
│ micro03 │ 10.83.11.7  │  1   │   mds,mgr,mon   │ central,chassis,switch │ ONLINE  │
└─────────┴─────────────┴──────┴─────────────────┴────────────────────────┴─────────┘
```

I would expect that both the Ceph and OVN units are still reported.

cmd/microcloud/status.go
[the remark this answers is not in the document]
masnax: After speaking a bit with Maria, I'm actually in agreement with you that we should make these warnings.

cmd/tui/style.go
```diff
@@ -0,0 +1,153 @@
...
+type ColorErr struct{}
+
+// Write colors the given error red and sends it to os.Stderr.
+func (*ColorErr) Write(p []byte) (n int, err error) {
+	return os.Stderr.WriteString(ErrorColor(strings.TrimSpace(string(p)), false) + "\n")
+}
+
+// PrintWarning calls Println but it appends "! Warning:" to the front of the message.
+func PrintWarning(s string) {
+	warning := fmt.Sprintf("%s %s: %s", WarningSymbol(), WarningColor("Warning", true), s)
```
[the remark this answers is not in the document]
masnax: Fixed.

api/status.go
```diff
@@ -0,0 +1,249 @@
...
+	return status, nil
+}
+
+func microStatus(ctx context.Context, microClient *microClient.Client, s service.Service, status *types.Status) (*types.Status, error) {
+	clusterMembers, err := microClient.GetClusterMembers(context.Background())
+	if err != nil && !lxdAPI.StatusErrorCheck(err, http.StatusServiceUnavailable) {
+		return nil, err
+	}
+
+	status.Clusters[s.Type()] = clusterMembers
```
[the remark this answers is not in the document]
masnax: I've reworked this based on your other comment about stopping LXD also breaking MicroCeph detection.

Now, each helper returns the fields it affects, and we assign the fields to `status` in the `RunConcurrent` block. I've also added a mutex for the map write.

cmd/microcloud/status.go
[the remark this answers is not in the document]
masnax: The OSD & HA messages now count as warnings.

cmd/microcloud/status_test.go
masnax: I've made these into errors, and now they also include a comma delimited list of services which are not available.

masnax:
> Thanks, only one last thing from my side. I see this "micro01 is not available on LXD" but shouldn't it be reversed: "LXD is not available on micro01"?

Good point, fixed now.

roosterfish:
Looks good to me!

PR #387, opened 2024-08-30
test/includes: Clean up microcloud_interactive logs

masnax:
How microcloud_interactive calculates its output is not useful to the test logs and is incredibly verbose, making it hard to read the logs for the workflow.

Instead, we can just print the final output to stderr to see if there is a mistake


simondeziel (approved):
Nice, TIL the `set -o` + `xtrace` thing.

PR #388, opened 2024-08-30
test/suites: Don't use FAN network in test suite

masnax:
See: https://github.com/canonical/lxd/issues/14025


tomponline (approved):
This is a good idea beyond the crash as micro cloud only supports ovn anyway and fan is on the way out.

PR #394, opened 2024-09-06
Fix underlay configuration for `preseed.yaml`

masnax:
The logic for detecting the OVN underlay configuration for preseed was not correct, it was relying on `c.state` which contains the resources available and already set up on existing cluster members, as well as new cluster members. It is used in interactive setup to manage additional configuration being supplied to the existing systems.

In preseed, we don't populate it yet because we only want to apply the precise recipe that is defined in the preseed, and not change any other cluster information implicitly (in interactive setup we can ask the user). 

Additionally, the input parsing was not correct, since it expected a CIDR per system however the ovn encap configuration wants an IP.

So this addresses that by instead using the list of systems parsed from the yaml itself in `p.Systems`, and uses the same set of interfaces that were detected for the internal ceph cluster network, since they have the same restrictions.  Also, rather than specifying a CIDR per system, instead we can specify the encap IP directly, and validate that it fits in one of the available subnets.

Preseed tests have also been updated for the new key, which is prefixed with `ovn_` now to be clear what it affects. The `cephfs ` key has also been moved into the global ceph configuration because it's not system-specific.

### DIFF
--- a/cmd/microcloud/main_init.go
+++ b/cmd/microcloud/main_init.go
@@ -651,6 +651,28 @@ func validateGatewayNet(config map[string]string, ipPrefix string, cidrValidator
 }
 
 func (c *initConfig) validateSystems(s *service.Handler) (err error) {
+	for _, sys := range c.systems {
+		if sys.MicroCephInternalNetworkSubnet == "" || sys.OVNGeneveAddr == "" {
+			continue
+		}
+
+		_, subnet, err := net.ParseCIDR(sys.MicroCephInternalNetworkSubnet)
+		if err != nil {
+			return fmt.Errorf("Failed to parse available network interface CIDR address: %q: %w", subnet, err)
+		}
+
+		underlayIP := net.ParseIP(sys.OVNGeneveAddr)
+		if underlayIP == nil {
+			return fmt.Errorf("OVN underlay IP %q is invalid", sys.OVNGeneveAddr)
+		}
+
+		if subnet.Contains(underlayIP) {
+			fmt.Printf("Warning: OVN underlay IP (%s) is shared with the Ceph cluster network (%s)\n", underlayIP.String(), subnet.String())
+
+			break
+		}
+	}
+
 	if !c.bootstrap {
 		return nil
 	}
--- a/cmd/microcloud/main_init_preseed.go
+++ b/cmd/microcloud/main_init_preseed.go
@@ -797,19 +797,6 @@ func (p *Preseed) Parse(s *service.Handler, c *initConfig) (map[string]InitSyste
 			}
 		}
 
-		if c.bootstrap {
-			osdHosts := 0
-			for _, system := range c.systems {
-				if len(system.MicroCephDisks) > 0 {
-					osdHosts++
-				}
-			}
-
-			if osdHosts < RecommendedOSDHosts {
-				fmt.Printf("Warning: OSD host count is less than %d. Distributed storage is not fault-tolerant\n", RecommendedOSDHosts)
-			}
-		}
-
 		for _, filter := range p.Storage.Local {
 			// No need to check filters anymore if each machine has a disk.
 			if len(zfsMachines) == len(c.systems) {
@@ -837,6 +824,19 @@ func (p *Preseed) Parse(s *service.Handler, c *initConfig) (map[string]InitSyste
 		}
 
 		c.systems[peer] = system
+	}
+
+	if c.bootstrap {
+		osdHosts := 0
+		for _, system := range c.systems {
+			if len(system.MicroCephDisks) > 0 {
+				osdHosts++
+			}
+		}
+
+		if osdHosts < RecommendedOSDHosts {
+			fmt.Printf("Warning: OSD host count is less than %d. Distributed storage is not fault-tolerant\n", RecommendedOSDHosts)
+		}
 	}
 
 	// Initialize Ceph network if specified.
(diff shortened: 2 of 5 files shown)


cmd/microcloud/main_init_preseed.go
```diff
@@ -564,55 +564,42 @@ func (p *Preseed) Parse(s *service.Handler, c *initConfig) (map[string]InitSyste
 			c.systems[peer] = system
 		}
 
-		// Check the preseed underlay network configuration against the available ips.
+		// TODO: call `s.Services[types.MicroOVN].(*service.OVNService).SupportsFeature(context.Background(), "custom_encapsulation_ip")`
```
roosterfish, line 567: I suspect for the next MicroCloud LTS release we would anyway not recommend installing an OVN which isn't using microcluster `v2`?

But nevertheless good to have it.

cmd/microcloud/main_init_preseed.go
```diff
@@ -564,55 +564,42 @@ func (p *Preseed) Parse(s *service.Handler, c *initConfig) (map[string]InitSyste
...
-								if underlays[sys.Name] == cidrAddr {
-									underlayCount = underlayCount + 1
-									goto out
-								}
-							}
+				for _, iface := range addressedInterfaces[sys.Name] {
+					for _, cidr := range iface.Addresses {
+						_, subnet, err := net.ParseCIDR(cidr)
+						if err != nil {
+							return nil, fmt.Errorf("Failed to parse available network interface %q cidr address: %q: %w", iface.Network.Name, cidr, err)
```
roosterfish, line 586: `CIDR` using uppercase?

cmd/microcloud/main_init_preseed.go
```diff
@@ -519,9 +519,9 @@ func (p *Preseed) Parse(s *service.Handler, c *initConfig) (map[string]InitSyste
 	// If an uplink interface was explicitly chosen, we will try to set up an OVN network.
 	explicitOVN := len(ifaceByPeer) > 0
 
-	cephInterfaces := map[string]map[string]service.DedicatedInterface{}
+	addressedInterfaces := map[string]map[string]service.DedicatedInterface{}
```
roosterfish, line 522: If the same interface is selected for both the OVN underlay and Ceph would it be helpful to print a message saying the same interface was selected twice?

Not that this isn't a valid setup, but if you proactively select dedicated interfaces, you might not want to mix those two.

roosterfish (approved):
LGTM! Only some minor comments but they can also be addressed later.

cmd/microcloud/main_init_preseed.go
masnax: It's still not fully clear which tracks are going to support what (except for the LTS) so I think keeping these extension checks consistent throughout the code will be useful if we ever have a parallel non-long-term support track with some more loose restrictions.

cmd/microcloud/main_init_preseed.go
masnax: Good idea, actually I first tried to subtract the selected interface from availability if it was chosen for the other service, but it's technically a valid setup (since by default everything uses the microcluster network anyway). 

I'll add a warning here.

roosterfish (changes requested):
LGTM, there is only one warning about usage of `Println` instead of `Printf`.

gabrielmougard (dismissed):
It looks good!

cmd/microcloud/main_init_preseed.go
roosterfish: Worth removing this comment now and doing the actual check after https://github.com/canonical/microcloud/pull/398 got merged?

cmd/microcloud/main_init.go
```diff
@@ -651,6 +651,28 @@ func validateGatewayNet(config map[string]string, ipPrefix string, cidrValidator
...
+
+		underlayIP := net.ParseIP(sys.OVNGeneveAddr)
+		if underlayIP == nil {
+			return fmt.Errorf("OVN underlay IP %q is invalid", sys.OVNGeneveAddr)
+		}
+
+		if subnet.Contains(underlayIP) {
+			fmt.Printf("Warning: OVN underlay IP (%s) is shared with the Ceph cluster network (%s)\n", underlayIP.String(), subnet.String())
+
+			break
```
roosterfish, line 672: Instead of breaking here should we also log a warning for other system's OVN underlay IPs that collide with the Ceph cluster network?

roosterfish (changes requested):
Only two smaller comments, rest still looks good.

cmd/microcloud/main_init.go
masnax: I think that would get a bit noisy, and not particularly helpful. If you care about conflicting IPs with the ceph cluster network, the amount of conflicting IPs shouldn't really affect your behaviour. 

If you don't want any conflicting IPs, then one error should suffice to get you to restart the configuration. If you don't care about it, then you're probably not going to care if it's 2+ that collide versus 1.

cmd/microcloud/main_init_preseed.go
masnax: Fixed.

roosterfish (approved):

masnax:
Looks like `SupportsFeature` still doesn't work because we haven't added the fix to a tagged release yet.

roosterfish (approved):

PR #396, opened 2024-09-09
CLI: Remove `--auto` flag

roosterfish:
This PR removes the `--auto` flag from both the `init` and `add` subcommands.

The internal behavior of the `initConfig`'s `autoSetup` is kept to still allow using the concept in situations like preseed.


roosterfish:
@masnax removed the obsolete checks for `c.autoSetup`. A few are kept as they are used as part of `service add` and `preseed`.

masnax (approved):
LGTM! I think it's sufficient for now, but we should revisit whether using this flag for the preseed/service cases is really necessary.

PR #398, opened 2024-09-12
Move MicroCloud to microcluster v2

masnax:


go.mod
```diff
@@ -1,34 +1,38 @@
 module github.com/canonical/microcloud/microcloud
 
-go 1.22.4
```
roosterfish, line 3: The commit message still mentions replacement of MicroCeph.

roosterfish (approved):
Thanks, LGTM!

cmd/microcloud/main.go
```diff
@@ -19,8 +20,6 @@ type CmdControl struct {
 
 	FlagHelp          bool
 	FlagVersion       bool
-	FlagLogDebug      bool
-	FlagLogVerbose    bool
```
MggMuggins, line 23: Are we maintaining CLI compatibility between microcloud 0.99 and 1.0?

masnax: No, I think the secure key exchange prevents that since creating/growing the cluster is pretty much the only thing the old CLI did, and the old mechanism isn't going to work anymore.

I think 1 of the reasons we do this for LXD is that the client can be installed separately from the daemon, which isn't the case for MicroCloud.

tomponline (approved):

PR #407, opened 2024-09-25
Add `--heartbeat` to `microcloudd`

MggMuggins:
Microcluster supports it so we might as well pass it through. If we're able to configure this for the snap service as well it would let us sleep for less time in the recovery tests.

### DIFF
--- a/cmd/microcloudd/main.go
+++ b/cmd/microcloudd/main.go
@@ -37,7 +37,8 @@ type cmdGlobal struct {
 type cmdDaemon struct {
 	global *cmdGlobal
 
-	flagMicroCloudDir string
+	flagMicroCloudDir     string
+	flagHeartbeatInterval time.Duration
 }
 
 func (c *cmdDaemon) Command() *cobra.Command {
@@ -139,7 +140,7 @@ func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
 		api.OVNProxy(s),
 	}
 
-	return s.Services[types.MicroCloud].(*service.CloudService).StartCloud(context.Background(), s, endpoints, c.global.flagLogVerbose, c.global.flagLogDebug)
+	return s.Services[types.MicroCloud].(*service.CloudService).StartCloud(context.Background(), s, endpoints, c.global.flagLogVerbose, c.global.flagLogDebug, c.flagHeartbeatInterval)
 }
 
 func main() {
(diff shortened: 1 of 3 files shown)


cmd/microcloudd/main.go
```diff
@@ -160,6 +161,7 @@ func main() {
 	app.PersistentFlags().BoolVarP(&daemonCmd.global.flagLogVerbose, "verbose", "v", false, "Show all information messages")
 
 	app.PersistentFlags().StringVar(&daemonCmd.flagMicroCloudDir, "state-dir", "", "Path to store state information for MicroCloud"+"``")
+	app.PersistentFlags().DurationVar(&daemonCmd.flagHeartbeatInterval, "heartbeat", time.Second*10, "Time between attempted heartbeats")
```
roosterfish, line 164: Worth indicating in the comment that the default is 10s?

roosterfish (changes requested):

cmd/microcloudd/main.go
MggMuggins: Cobra adds it for us :)
```
MicroCloud daemon

Usage:
  microcloudd [flags]

Flags:
  -d, --debug                Show all debug messages
      --heartbeat duration   Time between attempted heartbeats (default 10s)
```

roosterfish: Ah true, thanks!

roosterfish (approved):

roosterfish:
> If we're able to configure this for the snap service

Can you think of a situation in which a user would modify the default of 10s?
I could think of a concept like we have in LXD where you can set the debug flag using `snap set lxd daemon.debug=true`. But wondering about the specific use case here.

Anyway I'll merge it now as it's currently more like a hidden feature we can use in the pipeline to speed up the tests.

MggMuggins:
I should have closed this; Max and I decided to hold off merging since this is not possible to consume from a snap deployment. I think he didn't want to do snap config in addition to the config that we can set in MicroCloud itself.

@masnax let me know if you'd like a revert of `--heartbeat`.

roosterfish:
> I should have closed this; Max and I decided to hold off merging since this is not possible to consume from a snap deployment. I think he didn't want to do snap config in addition to the config that we can set in MicroCloud itself.

Ups, my bad then. Wasn't aware of the details.

masnax:
@roosterfish @MggMuggins No no, I'm okay with this. Only thing is we should have a minimum heartbeat if we're going to make this user-configurable with a snap config key. Eventually, somebody is definitely going to set it to 0 thinking they're unsetting it.

PR #408, opened 2024-09-26
Trivy vulnerability scanning

hamistao:
This adds two jobs to our workflows. One uses Trivy to scan MicroCloud's dependencies and the other scans the snap.


masnax (approved):
lgtm!

PR #414, opened 2024-10-08
cmd/microcloudd: Set minimum heartbeat value

MggMuggins:
Follow-up from #407 

200ms is somewhat arbitrary; we want to explicitly disallow 0 but still let fairly small values through for future use in the test suite.

### DIFF
--- a/cmd/microcloudd/main.go
+++ b/cmd/microcloudd/main.go
@@ -18,6 +18,10 @@ import (
 	"github.com/canonical/microcloud/microcloud/version"
 )
 
+// MinimumHeartbeatInterval is used to prevent a user from setting the MicroCluster
+// heartbeat to 0.
+const MinimumHeartbeatInterval = time.Millisecond * 200
+
 // Debug indicates whether to log debug messages or not.
 var Debug bool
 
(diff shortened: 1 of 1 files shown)


cmd/microcloudd/main.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/canonical/microcloud/microcloud/version"
 )
 
+const MinimumHeartbeatInterval = time.Millisecond * 200
```
masnax, line 21: Exported declarations should have a comment. (static analysis should have caught this, guess it's incomplete in microcloud).

masnax (changes requested):

cmd/microcloudd/main.go
```diff
@@ -58,6 +60,10 @@ func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
 		return cmd.Help()
 	}
 
+	if c.flagHeartbeatInterval < MinimumHeartbeatInterval {
+		return fmt.Errorf("Invalid heartbeat interval: must be >%s", MinimumHeartbeatInterval)
```
roosterfish, line 64:
```suggestion
		return fmt.Errorf("Invalid heartbeat interval: Must be >%s", MinimumHeartbeatInterval)
```

roosterfish (changes requested):
Thanks for the follow up, only a small nit.

roosterfish (approved):

roosterfish:
Is 200ms a strict limitation or did you see values less than 200ms cause trouble elsewhere?

MggMuggins:
It's entirely arbitrary, I just needed a duration >0. It's used as the [dqlite cluster role adjustment frequency](https://pkg.go.dev/github.com/canonical/go-dqlite@v1.22.0/app#WithRolesAdjustmentFrequency) and the MicroCluster heartbeat interval, both of which used to default to 30s. If this were MicroCluster I'd be a little more cautious about the value. You're likely right that there would be issues turning it below 200ms, but I have no evidence for that.

masnax (approved):

PR #415, opened 2024-10-08
.github/workflows: Refresh snapd before installing LXD

masnax:
Attempt to fix this error in gh workflows:

```
error: cannot install "lxd": snap "lxd" assumes unsupported features: snapd2.65
       (try to refresh snapd)
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -211,7 +211,19 @@ jobs:
       - name: "Setup host LXD"
         run: |
           set -eux
-          sudo snap install lxd --channel latest/edge || sudo snap refresh lxd --channel latest/edge
+
+          if snap list snapd > /dev/null ; then
+            sudo snap refresh snapd
+          else
+            snap install snapd
+          fi
+
+          if snap list lxd > /dev/null ; then
+            sudo snap refresh lxd --channel latest/edge
+          else
+            snap install lxd --channel latest/edge
+          fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -211,7 +211,19 @@ jobs:
...
+          if snap list snapd > /dev/null ; then
+            sudo snap refresh snapd
+          else
+            snap install snapd
+          fi
+
+          if snap list lxd > /dev/null ; then
+            sudo snap refresh lxd --channel latest/edge
+          else
+            snap install lxd --channel latest/edge
```
roosterfish, line 224: `sudo` missing?

roosterfish (changes requested):

.github/workflows/tests.yml
roosterfish: Also on my local system I had to use `snap refresh snapd --edge` to get `2.65`.
Without `--edge` it didn't refresh even though I was tracking `latest/stable` which currently holds `2.65.3` according to `snap info snapd`.

roosterfish: The change on the LXD snap got reverted so probably the PR isn't required anymore. See https://github.com/canonical/lxd-pkg-snap/pull/565

tomponline: yeah snapd got rolled back

PR #417, opened 2024-10-09
ceph: Add support for Ceph public network seggregation

gabrielmougard:
JIRA ticket: https://warthogs.atlassian.net/browse/LXD-1808

### DIFF
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -490,10 +490,10 @@ func validateCephInterfacesForSubnet(lxdService *service.LXDService, systems map
 
 // getTargetCephNetworks fetches the Ceph network configuration from the existing Ceph cluster.
 // If the system passed as an argument is nil, we will fetch the local Ceph network configuration.
-func getTargetCephNetworks(sh *service.Handler, s *InitSystem) (internalCephNetwork *net.IPNet, err error) {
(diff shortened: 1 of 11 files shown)


masnax (approved):
LGTM, thanks @gabrielmougard!

roosterfish (changes requested):
Hi, the changes are looking good. I have only found one situation in which if you only set the public network, it overwrites the default setting for the internal traffic too:

```
...
What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph internal traffic on? [default: 10.83.11.0/24] 
What subnet (either IPv4 or IPv6 CIDR notation) would you like your Ceph public traffic on? [default: 10.83.11.0/24] 1.1.2.0/24
...

root@micro01:~# microceph.ceph config dump
WHO     MASK  LEVEL     OPTION                       VALUE       RO
global        advanced  cluster_network              1.1.2.0/24  * 
global        advanced  mon_allow_pool_size_one      true          
global        advanced  osd_pool_default_crush_rule  1             
global        advanced  osd_pool_default_size        1  

root@micro01:~# microceph.ceph config show osd.1 cluster_network
1.1.2.0/24
root@micro01:~# microceph.ceph config show osd.1 public_network
1.1.2.0/24
```

I understand if you select a custom subnet for the internal traffic, it prompts you to use the same also for the public traffic.
But in this case if you only set the public traffic subnet, it also takes this value for the internal traffic even though the dialogue says `default: 10.83.11.0/24`.

gabrielmougard:
@roosterfish should be fixed now :)

PR #418, opened 2024-10-09
Discovery: Replace mDNS library with Go's multicast from std library

roosterfish:
This PR removes the dependency from the mDNS library and instead makes use of Go's native multicast functionality which is part of the standard library.

Another port 9444 is introduced for the use of multicast UDP traffic together with the address 239.100.100.100. [IANA](https://www.iana.org/assignments/multicast-addresses/multicast-addresses.xhtml#multicast-addresses-12) shows the range 239.0.0.0-239.255.255.255 to be used for organization-local workloads.
I have not reused the designated mDNS port and multicast address as our protocol is not mDNS but a basic question <-> answer pattern using a JSON payload.

Furthermore the docs are adapted to remove any mentioning of mDNS.

The commit https://github.com/roosterfish/microcloud/commit/46fd760dcffa337a73a8c3d504de6e4fbe28a848 is reverted as it looks to be a problem with LXD which wrongly masquerades multicast traffic if `br_netfilter` is enabled on the host.


roosterfish:
@masnax this is now ready for review. 

The GitHub runners have `br_netfilter` module loaded which causes the native LXD bridge to falsely masquerade multicast traffic too. 
I am checking what we can do on LXD side to resolve this but disabling the `br_netfilter` kernel module for the tests fixes it and also allowed me to revert the temporary fix we put into place when disabling IPv6.

masnax (approved):
LGTM! Thanks @roosterfish

masnax:
Is this worth testing a bit more in environments outside of LXD VMs, like the old MicroCloud deployment box? If not then we can go ahead and merge.

roosterfish:
> Is this worth testing a bit more in environments outside of LXD VMs, like the old MicroCloud deployment box?

Yeah I'll give it a try in the

roosterfish:
@masnax I have now tested it between two Ubuntu nodes using an actual physical cable. Works as expected.

masnax (approved):

PR #419, opened 2024-10-09
test/includes/microcloud: install yq from deb now that we are on 24.04

simondeziel:
Installing from deb should be faster than pulling from the snapstore. Before switching to 24.04, this was not possible as Noble is the first release with a yq deb.


roosterfish (approved):

simondeziel:
The one failure is because the `basic` tests has picked up a too recent LXD latest/edge build that's still broken. The important test, the `recover` one is green at least.

PR #420, opened 2024-10-09
Tiobe Integration

kadinsayani:
This PR adds code coverage for unit tests and snap system tests.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,6 +6,10 @@ on:
     branches:
       - main
   pull_request:
+  workflow_dispatch:
+
+env:
+  GOCOVERDIR: ${{ ( github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' ) && '/home/runner/work/microcloud/microcloud/cover' || '' }}
 
 permissions:
   contents: read
@@ -73,8 +77,30 @@ jobs:
       - name: Run static analysis
         run: make check-static
 
+      - name: Make GOCOVERDIR
+        run: mkdir -p "${GOCOVERDIR}"
+        if: env.GOCOVERDIR != ''
+
       - name: Unit tests
-        run: make check-unit
+        run: |
+          set -eux
+          make check-unit 
+
+      - name: Upload coverage data
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-unit
+          path: ${{env.GOCOVERDIR}}
+        if: env.GOCOVERDIR != '' 
+
+      - name: Upload system test dependencies
+        uses: actions/upload-artifact@v4
+        with:
+          name: system-test-deps
+          path: |
+            /home/runner/go/bin/microcloud
+            /home/runner/go/bin/microcloudd
+          retention-days: 1 
 
   system-tests:
     env:
@@ -82,7 +108,8 @@ jobs:
       SKIP_VM_LAUNCH: "1"
       SNAPSHOT_RESTORE: "1"
     name: System
-    runs-on: GitHubMicrocloud
+    runs-on: GitHubMicrocloud 
+    needs: code-tests
     strategy:
       fail-fast: false
       matrix:
@@ -193,11 +220,20 @@ jobs:
           sudo apt-get update
           sudo apt-get install --no-install-recommends -y libdqlite-dev pkg-config
 
-      - name: Build
+      - name: Download system test dependencies
+        uses: actions/download-artifact@v4
+        with:
+          name: system-test-deps
+          merge-multiple: true
+          path: /home/runner/go/bin
+
+      - name: Make GOCOVERDIR
+        run: mkdir -p "${GOCOVERDIR}"
+        if: env.GOCOVERDIR != ''
+
+      - name: Sideload debug binaries
         run: |
           set -eux
-
-          make
 
           # Binaries to sideload
           export MICROCLOUD_DEBUG_PATH=~/go/bin/microcloud
@@ -259,7 +295,7 @@ jobs:
           export MICROCLOUD_SNAP_CHANNEL="${{ matrix.microcloud }}"
 
           cd test
-          sudo --preserve-env=DEBUG,GITHUB_ACTIONS,MICROCLOUD_DEBUG_PATH,MICROCLOUDD_DEBUG_PATH,SKIP_VM_LAUNCH,SNAPSHOT_RESTORE,TEST_STORAGE_SOURCE,TESTBED_READY,BASE_OS,LXD_SNAP_CHANNEL,MICROCEPH_SNAP_CHANNEL,MICROOVN_SNAP_CHANNEL,MICROCLOUD_SNAP_CHANNEL ./main.sh setup
+          sudo --preserve-env=GOCOVERDIR,DEBUG,GITHUB_ACTIONS,MICROCLOUD_DEBUG_PATH,MICROCLOUDD_DEBUG_PATH,SKIP_VM_LAUNCH,SNAPSHOT_RESTORE,TEST_STORAGE_SOURCE,TESTBED_READY,BASE_OS,LXD_SNAP_CHANNEL,MICROCEPH_SNAP_CHANNEL,MICROOVN_SNAP_CHANNEL,MICROCLOUD_SNAP_CHANNEL ./main.sh setup
 
           echo "TESTBED_READY=1" >> "${GITHUB_ENV}"
           echo "BASE_OS=${BASE_OS}" >> "${GITHUB_ENV}"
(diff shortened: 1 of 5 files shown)


tomponline (dismissed):
Let me know when you need secrets.TICS_AUTH_TOKEN  setup

kadinsayani:
For the system tests using the MicroCloud step, the only requirements to generate the coverage files are:
- GOCOVERDIR needs to be set in the snap confinement space - I'm creating a directory at `/var/snap/microcloud/common/data/coverage`; and 
- The side loaded binary should be compiled with the `-cover` flag.

@tomponline, once I've got it working here, we can utilize a similar approach for LXD.

kadinsayani:
There seems to be an issue with the MicroCloud ci tests due to a regression relating to upstream Ceph versioning so I haven't tested this yet.

kadinsayani:
Should be good to go, just waiting on https://github.com/kadinsayani/microcloud/actions/runs/11408460306/job/31750619743.

kadinsayani:
Good to go on this, here's an example successful run on my fork: https://github.com/kadinsayani/microcloud/actions/runs/11413516898/job/31763088034.

tomponline (approved):

kadinsayani:
We're using the `go1.23.1` toolchain. To use `StaticCheck` with Go 1.23, it needs to be built with Go 1.23. Hence why the Tics job throws errors like this one:

```
2024-10-19 04:00:23.592: [INFO 3012] Analyzing 30/58 /home/runner/work/microcloud/microcloud/cmd/microcloud/session.go
start analysis: 2024-10-19 04:00:23.592
  Initialization
  Initialization time: 0.001 sec
Error: [ERROR 5073] The following error(s) occurred while 'StaticCheck' was running: '/home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.1.linux-amd64/src/maps/iter.go:51:20: cannot range over seq (variable of type iter.Seq2[K, V])'.
```

These errors are what caused the downtick in TQI scores across various Tiobe quality metrics. The analyzer runs are erroring out, so data isn't being sent to Tics for the coding standards runs. 

For the security checks, I'm seeing the following error on every run:

```
[ERROR 5255] Command 'cov-build' failed with exit status 256, error: -fno-stack-protector ./cgo.go See log '/tmp/tics/11413516898_1_tics_tics-github-action/ticstmpdir/Coverity/cov-build-console.log'.
```

Tics uses Coverity for security scanning, here's something I found from their docs:

> By default, cov-build returns the exit code from the native build.

Exit code 256 typically indicates that the system can't locate the binary to run the command, so I think this is also a dependency issue.

I propose that we bump the Go version in the Tics job to ensure we capture the right dependencies for Tics at the `Install dependencies` step. However, we may need to also bump the Go version for MicroCloud and the entire `Tests` action.

kadinsayani:
Good to go - all of the Tiobe runs are operational and the quality metrics are accurate.

test/includes/microcloud.sh
```diff
@@ -271,8 +271,8 @@ set_debug_binaries() {
     lxc exec "${name}" -- rm -f /var/snap/microcloud/common/microcloudd.debug
     lxc exec "${name}" -- rm -f /var/snap/microcloud/common/microcloud.debug
 
-    lxc file push --quiet "${MICROCLOUDD_DEBUG_PATH}" "${name}"/var/snap/microcloud/common/microcloudd.debug
-    lxc file push --quiet "${MICROCLOUD_DEBUG_PATH}" "${name}"/var/snap/microcloud/common/microcloud.debug
+    lxc file push --quiet "${MICROCLOUDD_DEBUG_PATH}" "${name}"/var/snap/microcloud/common/microcloudd.debug --mode 0755
```
roosterfish, line 274: Out of interest what is the reason for changing the permission for the debug binaries?

staticcheck.conf
```diff
@@ -0,0 +1,3 @@
+# Checks being ignored:
```
roosterfish, line 1: Why is this required?

test/main.sh
```diff
@@ -192,6 +192,8 @@ run_test() {
 	${TEST_CURRENT}
 	END_TIME="$(date +%s)"
 
+	collect_go_cover_files
```
roosterfish, line 195: Why is it not sufficient to do the collection here but also in `reset_systems` and `restore_systems`?

.github/workflows/tests.yml
```diff
@@ -166,11 +194,22 @@ jobs:
           sudo apt-get update
           sudo apt-get install --no-install-recommends -y libdqlite-dev pkg-config
 
+      - name: Download system test dependencies
```
roosterfish, line 197: I understand here we fetch the deps from the previous build but in the step after we again run `make` to build them again. What is the reason for this?

roosterfish (changes requested):
Thanks for this! A few comments from my side, mostly for clarification.

test/includes/microcloud.sh
kadinsayani: `lxc file push` sets the file mode to `0644` by default - we want to make sure the binaries are executable and maintain the file mode set by the Go compiler. We also check for the presence of these executables in `commands/daemon.start` and `commands/microcloud` in the MicroCloud snap. When they are present and executable, the `GOCOVERDIR` is created. Here's the relevant PR: https://github.com/canonical/microcloud-pkg-snap/pull/15.

staticcheck.conf
kadinsayani: We wanted to see if adding a `staticcheck.conf` with this rule would silence the Coding standards violations, but it looks like Tics isn't picking this up so our Coding standards TQI score remains the same. See https://github.com/kadinsayani/microcloud/actions/runs/11446739960/job/31905615699.

test/main.sh
kadinsayani: A few of the test suites use `reset_systems` during test cases. `reset_systems` may call `restore_systems` when `SNAPSHOT_RESTORE` is set to `1`.

.github/workflows/tests.yml
kadinsayani: We probably don't need this step actually. Thanks for pointing this out!

I replicated logic from the `lxd` and `microcluster` GitHub Tiobe actions and overlooked this.

staticcheck.conf
kadinsayani: We can still keep this staticcheck rule since it doesn't apply to us.

test/includes/microcloud.sh
tomponline: You can set mode with file push

https://documentation.ubuntu.com/lxd/en/latest/reference/manpages/lxc/file/push/

.github/workflows/tests.yml
kadinsayani: I've cleaned up the build steps :)

kadinsayani:
I've rebased and resolved a naming issue with the coverage file directories. Test run: https://github.com/kadinsayani/microcloud/actions/runs/11493847667/job/31990262151.

kadinsayani:
> I've rebased and resolved a naming issue with the coverage file directories. Test run: https://github.com/kadinsayani/microcloud/actions/runs/11493847667/job/31990262151.

Looks like there's still an issue with the naming. I'll try and fix this some time today. Just need a unique identifier to append so there aren't naming conflicts.

staticcheck.conf
roosterfish: Thanks!

test/includes/microcloud.sh
roosterfish: Are you sure LXD changes to permissions to 0644 by default? From what I have observed it picks up the same permissions as they are present on the host.

.github/workflows/tests.yml
```diff
@@ -259,7 +295,7 @@ jobs:
           export MICROCLOUD_SNAP_CHANNEL="${{ matrix.microcloud }}"
 
           cd test
-          sudo --preserve-env=DEBUG,GITHUB_ACTIONS,MICROCLOUD_DEBUG_PATH,MICROCLOUDD_DEBUG_PATH,SKIP_VM_LAUNCH,SNAPSHOT_RESTORE,TEST_STORAGE_SOURCE,TESTBED_READY,BASE_OS,LXD_SNAP_CHANNEL,MICROCEPH_SNAP_CHANNEL,MICROOVN_SNAP_CHANNEL,MICROCLOUD_SNAP_CHANNEL ./main.sh setup
+          sudo --preserve-env=PATH,GOPATH,GOCOVERDIR,DEBUG,GITHUB_ACTIONS,MICROCLOUD_DEBUG_PATH,MICROCLOUDD_DEBUG_PATH,SKIP_VM_LAUNCH,SNAPSHOT_RESTORE,TEST_STORAGE_SOURCE,TESTBED_READY,BASE_OS,LXD_SNAP_CHANNEL,MICROCEPH_SNAP_CHANNEL,MICROOVN_SNAP_CHANNEL,MICROCLOUD_SNAP_CHANNEL ./main.sh setup
```
roosterfish, line 298: Just one more here, what is the reason we have to pass the `PATH` and `GOPATH` variables here? 
The test suite grew the `collect_go_cover_files` func but it doesn't look to consume any executable out of those two paths?

.github/workflows/tests.yml
```diff
@@ -273,7 +309,79 @@ jobs:
           set -eux
           chmod +x ~
           cd test
-          sudo --preserve-env=DEBUG,GITHUB_ACTIONS,MICROCLOUD_DEBUG_PATH,MICROCLOUDD_DEBUG_PATH,SKIP_VM_LAUNCH,SNAPSHOT_RESTORE,TEST_STORAGE_SOURCE,TESTBED_READY,BASE_OS,LXD_SNAP_CHANNEL,MICROCEPH_SNAP_CHANNEL,MICROOVN_SNAP_CHANNEL,MICROCLOUD_SNAP_CHANNEL ./main.sh ${{ matrix.suite }}
+          sudo --preserve-env=PATH,GOPATH,GOCOVERDIR,DEBUG,GITHUB_ACTIONS,MICROCLOUD_DEBUG_PATH,MICROCLOUDD_DEBUG_PATH,SKIP_VM_LAUNCH,SNAPSHOT_RESTORE,TEST_STORAGE_SOURCE,TESTBED_READY,BASE_OS,LXD_SNAP_CHANNEL,MICROCEPH_SNAP_CHANNEL,MICROOVN_SNAP_CHANNEL,MICROCLOUD_SNAP_CHANNEL ./main.sh ${{ matrix.suite }}
```
roosterfish, line 312: Same comment here on the `PATH` and `GOPATH`.

test/includes/microcloud.sh
kadinsayani: LXD doesn't change the permissions to 0644 by default unless it can't grab the host file permissions. When I initially tested this, the `GOCOVERDIR` wasn't being created and I also didn't see the echo'd message "==> WARNING: Using a custom debug MicroCloud binary!". Upon further investigation, it looks like the upload artifact GitHub action doesn't maintain permissions: https://github.com/actions/upload-artifact?tab=readme-ov-file#permission-loss. We can either tar the binaries together prior to uploading the artifact or continue using `lxc file push` with the mode flag.

.github/workflows/tests.yml
kadinsayani: I added `PATH` and `GOPATH` for verbosity but upon further investigation I don't believe they are required - only `GOCOVERDIR` is required.

.github/workflows/tests.yml
kadinsayani: Same as https://github.com/canonical/microcloud/pull/420#discussion_r1822738679.

test/includes/microcloud.sh
roosterfish: I see ok, thanks for checking!

.github/workflows/tests.yml
roosterfish: Ok, can we get rid of `PATH` and `GOPATH` then?

kadinsayani: Sure thing!

test/includes/microcloud.sh
kadinsayani: No problem!

.github/workflows/tests.yml
kadinsayani: Done :)

roosterfish (approved):
Thanks!

PR #421, opened 2024-10-11
Daemon version enforcement

masnax:
Adds version enforcement to MicroCloud.

The `services` list used to keep track of the set of installed services is repurposed as a `map[types.ServiceType]string` which maps each installed service to the version it's installed at. Then, when validating the set of services when receiving a join intent, we also check that the versions match exactly.

When running any `microcloud` initialization command, we first compile the list of installed services, along with their versions, and check that they match the pre-defined minimum versions. For semantic versions like LXD and MicroCloud, we only check against the major and minor versions, so patch versions are acceptable. Patch versions must still match between initiators and joiners.

For MicroCeph, we parse out the ceph version. This is not a standard semantic version (see: https://docs.ceph.com/en/latest/releases/general/#understanding-the-release-cycle), however, we can treat it like one because we require the first 2 numbers to be equal, and the last represents bugfixes which are allowed.

MicroOVN uses a named version, represented by the release window (for example 24.03), so we can use direct string comparison here.

### DIFF
--- a/api/session_join.go
+++ b/api/session_join.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -51,7 +52,7 @@ func sessionJoinPost(sh *service.Handler) func(state state.State, r *http.Reques
 			// Only validate the intent (services) on the initiator.
 			// The joiner has to accept the services from the initiator.
 			if session.Role() == types.SessionInitiating {
-				err = validateIntent(sh, req)
+				err = validateIntent(r.Context(), sh, req)
 				if err != nil {
 					return api.NewStatusError(http.StatusBadRequest, err.Error())
 				}
@@ -83,11 +84,22 @@ func sessionJoinPost(sh *service.Handler) func(state state.State, r *http.Reques
 
 // validateIntent validates the given join intent.
 // It checks whether or not the peer is missing any of our services and returns an error if one is missing.
-func validateIntent(sh *service.Handler, intent types.SessionJoinPost) error {
+// Also compares each service's daemon version between the joiner and initiator.
+func validateIntent(ctx context.Context, sh *service.Handler, intent types.SessionJoinPost) error {
 	// Reject any peers that are missing our services.
-	for service := range sh.Services {
-		if !shared.ValueInSlice(service, intent.Services) {
-			return fmt.Errorf("Rejecting peer %q due to missing services (%s)", intent.Name, string(service))
+	for _, service := range sh.Services {
+		intentVersion, ok := intent.Services[service.Type()]
+		if !ok {
+			return fmt.Errorf("Rejecting peer %q due to missing services (%s)", intent.Name, string(service.Type()))
+		}
+
+		version, err := service.GetVersion(ctx)
+		if err != nil {
+			return fmt.Errorf("Unable to determine initiator's %s version: %w", service.Type(), err)
+		}
+
+		if intentVersion != version {
+			return fmt.Errorf("Rejecting peer %q due to invalid %s version. (Want: %q, Detected: %q)", intent.Name, service.Type(), version, intentVersion)
 		}
 	}
 
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
-	"slices"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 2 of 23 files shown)


service/version.go
```diff
@@ -0,0 +1,55 @@
...
+	"strings"
+
+	"golang.org/x/mod/semver"
+
+	"github.com/canonical/microcloud/microcloud/api/types"
+)
+
+const (
+	// lxdMinVersion is the minimum version of LXD that fully supports all MicroCloud features.
+	lxdMinVersion = "6.1"
```
roosterfish, line 14: I think we still have to use `5.21` as this is the LTS.

service/microceph.go
```diff
@@ -262,6 +263,25 @@ func (s CephService) Port() int64 {
 	return s.port
 }
 
+// GetVersion gets the installed daemon version of the service, and returns an error if the version is not supported.
+func (s CephService) GetVersion(ctx context.Context) (string, error) {
+	status, err := s.m.Status(ctx)
+	if err != nil && api.StatusErrorCheck(err, http.StatusNotFound) {
```
roosterfish, line 269: Is this to support the check on older snap versions that have the status endpoint on a different URL?

api/session_join.go
```diff
@@ -83,11 +84,22 @@ func sessionJoinPost(sh *service.Handler) func(state state.State, r *http.Reques
...
-			return fmt.Errorf("Rejecting peer %q due to missing services (%s)", intent.Name, string(service))
+	for _, service := range sh.Services {
+		intentVersion, ok := intent.Services[service.Type()]
+		if !ok {
+			return fmt.Errorf("Rejecting peer %q due to missing services (%s)", intent.Name, string(service.Type()))
+		}
+
+		version, err := service.GetVersion(ctx)
+		if err != nil {
+			return fmt.Errorf("Unable to determine initiator's %s version", service.Type())
```
roosterfish, line 98: `"Unable to determine initiator's %s version: %w"`

cmd/microcloud/ask.go
```diff
@@ -1602,14 +1601,18 @@ func (c *initConfig) askJoinConfirmation(gw *cloudClient.WebsocketGateway, servi
 
 	fmt.Println("Successfully joined the MicroCloud cluster and closing the session.")
 
-	// Filter out MicroCloud.
-	services = slices.DeleteFunc(services, func(t types.ServiceType) bool {
-		return t == types.MicroCloud
-	})
+	newServices := make(map[types.ServiceType]string, len(services)-1)
+	for serviceType, version := range services {
+		if serviceType == types.MicroCeph {
```
roosterfish, line 1606: `types.MicroCloud`?

roosterfish (changes requested):
Looking great, only a few smaller remarks.

service/version.go
```diff
@@ -0,0 +1,55 @@
...
+	case types.MicroOVN:
+		if daemonVersion != microOVNMinVersion {
+			return fmt.Errorf("%s version %q is not supported", serviceType, daemonVersion)
+		}
+
+	case types.MicroCeph:
+		// Parse out the underlying ceph version from the version string.
+		before, after, ok := strings.Cut(daemonVersion, "ceph-version: ")
+		cephVersion := strings.Split(after, "~")
+		if !ok || before != "" || !strings.Contains(after, "~") || len(cephVersion) != 2 || cephVersion[0] == "" {
```
roosterfish, line 42: I was thinking a bit on this one whether or not it would make sense to use a regex instead to only match the version at the beginning of the string so we are protected against upstream version schema changes and don't have to deal with the `~` and splitting on it.

But if that happens we can possibly react on it quite quickly anyway so not a blocker.

service/microceph.go
masnax: Yeah, basically anything short of microcluster v2 is unsupported.

service/version.go
masnax: Oops, got too caught up in `.1` LTS syntax. Fixed now.

api/session_join.go
masnax: Fixed.

cmd/microcloud/ask.go
masnax: Fixed.

service/version.go
masnax: If there's a schema change, I wouldn't think it's safe to assume the first encountered semver will correspond to what we care about, we'd still have to verify manually. So in that sense I think it's fine to do string manipulation since it's easier to see what is happening to the string.

roosterfish (approved):
LGTM!

masnax:
> LGTM!

So we're explicitly blocking `6.1` now which means we can't use `latest/edge` in the test suite anymore for LXD. I've switched it to use 5.21/candidate, since it's a higher revision than 5.21/edge.

roosterfish (approved):

roosterfish (approved):

tomponline:
> 5.21/edge

Suggest using 5.21/edge to catch regressions earlier

tomponline:
> > > 5.21/edge
> > 
> > 
> > Suggest using 5.21/edge to catch regressions earlier
> 
> I went with `5.21/candidate` since it had a higher revision number, but if `5.21/edge` is better I'll switch to that.
> 
> Once we create a branch for the LTS, we can go back to using `latest/edge` in `main` for the next release.
> 
> I'll merge this now and make a follow-up PR to fix the test version. The one failing test is unrelated to this PR, seems that MicroOVN is timing out for some other reason.

Yeah the revision number being higher doesn't mean much (apart from it was built more recently) but doesnt indicate more recent contents of the snap. In this case 5.21/candidate has recently been updated to include a ui update but is still older than 5.21/edge

PR #422, opened 2024-10-11
docs: Update initialise.md

wideawakening:
minor notes to help on 
```
Insufficient number of disks available to set up distributed storage, skipping at this time    
```
it took me some time to realize i was allocating all my storage for local and had nothing left for distributed :D

### DIFF
--- a/doc/how-to/initialise.md
+++ b/doc/how-to/initialise.md
@@ -52,8 +52,9 @@ Complete the following steps to initialise MicroCloud:
 1. Select whether you want to set up local storage.
 
    ```{note}
-   To set up local storage, each machine must have a local disk.
-   The disks must not contain any partitions.
+   - To set up local storage, each machine must have a local disk.
+   - The disks must not contain any partitions.
+   - A disk used for local storage will not be available for distributed storage.
    ```
 
    If you choose `yes`, configure the local storage:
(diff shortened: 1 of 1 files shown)


doc/how-to/initialise.md
```diff
@@ -66,8 +67,9 @@ Complete the following steps to initialise MicroCloud:
 1. Select whether you want to set up distributed storage (using MicroCeph).
 
    ```{note}
-   To set up distributed storage, you need at least three additional disks on at least three different machines.
-   The disks must not contain any partitions.
+   - To set up distributed storage, you need at least three additional disks on at least three different machines.
+   - The disks must not contain any partitions.
+   - A disk used for local storage will not be available for distributed storage.
```
roosterfish, line 72: When you are configuring distributed storage the local storage is already configured. You cannot do it the other way around. 

I would suggest removing this line as you can safely pick any number of disks for distributed storage at this point without affecting local storage.

roosterfish (changes requested):
Thanks for the PR! Please have a look at my comment.

doc/how-to/initialise.md
wideawakening: you mean line 72, right?
i made it extra redundant, because the problem of not finding storage for distributed, happens at this step, after local storage setup.

maybe the proposal of my wording could be more explanatory, but i would not remove it.
i will not push much on this;  let me know how you feel about this other option

> A disc that was previously selected for local storage will not be shown for distributed storage

roosterfish: I see. Can you then please rephrase it to `A disk that was previously selected for local storage will not be shown for distributed storage.`

roosterfish (approved):
Thanks!

PR #423, opened 2024-10-11
Check microcloud exit codes

masnax:
The new `capture_and_join` pattern seems to ignore the exit codes returned by the microcloud commands, so we can't check if the command actually errored without reading the output. 

Instead, this stores the exit code from each command in a `code` file, which is then checked for `0` in `capture_and_join` before unblocking.

* Also updates the tests to use `5.21/edge` to catch regressions earlier.
* Also makes sure debug binaries are updated on subsequent test runs. This slows down the test suite slightly, but is quite helpful for local debugging, as you can just rebuild your binaries and run a single test again.

### DIFF
--- a/test/includes/microcloud.sh
+++ b/test/includes/microcloud.sh
@@ -172,7 +172,7 @@ fi
     args="--session-timeout=60"
   fi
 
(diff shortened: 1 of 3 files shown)


roosterfish (approved):
Thanks for this! Looks good.

PR #428, opened 2024-10-15
Test: Don't use `--accept-certificate` when using token

roosterfish:
After https://github.com/canonical/lxd/pull/14149 got merged in LXD we should not anymore use `--accept-certificate` when using the token as the token itself already contains the fingerprint which is enough.

This currently leads to failures in the test suite, see https://github.com/canonical/microcloud/actions/runs/11324655287/job/31553675652?pr=425.

### DIFF
--- a/test/includes/microcloud.sh
+++ b/test/includes/microcloud.sh
@@ -290,19 +290,18 @@ set_debug_binaries() {
 set_remote() {
   remote="${1}"
   name="${2}"
+  client="${3}"
 
   lxc remote switch local
-
-  addr="$(lxc exec "${name}" -- lxc config get cluster.https_address)"
 
   if lxc remote list -f csv | cut -d',' -f1 | grep -qwF "${remote}" ; then
     lxc remote remove "${remote}"
   fi
 
-  token="$(lxc exec "${name}" -- lxc config trust add --name test --quiet)"
+  token="$(lxc exec "${name}" -- lxc config trust add --name "${client}" --quiet)"
 
   # Suppress the confirmation as it's noisy.
-  lxc remote add "${remote}" "https://${addr}" --token "${token}" > /dev/null 2>&1
+  lxc remote add "${remote}" "${token}" > /dev/null 2>&1
   lxc remote switch "${remote}"
 }
 
@@ -580,8 +579,8 @@ validate_system_lxd() {
 
     lxc remote switch local
 
-    # Add the peer as a remote.
-    set_remote microcloud-test "${name}"
+    # Add the peer as a remote using the name test for the trust.
+    set_remote microcloud-test "${name}" test
 
     # Ensure we are clustered and online.
     lxc cluster list -f csv | sed -e 's/,\?database-leader,\?//' | cut -d',' -f1,7 | grep -qxF "${name},ONLINE"
@@ -629,6 +628,12 @@ validate_system_lxd() {
     fi
 
     lxc remote switch local
+
+    # Remove the trust on the remote which was added when adding the remote.
+    fingerprint="$(lxc query microcloud-test:/1.0/certificates?recursion=1 | jq -r '.[] | select(.name=="test") | .fingerprint')"
+    lxc config trust remove "microcloud-test:${fingerprint}"
+
+    # Remove the remote.
     lxc remote remove microcloud-test
 
     echo "==> ${name} Validated LXD"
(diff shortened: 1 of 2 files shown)


tomponline (dismissed):

roosterfish:
Mhh I have removed the suppressing of the output locally and the `lxc remote add` command now returns `Error: Invalid certificate type`.

roosterfish:
Ok this happens because the remote still has a trust entry for this host you can see using `lxc trust ls` on the remote. But the error message seems misleading.

tomponline:
@MusicDin can we do something about this?

MusicDin:
Turns out that MicroCloud uses `latest/edge` client (lxc) and `5.21/edge` servers. 

When adding a remote, the latest client now always sends the token to the remote in order to have it invalidated, but since there is no logic for invalidating/consuming tokens when client is trusted (before `latest/edge`), the remote assumes the client is trying to add a custom certificate, and error `Invalid certificate type` comes from the attempt to parse certificate type, which is not set (empty) when adding a remote.

This will be fixed once https://github.com/canonical/lxd/pull/14149 is backported to `5.21/edge`.

Thanks @roosterfish for helping me figure this one out.

tomponline:
> Turns out that MicroCloud uses `latest/edge` client (lxc) and `5.21/edge` servers.

You mean the LXD Go SDK, not "lxc" right?

tomponline:
> I actually meant `lxc` - The latest version of `lxc` is used within the tests

How do they use a different version of `lxc` compared to `lxd`?

MusicDin:
As far as I understand, in tests `lxc` from `latest/edge` snap is used: https://github.com/canonical/microcloud/blob/main/test/includes/microcloud.sh#L1224

But MicroCloud uses `5.21/edge`: https://github.com/canonical/microcloud/blob/main/test/main.sh#L126

tomponline:
> As far as I understand, in tests `lxc` from `latest/edge` snap is used: https://github.com/canonical/microcloud/blob/main/test/includes/microcloud.sh#L1224
> 
> But MicroCloud uses `5.21/edge`: https://github.com/canonical/microcloud/blob/main/test/main.sh#L126

@roosterfish @masnax want to comment on this? Is this intentional? Ta

roosterfish:
In the GitHub runner we install LXD `latest/edge` which is then used by the test suites to spawn nested LXD VMs in which we install Micro* and LXD `5.21/edge`.

For some tests we add the nested VMs as remote to the LXD on the runner which is causing this situation of a newer client and older LXD.

tomponline:
> In the GitHub runner we install LXD `latest/edge` which is then used by the test suites to spawn nested LXD VMs in which we install Micro* and LXD `5.21/edge`.
> 
> For some tests we add the nested VMs as remote to the LXD on the runner which is causing this situation of a newer client and older LXD.

Would it be more consistent to use 5.21/edge for everything?

roosterfish:
> Would it be more consistent to use 5.21/edge for everything?
> 
> Or a matrix where its all latest/edge and all 5.21/edge to catch regressions on latest/edge sooner?

The test suites run with `5.21/edge` LXD to be as close as possible to what the customer will use (`5.21/stable`) in a real world setup.
Due to the version enforcement when forming a MicroCloud currently you cannot form a MicroCloud using LXD 6.x as that is restricted. So this would also fail in the pipeline.

I think the more edge channels we use the better to catch such situations but I am not against switching the LXD installed on the runner to `5.21/edge` too.
@masnax can you think of anything that speaks against that?

tomponline:
We've backported the change to 5.21/edge now too https://github.com/canonical/lxd/pull/14283

masnax (approved):
I'm fine with moving the main LXD to `5.21/edge` for the LTS.

test/includes/microcloud.sh
```diff
@@ -629,6 +628,12 @@ validate_system_lxd() {
     fi
 
     lxc remote switch local
+
+    # Remove the trust on the remote which was added when adding the remote.
+    fingerprint="$(lxc query microcloud-test:/1.0/certificates?recursion=1 | jq -r '.[] | select(.name=="test") | .fingerprint')"
+    lxc config trust remove "microcloud-test:${fingerprint}"
```
MusicDin, line 633: Thanks, I think this should be no longer required - after backports to 5.21. Similarly, lxc 6.1 should now work with 5.21 (if still required).

roosterfish: I was thinking it would still be good to keep it in for clean up purposes? Meaning if you add the remote again, it also reads the trust on the remote.

PR #431, opened 2024-10-17
Report if skipping local storage & ovn config

masnax:
Prints a message rather than a log when skipping local storage. I thought to avoid a question in this case because the immediate question is for remote storage, and a user may not have enough disks for both.

For OVN config, we do ask a question because FAN networks are being phased out and is not an ideal alternative to MicroOVN anyway. Now, if any systems are ineligible, we immediately ask if the user wants to abort setup. If they choose to continue, we fallback to a FAN network.

### DIFF
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -244,7 +244,8 @@ func (c *initConfig) askLocalPool(sh *service.Handler) error {
 	for _, info := range c.state {
 		hasPool, supportsPool := info.SupportsLocalPool()
 		if !supportsPool {
-			logger.Warn("Skipping local storage pool setup, some systems don't support it")
+			fmt.Println("Skipping local storage pool setup, some systems don't support it")
+
 			return nil
 		}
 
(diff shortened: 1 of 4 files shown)


cmd/microcloud/ask.go
```diff
@@ -268,13 +269,10 @@ func (c *initConfig) askLocalPool(sh *service.Handler) error {
 		}
 	}
 
-	// Local storage is already set up on every system.
-	if len(askSystems) == 0 {
-		return nil
-	}
+	// Local storage is already set up on every system, or if not every system has a disk.
+	if len(askSystems) == 0 || len(availableDisks) != len(askSystems) {
+		fmt.Println("Skipping local storage pool setup, some systems do not have an available disk")
```
roosterfish, line 274: Let's be consistent and use the same scheme as the error for distributed storage which is `No disks available for distributed storage. Skipping configuration` and use this instead:
```suggestion
		fmt.Println("No disks available for local storage. Skipping configuration")
```

roosterfish (changes requested):
Looking good, only one small remark to have consistent error messages.

roosterfish (dismissed):

gabrielmougard (dismissed):
Very good :+1:

roosterfish:
Looks like this requires some alignments to the interactive vars. Two suites fail on `Some systems are ineligible for distributed networking, which requires either an interface with no IPs assigned or a bridge. Continue anyway?`

roosterfish (approved):
Looks good!

Issue #427, opened 2024-10-14
MicroCloud init hangs when Ceph encryption is selected on Ubuntu 22.04


mseralessandri:
`microcloud init` hangs on `Configuring cluster-wide devices ...` when Ceph encryption is selected on Ubuntu 22.04. 
Encryption requires a kernel with `dm_crypt` enabled which is not available on Ubuntu 22.04. The required steps are detailed in the [MicroCloud doc](https://canonical-microcloud.readthedocs-hosted.com/en/latest/microcloud/how-to/initialise/).

However `microcloud init` should not hang and should exit with an error.

OS Version: `Ubuntu 22.04.4 LTS (GNU/Linux 5.15.0-119-generic x86_64)`
Snap versions:
```
:~$ sudo snap list
Name        Version                    Rev    Tracking       Publisher   Notes
core22      20240904                   1621   latest/stable  canonical✓  base
core24      20240710                   490    latest/stable  canonical✓  base
go          1.23.2                     10730  latest/stable  canonical✓  classic
lxd         git-5d7521a                30585  5.21/edge      canonical✓  in-cohort
microceph   19.2.0~git+snapc76c1f5fe9  1189   latest/edge    canonical✓  in-cohort
microcloud  git-6d5f38f                1007   latest/edge    canonical✓  in-cohort
microovn    24.03.2+snapb5ffb86eb5     602    latest/edge    canonical✓  in-cohort
snapd       2.63                       21759  latest/stable  canonical✓  snapd
```

How to reproduce: I tried with a single node MicroCloud configuration selecting `yes` to `Do you want to encrypt the selected disks? (yes/no) [default=no]:` 

```
:~$ sudo microcloud init
Waiting for LXD to start ...
Do you want to set up more than one cluster member? (yes/no) [default=yes]: no
Select an address for MicroCloud's internal traffic:

 Using address "10.142.21.93" for MicroCloud

Gathering system information ...
Would you like to set up local storage? (yes/no) [default=yes]: yes
Select exactly one disk from each cluster member:

Select which disks to wipe:

 Using "/dev/disk/by-id/virtio-9c6b8b32-0573-4f2e-a" on "maria01" for local storage pool

Would you like to set up distributed storage? (yes/no) [default=yes]: yes
Select from the available unpartitioned disks:

Select which disks to wipe:

Disk configuration does not meet recommendations for fault tolerance. At least 3 systems must supply disks.
[...]

PR #433, opened 2024-10-17
Check MicroCeph for API errors

masnax:
Potential solution for #427 

Seems this isn't an issue with 22.04 necessarily, but with how we are using the microceph client. For the [`AddDisk`](https://github.com/canonical/microceph/blob/e32ad499ad83dcf6401891cc5739524bcdcd6dd4/microceph/client/disks.go#L18-L29) function, API side errors are compiled in `types.DiskAddResponse` instead of the error returned by the client function. 

As such, we were ignoring most disk-add errors. The reason this led to hangs is because we would proceed to create the LXD storage pools, which would hang forever on Ceph trying to create OSD pools without any OSDs. 

In particular, if encryption is requested, MicroCeph requires the following to be run before running `microcloud init`:
```bash
sudo snap connect microceph:dm-crypt
sudo snap restart microceph.daemon
```

We link to this in the MicroCeph docs in our initialization steps [here](canonical-microcloud.readthedocs-hosted.com/en/latest/microcloud/how-to/initialise/#interactive-configuration) where it mentions encryption.

In my testing, if the above is run before following the steps in the issue, then MicroCloud completes setup successfully. Otherwise, it hangs.

### DIFF
--- a/cmd/microcloud/main_init.go
+++ b/cmd/microcloud/main_init.go
@@ -763,9 +763,25 @@ func (c *initConfig) setupCluster(s *service.Handler) error {
 				}
 
 				logger.Debug("Adding disk to MicroCeph", logger.Ctx{"name": name, "disk": disk.Path})
-				_, err = cephClient.AddDisk(context.Background(), client, &disk)
+				resp, err := cephClient.AddDisk(context.Background(), client, &disk)
 				if err != nil {
 					return err
+				}
+
+				var diskErr string
+				for _, report := range resp.Reports {
+					if report.Error != "" {
+						if diskErr == "" {
+							diskErr = report.Error
+						} else {
+							// Populate errors backwards, as the latest error is at the end of the list.
+							diskErr = fmt.Sprintf("%s: %s", report.Error, diskErr)
+						}
+					}
+				}
+
+				if diskErr != "" {
+					return fmt.Errorf("Failed to add disk to MicroCeph: %s", diskErr)
 				}
 			}
 		}


cmd/microcloud/main_init.go
```diff
@@ -763,10 +763,16 @@ func (c *initConfig) setupCluster(s *service.Handler) error {
...
 				logger.Debug("Adding disk to MicroCeph", logger.Ctx{"name": name, "disk": disk.Path})
-				_, err = cephClient.AddDisk(context.Background(), client, &disk)
+				resp, err := cephClient.AddDisk(context.Background(), client, &disk)
 				if err != nil {
 					return err
 				}
+
+				for _, report := range resp.Reports {
+					if report.Error != "" {
+						return fmt.Errorf("Failed to add disk to MicroCeph: %s", report.Error)
```
roosterfish, line 773: Do we have to concatenate the errors from multiple `Reports`?

roosterfish (approved):
Good catch!

roosterfish:
> In particular, if encryption is requested, MicroCeph requires the following to be run before running

Can it be that this is auto connected sometimes? Because I remember doing some MicroCloud deployments in the past with Ceph encryption and I have never manually did the connection with `dm-crypt`.

tomponline:
> As such, we were ignoring most disk-add errors.

Ah good that was my hunch, so glad we dug into it to find this. Thanks

roosterfish:
> Can it be that this is auto connected sometimes?

Nvm the test suite is doing this that's why it works https://github.com/canonical/microcloud/blob/main/test/includes/microcloud.sh#L1215.

gabrielmougard (approved):

cmd/microcloud/main_init.go
masnax: Fixed.

roosterfish (approved):

PR #435, opened 2024-10-17
Discovery: Have a dedicated `StopResponder` to not rely on the context

roosterfish:
In one run of the unit tests on GH [an error surfaced](https://github.com/canonical/microcloud/actions/runs/11327243960/job/31497901025?pr=426) that has shown that subsequent starts of the responder might collide which each other if the connection hasn't yet been cleaned up successfully when cancelling the context.

By having a dedicated `StopResponder` func we can make sure that the underlying connection is closed before starting a new one.
As before with mDNS we now track the server in the session so we can close it when the session gets closed.

### DIFF
--- a/multicast/discovery.go
+++ b/multicast/discovery.go
@@ -26,9 +26,11 @@ type ServerInfo struct {
 
 // Discovery represents the information used for discovering peers using multicast.
 type Discovery struct {
-	iface string
-	port  int64
-	group net.IP
+	iface           string
+	port            int64
+	group           net.IP
+	responderConn   *ipv4.PacketConn
+	responderCancel context.CancelFunc
 }
 
 // NewDiscovery returns a new instance of Discovery which allows to lookup peers
@@ -57,13 +59,14 @@ func (d *Discovery) Respond(ctx context.Context, info ServerInfo) error {
 		return fmt.Errorf("Failed to listen on %d: %w", d.port, err)
 	}
 
-	receiverP := ipv4.NewPacketConn(receiver)
-	err = receiverP.JoinGroup(iface, &net.UDPAddr{IP: d.group})
+	ctx, d.responderCancel = context.WithCancel(ctx)
+	d.responderConn = ipv4.NewPacketConn(receiver)
+	err = d.responderConn.JoinGroup(iface, &net.UDPAddr{IP: d.group})
 	if err != nil {
 		return fmt.Errorf("Failed to join multicast group %q: %w", d.group.String(), err)
 	}
 
-	err = receiverP.SetControlMessage(ipv4.FlagDst, true)
+	err = d.responderConn.SetControlMessage(ipv4.FlagDst, true)
 	if err != nil {
 		return fmt.Errorf("Failed to set IPv4 control flag for destination address: %w", err)
 	}
@@ -72,7 +75,7 @@ func (d *Discovery) Respond(ctx context.Context, info ServerInfo) error {
 	// This allows existing the endpoint's blocking read using ReadFrom.
 	go func() {
 		<-ctx.Done()
-		err := receiverP.Close()
+		err := d.responderConn.Close()
 		if err != nil {
 			logger.Error("Failed to close network endpoint after context got cancelled", logger.Ctx{"err": err})
 		}
@@ -84,7 +87,7 @@ func (d *Discovery) Respond(ctx context.Context, info ServerInfo) error {
 		for {
 			// See the comment on the sender (lookup) for the reasoning about using 500.
 			b := make([]byte, 500)
-			n, cm, src, err := receiverP.ReadFrom(b)
+			n, cm, src, err := d.responderConn.ReadFrom(b)
 			if err != nil {
 				// Ignore "use of closed network connection" errors as this happens normally
 				// if the outer context gets cancelled in the connection closer go routine.
@@ -119,7 +122,7 @@ func (d *Discovery) Respond(ctx context.Context, info ServerInfo) error {
 					}
 
 					// Send a unicast message back to the source.
-					_, err = receiverP.WriteTo(bytes, nil, src)
+					_, err = d.responderConn.WriteTo(bytes, nil, src)
 					if err != nil {
 						logger.Error("Failed to send reply", logger.Ctx{"dest": src.String(), "err": err})
 						continue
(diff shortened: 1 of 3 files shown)


multicast/discovery.go
```diff
@@ -57,13 +59,14 @@ func (d *Discovery) Respond(ctx context.Context, info ServerInfo) error {
 		return fmt.Errorf("Failed to listen on %d: %w", d.port, err)
 	}
 
-	receiverP := ipv4.NewPacketConn(receiver)
-	err = receiverP.JoinGroup(iface, &net.UDPAddr{IP: d.group})
+	ctx, d.responderCancel = context.WithCancel(ctx)
```
gabrielmougard, line 62: Overwriting the ctx can lead to confusion and potential bugs, especially if the original context carries important deadlines, cancellations, or values that are relied upon elsewhere in the code (maybe there are none and then this is fine). Any cancellation or deadline on the original ctx will still propagate, but overwriting it here means that any modifications to ctx after this point will not affect the original context. Instead of overwriting ctx, we could introduce a new variable for the child context:
```go
responderCtx, responderCancel := context.WithCancel(ctx)
d.responderCancel = responderCancel
```

Then, use `responderCtx` within the goroutine down below. This approach maintains the integrity of the original ctx and also makes it easier to know which context is being used.

multicast/discovery_test.go
```diff
@@ -48,15 +93,31 @@ func (m *multicastSuite) Test_Lookup() {
...
-		m.Require().Equal(&c.responseInfo, receivedInfo)
+		if c.modifier != nil {
+			c.modifier(discovery)
+		}
+
+		testDiscovery := NewDiscovery(c.lookupIface, c.lookupPort)
+
+		ctx := context.Background()
+		if c.lookupTimeout > 0 {
+			ctx, _ = context.WithTimeoutCause(ctx, c.lookupTimeout, fmt.Errorf("Timeout exceeded"))
```
gabrielmougard, line 107: I don't know this new `context.WithTimeoutCause` function :) What is it exactly used for here? Why not using `context.WithTimeout` ?

multicast/discovery_test.go
```diff
@@ -34,12 +43,48 @@ func (m *multicastSuite) Test_Lookup() {
...
+			desc:          "Cannot lookup system if the responder is offline",
+			lookupVersion: "2.0",
+			lookupIface:   "lo",
+			lookupPort:    9444,
+			responseInfo: ServerInfo{
+				Version: "2.0",
+				Name:    "foo",
+				Address: "1.2.3.4",
+			},
+			lookupTimeout: 500 * time.Microsecond,
```
gabrielmougard, line 69: Timeouts of 500 microseconds seem extremely short... I'm a bit worried about ir leading to flaky tests due to variability in system scheduling, garbage collection pauses (or other factors in the CI workers) . What do you think? Is `lookupTimeout: 500 * time.Millisecond` safer for example?

multicast/discovery.go
```diff
@@ -134,6 +137,25 @@ func (d *Discovery) Respond(ctx context.Context, info ServerInfo) error {
 	return nil
 }
 
+// StopResponder stops the responder server and cancels it's inner context.
+func (d *Discovery) StopResponder() error {
+	// Check if this instance of discovery has an active responder server connection.
+	if d.responderConn != nil {
```
gabrielmougard, line 143: Calling `StopResponder` without a prior call to `Respond` may result in a nil `d.responderCancel`, which would cause a panic if `d.responderCancel()` is called. To avoid that, can we add a nil check for `d.responderCancel` dow below before calling it:
```go
if d.responderCancel != nil {
    d.responderCancel()
}
```

multicast/discovery_test.go
roosterfish: It helps to to reason about why a context got cancelled and allows you to transport the reason for cancellation to any recipient who is watching the context.

Take this as an example:

```go
ctx := context.WithTimeoutCause(ctx, 10*time.Second, fmt.Errorf("A more descriptive timeout error than the standard one")

// do something

go func() {
     <-ctx.Done()
     // Now the context is done but what has happened?
     // Was the parent context cancelled or our child one?
     err := context.Cause(ctx)
     fmt.Println(err.Error()) // Returns "A more descriptive timeout error than ..."
}()

```

So you can both override the default "context exceeded" error (e.g. there is also `WithCancelCause` that returns a canceller that allows you to put in an arbitrary `error` when cancelling) but you can also use it to react on certain errors with `errors.Is()` to see if your child context got cancelled with a specific error or if the parent context got cancelled with another error.

roosterfish: In this PR its used to overwrite the default error.

multicast/discovery.go
```diff
@@ -134,6 +137,25 @@ func (d *Discovery) Respond(ctx context.Context, info ServerInfo) error {
...
+		err := d.responderConn.Close()
+		// Ignore errors if the connection is already closed.
+		// This can happen if the responders context already got cancelled
+		// which also triggers a close of the connection.
+		if err != nil && !errors.Is(err, net.ErrClosed) {
```
gabrielmougard, line 148: `net.ErrClosed` is the sentinel error for closed network connections, but depending on how the error is wrapped, is there a chance that `errors.Is` may not correctly identify it and lead to unexpected error handling?

gabrielmougard (commented):
just a couple of questions :)

multicast/discovery.go
roosterfish: Isn't this the go-to pattern for deriving new contexts if you don't care about the parent context? See also the docs of [`WithTimeout`](https://pkg.go.dev/context#WithTimeout).

In this case a new context is derived from the outer context and used internally so we can cancel it ourselves (as we don't control the outer context) any time. And if the parent context gets cancelled also our derived child context exits.

multicast/discovery_test.go
roosterfish: Ah yeah actually my intention was to use `Millisecond` but it shouldn't really matter. As there is no server running (see the line below, it's getting stopped beforehand), we could also have a timeout of 0. So it's not like a sleep or something. 

In general I have tried to use a very short amount of time to not much influence the runtime of the test.

multicast/discovery.go
roosterfish: If we call `StopResponder` there can be two situations:

1) The responder is still running (connection still alive and running go routines)
2) The responder is already stopped (connection closed and context cancelled)

Now if we try to close the connection but it is already closed due to the context being cancelled before (we cannot control this as its the outer context), it would be annoying to see an error from closing the connection. Therefore this one is being filtered out. 

If it's wrapped differently we will return it as there is clearly something else going on and we would like to see it.

multicast/discovery.go
roosterfish: Fixed

multicast/discovery_test.go
gabrielmougard: Got it :) Thanks!

multicast/discovery.go
gabrielmougard: Yes true. Nevermind then :)

gabrielmougard (approved):
LGTM !

PR #436, opened 2024-10-18
Upgrade: Add docs and tests

roosterfish:
This PR adds MicroCloud upgrade tests from `1/stable` to `latest/edge` as well as for the MicroCeph, MicroOVN and LXD dependencies.

As there might be MicroClouds running on 22.04, the test suite is modified to allow specifying a `BASE_OS` that will be used when setting up the testbed. 
This allows running the new `upgrade` test suite for both 22.04 and 24.04.

To verify the upgrade we check the cluster status of MicroCloud and its dependencies after the upgrade. In addition some workload is deployed before the upgrade whose boot ID and network connectivity is checked afterwards.

Furthermore dedicated docs are now added for update and upgrade procedures.

### DIFF
--- a/doc/how-to/support.md
+++ b/doc/how-to/support.md
@@ -3,13 +3,14 @@
 
 We recommend using the following channels for the snaps required to run MicroCloud:
 
-* For MicroCloud: `latest/[stable|candidate|edge]`
-* For LXD: `5.21/[stable|candidate|edge]`
-* For MicroCeph: `quincy/[stable|candidate|edge]`
-* For MicroOVN: `22.03/[stable|candidate|edge]`
+* For MicroCloud: `2/(stable|candidate|edge)`
+* For LXD: `5.21/(stable|candidate|edge)`
+* For MicroCeph: `squid/(stable|candidate|edge)`
+* For MicroOVN: `24.03/(stable|candidate|edge)`
 
 ```{note}
-Currently, there is no LTS version of MicroCloud.
+The LTS version of MicroCloud is available in the `2` track.
+It's recommended to use the `<track>/stable` channels for production deployments.
 ```
 
 ## Support and community
--- a/doc/how-to/update_upgrade.md
+++ b/doc/how-to/update_upgrade.md
@@ -1,0 +1,141 @@
+(howto-update-upgrade)=
+# How to update and upgrade
+
+MicroCloud is made of several snaps that are closely coupled with each other.
+The cluster members that are part of the MicroCloud deployment must always run the same version of the snaps.
+This means that when the snaps on one of the cluster members are refreshed, they must also be refreshed on all other cluster members before MicroCloud is operational again.
+See {ref}`howto-update-upgrade-deps` for the recommended order.
+
+Before performing an update or upgrade make sure to backup your data to prevent any data loss in case of failure.
+See the following backup guides for each of the snaps:
+
+* {doc}`How to backup MicroCeph <microceph:explanation/taking-snapshots>`
+* {ref}`How to backup LXD <lxd:backups>`
+
+In case of error see {ref}`howto-recover` for troubleshooting details.
+
+(howto-update-upgrade-update)=
+## Update MicroCloud
+
+Updating MicroCloud allows getting the latest set of features and fixes in the tracked channels for the various snaps.
+Performing an update requires going through the list of snaps one after another and update each of the individual cluster members.
+
+```{note}
+Depending on which snap gets updated, some services of this snap (e.g. API endpoints) might not be available whilst performing the update.
+Check the respective services documentation on updates for more information.
+```
+
+During an update the snaps channel won't be modified so the snaps get updated to the last available version inside of the current channel.
+This does not introduce breaking changes and can be used on a regular basis to update the MicroCloud.
+
+As MicroCloud consumes the services offered by the dependant snaps (MicroCeph, MicroOVN and LXD), the update procedure starts by updating
+the list of dependencies first.
+
+(howto-update-upgrade-deps)=
+### Update dependency snaps
+
+Updating the dependencies can be done by running `snap refresh` against the respective snap.
+For MicroCloud automatic snap refreshes are put on hold. See {ref}`howto-snap-control-updates` for more information.
+
+To start the update procedure, enter the following command on the first machine:
+
+    sudo snap refresh microceph --cohort="+"
+
+If the command has succeeded, run the same command on the next machine and so on.
+
+```{note}
+Make sure to validate the health of the recently updated dependency before continuing with the next one.
+```
+
+After successfully updating MicroCeph continue with MicroOVN.
+Again enter the following command on the first machine:
+
+    sudo snap refresh microovn --cohort="+"
+
+As before run the command on all the other machines one after another if you don't observe any errors.
+Next we can continue to update LXD.
+
+The refresh will block until each of the LXD cluster members is updated so make sure to perform the following
+command on all machines in parallel:
+
+    sudo snap refresh lxd --cohort="+"
+
+### Update MicroCloud snap
+
+Last but not least we can update MicroCloud.
+As before enter the following command on the first machine:
+
+    sudo snap refresh microcloud --cohort="+"
+
+Continue running the command on the rest of the machines to finish the update.
+You can confirm a healthy state of the MicroCloud after the update by running the following command:
+
+    sudo microcloud status
+
+```{note}
+The status command was introduced in MicroCloud version 2.
+See {ref}`howto-update-upgrade-upgrade` on how to upgrade to another track.
+```
+
+(howto-update-upgrade-upgrade)=
+## Upgrade MicroCloud
+
+Upgrading MicroCloud allows switching to another track with major improvements and enhanced functionality.
+Performing an upgrade requires going through the list of snaps one after another and upgrade each of the individual cluster members.
+
+```{note}
+Depending on which snap gets upgraded, some services of this snap (e.g. API endpoints) might not be available whilst performing the upgrade.
+Check the respective services documentation on upgrades for more information.
+```
+
+During an upgrade the snaps channel will be switched to another track.
+This might introduce breaking changes for MicroCloud and its dependencies and should be done with care.
+See {ref}`howto-update-upgrade-update` for regular non-breaking updates.
+
+As MicroCloud consumes the services offered by the dependant snaps (MicroCeph, MicroOVN and LXD), the update procedure starts by updating
+the list of dependencies first.
+
+### Upgrade dependency snaps
+
+Upgrading the dependencies can be done by running `snap refresh --channel <new track/stable>` against the respective snap.
+
+Make sure to consult the dedicated upgrade guides of each dependency before you perform the actual upgrade:
+
+* {doc}`How to upgrade MicroCeph <microceph:how-to/reef-upgrade>`
+* {doc}`How to upgrade MicroOVN <microovn:how-to/major-upgrades>`
+* {doc}`How to upgrade LXD <lxd:howto/cluster_manage>`
+
+To start the upgrade procedure, enter the following command on the first machine:
+
+    sudo snap refresh microceph --channel "squid/stable" --cohort="+"
+
+If the command has succeeded, run the same command on the next machine and so on.
+
+```{note}
+Make sure to validate the health of the recently upgraded dependency before continuing with the next one.
+```
+
+After successfully upgrading MicroCeph continue with MicroOVN.
+Again enter the following command on the first machine:
+
+    sudo snap refresh microovn --channel "24.03/stable" --cohort="+"
+
+As before run the command on all the other machines one after another if you don't observe any errors.
+Next we can continue to upgrade LXD.
+
+The installer will block until each of the LXD cluster members is upgraded so make sure to perform the following
+command on all machines in parallel:
+
+    sudo snap refresh lxd --channel "5.21/stable" --cohort="+"
+
+### Upgrade MicroCloud snap
+
+Last but not least we can upgrade MicroCloud.
+As before enter the following command on the first machine:
+
+    sudo snap refresh microcloud --channel "2/stable" --cohort="+"
+
+Continue running the command on the rest of the machines to finish the upgrade.
+You can confirm a healthy state of the MicroCloud after the upgrade by running the following command:
+
+    sudo microcloud status
--- a/test/includes/microcloud.sh
+++ b/test/includes/microcloud.sh
@@ -586,7 +586,15 @@ validate_system_lxd() {
     [ "$(lxc cluster list -f csv | wc -l)" = "${num_peers}" ]
 
     # Check core config options
-    [ "$(lxc config get core.https_address)" = "[::]:8443" ]
+    lxd_address="$(lxc config get core.https_address)"
+    if [ "${MICROCLOUD_SNAP_CHANNEL}" = "1/stable" ]; then
+      # There was a bug in MicroCloud 1 that set different addresses.
+      # See https://github.com/canonical/microcloud/issues/214
+      system_address="$(lxc ls local:"${name}" -f json -c4 | jq -r '.[0].state.network.enp5s0.addresses[] | select(.family == "inet") | .address')"
+      [ "${lxd_address}" = "${system_address}:8443" ] || [ "${lxd_address}" = "[::]:8443" ]
+    else
+      [ "${lxd_address}" = "[::]:8443" ]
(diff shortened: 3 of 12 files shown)


test/suites/upgrade.sh
```diff
@@ -0,0 +1,225 @@
...
+    # Deactivate the fourth MicroCloud so we can add it later after upgrade.
+    lxc exec micro04 -- snap stop microcloud
+
+    # Deploy a version 1 MicroCloud and launch some instances.
+    lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud init --preseed <<< "$preseed"
+
+    if [ "${SKIP_VM_LAUNCH}" = "1" ]; then
+      echo "::warning::SKIPPING VM LAUNCH TEST"
+    else
+      lxc exec micro01 -- lxc launch ubuntu:noble v1 --vm
```
simondeziel, line 66: Should we use the same minimal-daily image we use elsewhere? Possibly with `security.devlxd.images=true`?

test/suites/upgrade.sh
```diff
@@ -0,0 +1,225 @@
...
+
+      v1_uptime="$(lxc exec micro01 -- lxc exec v1 -- uptime -s)"
```
simondeziel, line 68: TIL about `uptime -s` which is nice, thanks!

test/suites/upgrade.sh
```diff
@@ -0,0 +1,225 @@
...
+    # Perform some pings to check the network.
+    gateway="$(lxc network get lxdbr0 ipv4.address | awk -F'/' '{print $1}')"
+    if [ "${SKIP_VM_LAUNCH}" = "1" ]; then
+      echo "::warning::SKIPPING VM TESTS"
+    else
+      lxc exec micro01 -- lxc exec v1 -- ping -c 2 "${gateway}"
+
+      # Compare the uptimes to verify the instances haven't been restarted.
+      v1_uptime_after_upgrade="$(lxc exec micro01 -- lxc exec v1 -- uptime -s)"
+      [ "${v1_uptime}" = "${v1_uptime_after_upgrade}" ]
```
simondeziel, line 182: This `_uptime` variable suffix feels misleading. An alternative would be to compare the `boot_id` which is guarantied to be different between boots. That's what [`lxd-ci` uses](https://github.com/canonical/lxd-ci/blob/main/tests/storage-vm#L686-L697).

test/suites/upgrade.sh
```diff
@@ -0,0 +1,225 @@
...
+
+    for m in micro01 micro02 micro03; do
+      retries=0
+      while true; do
+        if [ "${retries}" -gt 60 ]; then
+          echo "MicroCeph member ${m} failed to come up after upgrade"
+          exit 1
+        fi
+
+        if lxc exec "${m}" -- sh -c "microceph cluster list | grep ${m} | grep -q ONLINE"; then
```
simondeziel, line 86: The `sh -c` is potentially not needed.

test/suites/upgrade.sh
```diff
@@ -0,0 +1,225 @@
...
+
+    for m in micro01 micro02 micro03; do
+      retries=0
+      while true; do
+        if [ "${retries}" -gt 60 ]; then
+          echo "MicroOVN member ${m} failed to come up after upgrade"
+          exit 1
+        fi
+
+        if lxc exec "${m}" -- sh -c "microovn cluster list -f json | jq -r '.[] | select(.name == \"${m}\") | .status' | grep -q ONLINE"; then
```
simondeziel, line 111: Same re `sh -c`

test/suites/upgrade.sh
```diff
@@ -0,0 +1,225 @@
...
+
+    for m in micro01 micro02 micro03; do
+      retries=0
+      while true; do
+        if [ "${retries}" -gt 60 ]; then
+          echo "LXD member ${m} failed to come up after upgrade"
+          exit 1
+        fi
+
+        if lxc exec "${m}" -- sh -c "lxc cluster list -f json | jq -r '.[] | select(.server_name == \"${m}\") | .status' | grep -q Online"; then
```
simondeziel, line 138: Same re `sh -c`.

test/suites/upgrade.sh
```diff
@@ -0,0 +1,225 @@
...
+
+    for m in micro01 micro02 micro03; do
+      retries=0
+      while true; do
+        if [ "${retries}" -gt 60 ]; then
+          echo "LXD member ${m} failed to come up after upgrade"
+          exit 1
+        fi
+
+        if lxc exec "${m}" --env TEST_CONSOLE=0 -- sh -c "microcloud cluster list -f json | jq -r '.[] | select(.name == \"${m}\") | .status' | grep -q ONLINE"; then
```
simondeziel, line 163: Same re `sh -c`.

test/suites/upgrade.sh
```diff
@@ -0,0 +1,225 @@
...
+    # Upgrade micro04.
+    lxc exec micro04 -- snap refresh microceph --channel "${microceph_target}"
+    lxc exec micro04 -- snap refresh microovn --channel "${microovn_target}"
+    lxc exec micro04 -- snap refresh lxd --channel "${lxd_target}"
+    lxc exec micro04 -- snap refresh microcloud --channel "${microcloud_target}"
+    lxc exec micro04 -- snap start microcloud
+
+    # Join micro04 to the old cluster using the MicroCloud 2 preseed format.
+    lookup_gateway=$(lxc network get lxdbr0 ipv4.address)
+    preseed="$(cat << EOF
```
simondeziel, line 197: Same re subshell+cat.

test/suites/upgrade.sh
```diff
@@ -0,0 +1,225 @@
...
+      done
+    done
+
+    # Check if the workload survived the upgrade.
+    # Perform some pings to check the network.
+    gateway="$(lxc network get lxdbr0 ipv4.address | awk -F'/' '{print $1}')"
+    if [ "${SKIP_VM_LAUNCH}" = "1" ]; then
+      echo "::warning::SKIPPING VM TESTS"
+    else
+      lxc exec micro01 -- lxc exec v1 -- ping -c 2 "${gateway}"
```
simondeziel, line 178: `ping -nc2` should avoid doing reverse DNS lookup.

test/suites/upgrade.sh
```diff
@@ -0,0 +1,225 @@
...
+    lxc exec micro04 --env TEST_CONSOLE=0 -- sh -c 'microcloud preseed > out' <<< "$preseed" &
+    lxc exec micro01 --env TEST_CONSOLE=0 -- sh -c 'microcloud preseed > out' <<< "$preseed"
+
+    lxc exec micro01 -- tail -1 out | grep "MicroCloud is ready" -q
+    lxc exec micro04 -- tail -2 out | head -1 | grep "Successfully joined the MicroCloud cluster and closing the session" -q
+
+    validate_system_microceph micro04 0 0 disk2
+    validate_system_microovn micro04
+    validate_system_lxd micro04 4 disk1 1 0 enp6s0 10.1.123.1/24 10.1.123.100-10.1.123.254 fd42:1:1234:1234::1/64
+    lxc exec micro01 --env TEST_CONSOLE=0 -- sh -c "microcloud cluster list -f json | jq -r '.[] | select(.name == \"micro04\") | .status' | grep -q ONLINE"
```
simondeziel, line 223: Same re `sh -c`.

test/suites/upgrade.sh
```diff
@@ -0,0 +1,225 @@
+#!/bin/bash
+
+test_upgrade() {
+  reset_systems 4 2 3
+
+  # Perform upgrade test from MicroCloud 1 to 2.
+  if [ "${MICROCLOUD_SNAP_CHANNEL}" = "1/stable" ]; then
+    microceph_target="squid/edge" # TODO: squid/stable when available
+    microovn_target="latest/edge" # TODO: 24.03/stable when available
+    lxd_target="5.21/edge"
```
simondeziel, line 10: Since we are always testing with `5.21/edge` would you mind updating the check in `test/includes/check.sh` while at it, please?

doc/how-to/update_upgrade.md
```diff
@@ -0,0 +1,135 @@
+(howto-update-upgrade)=
+# How to update and upgrade
+
+MicroCloud is made of several snaps that are closely coupled with each other.
+The cluster members that are part of the MicroCloud deployment must always run the same version of the snaps.
+This means that when the snaps on one of the cluster members are refreshed, they must also be refreshed on all other cluster members before MicroCloud is operational again.
+See {ref}`howto-update-upgrade-deps` for the recommended order.
+
+Before performing an update or upgrade make sure to backup your data to prevent any data loss in case of failure.
```
simondeziel, line 9: Should we link to the backup section of `doc/how-to/recover.md`?

doc/how-to/update_upgrade.md
```diff
@@ -0,0 +1,135 @@
...
+
+After successfully updating MicroCeph continue with MicroOVN.
+Again enter the following command on the first machine:
+
+    sudo snap refresh microovn --cohort="+"
+
+As before run the command on all the other machines one after another if you don't observe any errors.
+Next we can continue to update LXD.
+
+The installer will block until each of the LXD cluster members is updated so make sure to perform the following
```
simondeziel, line 52: `installer`?

test/suites/upgrade.sh
```diff
@@ -0,0 +1,225 @@
...
+    microcloud_target="latest/edge" # TODO: 2/stable when available
+
+    # The lookup subnet has to contain the netmask and the address has to be the one used by MicroCloud not the gateway.
+    lookup_subnet="$(lxc ls micro01 -f json -c4 | jq -r '.[0].state.network.enp5s0.addresses[] | select(.family == "inet") | .address + "/" + .netmask')"
+
+    # Use the MicroCloud version 1 preseed format with all the features available at the time of release:
+    # - local and remote storage
+    # - wiping disks
+    # - OVN uplink interface
+    preseed="$(cat << EOF
```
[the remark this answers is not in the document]
roosterfish: That's a good question. 
I am very much used to this pattern but agree that the latter is way simpler. Will switch it.

test/suites/upgrade.sh
roosterfish: Thanks, wasn't aware of the boot id. I'll switch to that instead.

test/suites/upgrade.sh
roosterfish: Sure, thanks for the reminder.

doc/how-to/update_upgrade.md
roosterfish: Ups, switched to `refresh`.

roosterfish:
@masnax regarding your comment from yesterday whether or not we should `--hold` the snaps. 
Even if we don't change the schema, we could still introduce new API extensions as those don't break existing functionality.
So if we hold the snaps we can at least make sure that if there is a new API extension it's getting rolled out to all the members when running an update.

test/suites/upgrade.sh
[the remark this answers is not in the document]
roosterfish: Fixed

test/suites/upgrade.sh
roosterfish: Fixed

test/suites/upgrade.sh
roosterfish: Fixed

test/suites/upgrade.sh
roosterfish: Fixed

test/suites/upgrade.sh
roosterfish: Fixed

test/suites/upgrade.sh
roosterfish: Fixed

test/suites/upgrade.sh
roosterfish: Fixed

test/suites/upgrade.sh
roosterfish: Fixed

doc/how-to/update_upgrade.md
roosterfish: Fixed. Also updated `doc/how-to/support.md` which also required an update.

test/suites/upgrade.sh
simondeziel: Even using the `-daily` image isn't a silver bullet against needing to run `apt-get update`, unfortunately.

doc/how-to/support.md
```diff
@@ -3,13 +3,14 @@
 
 We recommend using the following channels for the snaps required to run MicroCloud:
 
-* For MicroCloud: `latest/[stable|candidate|edge]`
+* For MicroCloud: `2/[stable|candidate|edge]`
 * For LXD: `5.21/[stable|candidate|edge]`
-* For MicroCeph: `quincy/[stable|candidate|edge]`
-* For MicroOVN: `22.03/[stable|candidate|edge]`
+* For MicroCeph: `squid/[stable|candidate|edge]`
+* For MicroOVN: `24.03/[stable|candidate|edge]`
```
simondeziel, line 9: I'm used to seeing `[]` for optional bits but in this case we mean an alternation where only one of `stable`, `candidate` or `edge` is picked. For such notation, `24.03/(stable|candidate|edge)` feels more natural... but maybe I've seen too many regexps ;)

doc/how-to/update_upgrade.md
simondeziel: The link to the `howto-recover` is for troubleshooting details but I think we should link to the specific part that talks about backups in the very sentence we recommend doing backups.

Having a link to the whole recover page for generic troubleshooting details makes sense though.

simondeziel (approved):

test/suites/upgrade.sh
roosterfish: I am not against using the minimal daily instead. I would not duplicate the image on the runner to prevent confusion later on when you expect to get a `ubuntu-minimal` but instead you get the `-daily`.

simondeziel: I was just suggesting a possible cache seed. This wouldn't have any visible side effect as the requested image would be the same. The only difference would be that child VMs would download it much faster if it's coming from their parent's cache rather than from the network.

`security.devlxd.images=true` just tells LXD to check over the devlxd communication channel if the precise fingerprint it wants is downloadable over that fast devlxd connection or not. If yes, it downloads that image from there, if not, it uses the slower network to get what it needs.

roosterfish: Yeah I did some tests before with `security.devlxd.images=true` but probably I did a mistake elsewhere as it didn't download the image from the host.

I'll shortly push another change that uses the `ubuntu-minimal-daily` also for the workload tests. In addition the equivalent container image gets pre fetched too so that the nested LXDs can consume it.

Thanks for the hint, wasn't aware of the feature till now :)

doc/how-to/support.md
roosterfish: Agree, the `()` look more familiar. Fixed

roosterfish:
> The link to the howto-recover is for troubleshooting details but I think we should link to the specific part that talks about backups in the very sentence we recommend doing backups.
>
> Having a link to the whole recover page for generic troubleshooting details makes sense though.

Fixed.

test/suites/upgrade.sh
roosterfish: Fixed.

As `$BASE_OS` can be different for the `micro*` VMs, there is now a step to pre fetch Ubuntu 22.04 and 24.04 images for both VMs and containers so we always have them at hand.

test/includes/microcloud.sh
```diff
@@ -1146,7 +1156,13 @@ create_system() {
...
     fi
 
-    lxc init ubuntu-minimal-daily:24.04 "${name}" --vm -c limits.cpu=4 -c limits.memory=4GiB
+    lxc init "ubuntu-minimal-daily:${os}" "${name}" --vm -c limits.cpu=4 -c limits.memory=4GiB -c security.devlxd.images=true
+
+    # Pre fetch additional images to be used by the VM through security.devlxd.images=true
+    lxc image copy ubuntu-minimal-daily:24.04 local:
+    lxc image copy ubuntu-minimal-daily:22.04 local:
+    lxc image copy ubuntu-minimal-daily:24.04 local: --vm
+    lxc image copy ubuntu-minimal-daily:22.04 local: --vm
```
simondeziel, line 1165: LGTM but I'd probably do the pre-fetching before the first `lxc init`.

roosterfish: Agree, fixed.

simondeziel (approved):
Thanks!

Issue #410, opened 2024-10-02
docs: add LXD purge step to installation instructions


token47:
In the page https://canonical-microcloud.readthedocs-hosted.com/en/latest/microcloud/how-to/install/ there are installation instructions.

Turns out it is very very easy to fall into some traps while doing it:

- you may just paste the install command and it will not work because ubuntu comes with lxd 5.0 already installed, it will just say it is installed and do nothing. But you do need to refresh it to version 5.21 otherwise you get an error later during init. [1]
- you may realize that and do a refresh instead, upgrading LXD to 5.21 and you still get an error, because the upgrade leaves behind some problems. What you actually need is to PURGE the 5.0 version an then install the 5.21 version. Which is not obvious at all and very much counter intuitive. [2]

So to solve both issues, you could just add an LXD purge before the LXD install in the instructions.

[1] - the error you get is `Error: Failed to get OVN client: Failed to get OVN southbound connection string: Failed to run: ovs-vsctl get open_vswitch . external_ids:ovn-remote: exit status 1 (ovs-vsctl: unix:/var/run/openvswitch/db.sock: database connection failed (No such file or directory))`
[2] - the error you get is `Error: System "ob76-node10" failed to join the cluster: Failed to update cluster status of services: Failed to join "LXD" cluster: Failed to configure cluster :Failed to initialize member: Failed to initialize storage pools and networks: Failed to update local member project "default": Invalid project configuration key "features.networks.zones"`

---
*Reported from: https://canonical-microcloud.readthedocs-hosted.com/en/latest/microcloud/how-to/install/*

gabrielmougard:
@token47 what do you think about https://github.com/canonical/microcloud/pull/437 :) ?

roosterfish:
@token47 on [1] I am not sure if MicroCloud 1.1 was ever intended to run with LXD 5.0.x. IIRC this error is caused by not having at least LXD 5.21.x.

PR #437, opened 2024-10-18
doc: Add documentation note for when LXD 5.0 is already installed

gabrielmougard:
fixes https://github.com/canonical/microcloud/issues/410

### DIFF
--- a/doc/how-to/install.md
+++ b/doc/how-to/install.md
@@ -101,5 +101,10 @@ If you don't want to use MicroCloud's full functionality, you can install only s
 However, this is not recommended.
 ```
 
+```{note}
+It's possible that a required snap is already installed on your machine. For example, it might be running a version of Ubuntu that comes with LXD pre-installed.
+In this case, run `sudo snap refresh <snap> --channel=<track>/stable --cohort="+"`
+``` to refresh (update) the installed snap.
+
 After installing the snaps make sure to hold any automatic updates to keep the used snap versions across MicroCloud in sync.
 See {ref}`howto-snap-hold-updates` for more information.


token47 (approved):
Looks good to me!!

doc/how-to/install.md
```diff
@@ -17,3 +17,8 @@ See {ref}`howto-snap` for more information.
 If you don't want to use MicroCloud's full functionality, you can install only some of the snaps.
 However, this is not recommended.
 ```
+
+```{note}
+Depending on you machine OS and state, the LXD snap might already be installed (e.g, you have a version of Ubuntu on the machine having LXD 5.0 already installed).
```
roosterfish, line 22:
```suggestion
Depending on your machines OS and state, the LXD snap might already be installed (e.g you have a version of Ubuntu on the machine with LXD 5.0 already installed).
```

doc/how-to/install.md
```diff
@@ -17,3 +17,8 @@ See {ref}`howto-snap` for more information.
...
+In this case instead of doing `sudo snap install lxd --channel=5.21/stable --cohort="+"`, just do `sudo snap remove lxd --purge && sudo snap install lxd --channel=5.21/stable --cohort="+"`
```
roosterfish, line 23: You don't have to uninstall the snap as you can just refresh it:

```suggestion
In this case instead of running `sudo snap install lxd --channel=5.21/stable --cohort="+"`, refresh the snap with `sudo snap refresh lxd --channel=5.21/stable --cohort="+"`
```

roosterfish (changes requested):

doc/how-to/install.md
token47: This does not work. I specifically mentioned issues if refreshing (my second bullet in original post). Something is left behind during the upgrade process. This may be a separate bug, but for now it is what happens.

token47: Somehow you get "Invalid project configuration key "features.networks.zones"" if you refresh, but you don't get this error if you purge and install again.

masnax: I remember this issue. Looking at LXD, `features.networks.zones` does have support in [`5.21`](https://github.com/canonical/lxd/blob/a921fc37579ab2edc2d103f77416bfeb9aba552e/lxd/api_project.go#L1081), but is not supported in [`5.0`](https://github.com/canonical/lxd/blob/04d273193898512178bd46072a0948f7a60a0aa1/lxd/api_project.go#L1078), which suggest the rejecting system was still on `5.0`, but being supplied data created on `5.21` since the config field is set. Since we got so far as updating tables on the joiner, what's odd to me is that such a combination passed API/database version checking in LXD in the first place. 

I do remember from the initial Mattermost thread that the snaps were listed as `5.21`.

Worth a bit more investigation I think. @token47 are you able to reproduce this with any consistency?

masnax: I just gave this a try by running `microcloud init` on a system with lxd 5.21 installed, and lxd 5.0 installed on a joiner. Indeed this is the error I see:
```
Error: Failed to join cluster: Failed to initialize member: Failed to initialize storage pools and networks: Failed to update local member project "default": Invalid project configuration key "features.networks.zones"
```

So it seems we eagerly try to update the projects table before doing any API/database version checking in `5.0`, thus getting back a misleading/non-descriptive error. The real error is the daemon versions are incompatible.

Still not sure why `snap info` listed `5.21`, but maybe I just failed to catch a `5.0` in there.

doc/how-to/install.md
gabrielmougard: fixed

doc/how-to/install.md
token47: Interestingly, I'm not being able to reproduce this anymore. Maybe something changed on LXD snap in the last days?

This is a test without upgrading lxd snap from 5.0 to 5.21 -> https://pastebin.canonical.com/p/mjpSWMhKmp/
This is a test with the refresh (not purge+install) -> https://pastebin.canonical.com/p/q75c9c25tc/

As you can see, second test was giving me error before but now seems to have worked.

minaelee:
Question: Does the same apply for all snaps, not just lxd? That is, could a machine intended for a MicroCloud cluster already have microceph or microovn installed previously, and thus also need to those to be refreshed instead of installed? If so, the wording of the addition should be changed. If not, disregard.

roosterfish:
> Question: Does the same apply for all snaps, not just lxd?

@minaelee valid point. You can potentially already have LXD, MicroCloud, MicroCeph and MicroOVN installed from either the right or different channel.

doc/how-to/install.md
roosterfish: @token47 IIRC the error in the first link seems to be caused by not having at least LXD 5.21.x.

If this still isn't reproducible anymore please let us know so we can close the PR and the related issue. LXD 5.21/stable channel was updated recently so maybe it contained a fix for this.

token47: Sorry but have been unable to test this since. You seem to have hit the same issue, so maybe you could try one more time to make sure it was not a fluke for me? Maybe something actually changed that indirectly fixed this.

roosterfish: Just tested this again with three machines refreshing from `5.0/stable` to `5.21/stable`. It's working fine.

roosterfish:
@gabrielmougard can we please merge this? The original issue which didn't allow the refresh doesn't seem to be around anymore.

Based on the comment from @minaelee, can you please rephrase the note saying that if one of the snaps is already installed, you can replace `install` with `refresh` to install the required version?

roosterfish (approved):

doc/how-to/install.md
```diff
@@ -101,5 +101,10 @@ If you don't want to use MicroCloud's full functionality, you can install only s
 However, this is not recommended.
 ```
 
+```{note}
+Depending on your machines OS and state, the LXD snap might already be installed (e.g you have a version of Ubuntu on the machine with LXD 5.0 already installed).
+In this case run `sudo snap refresh <snap> --channel=<track>/stable --cohort="+"`
```
roosterfish, line 106: Oops, just wanted the merge this, there is a dot missing at the end. Could you please add this one in?

roosterfish (changes requested):

doc/how-to/install.md
```diff
@@ -101,5 +101,10 @@ If you don't want to use MicroCloud's full functionality, you can install only s
 However, this is not recommended.
 ```
 
+```{note}
+Depending on your machines OS and state, the LXD snap might already be installed (e.g you have a version of Ubuntu on the machine with LXD 5.0 already installed).
```
minaelee, line 105:
```suggestion
It's possible that a required snap is already installed on your machine. For example, it might be running a version of Ubuntu that comes with LXD pre-installed.
```

doc/how-to/install.md
minaelee, line 106:
```suggestion
In this case, run `sudo snap refresh <snap> --channel=<track>/stable --cohort="+"`
``` to refresh (update) the installed snap.

doc/how-to/install.md
gabrielmougard: fixed

doc/how-to/install.md
gabrielmougard: fixed

PR #438, opened 2024-10-18
Cmd: Wait for the local daemon to become ready

roosterfish:
I have seen [this](https://github.com/canonical/microcloud/actions/runs/11404169982/job/31732770512?pr=436) error multiple times now in the pipeline which indicates that the local microcluster daemon isn't yet ready and therefore hasn't yet setup its socket.

This PR puts a check in each CLI command to wait until the daemon is ready.

### DIFF
--- a/cmd/microcloud/add.go
+++ b/cmd/microcloud/add.go
@@ -60,6 +60,11 @@ func (c *cmdAdd) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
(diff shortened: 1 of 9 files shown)


masnax (changes requested):
Why not put this check in the `Start` hook so that it's not duplicated everywhere and applies to all MicroCloud commands?

masnax (approved):

PR #439, opened 2024-10-18
Restrict removing from a 2 node ceph cluster

masnax:
MicroCeph does not clean up the monmap on cluster removal below 3 nodes, so if we have a 2-node cluster and we remove a node which is still in the monmap, the remaining node will become unresponsive until the monmap is manually repaired. 

In our case, it makes more sense to just inform the user to do this _before_ it gets to a point where their ceph cluster is broken.


api/services_cluster.go
```diff
@@ -59,6 +60,35 @@ func removeClusterMember(state state.State, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	ceph := sh.Services[types.MicroCeph]
+	if ceph != nil {
+		// If we got a 503 error back, that means the service is installed, but hasn't been set up yet, so there are no cluster members to remove.
```
roosterfish, line 66: I understand this situation can happen if you form a MicroCloud without MicroCeph snap installed, and then at a later point in time install MicroCeph on this member and try to remove it afterwards?

doc/how-to/remove_machine.md
```diff
@@ -10,8 +10,29 @@ Before removing the machine, ensure that there are no LXD instances or storage v
...
 
+```{note}
+If local storage was created, MicroCloud will have also added some default storage volumes that will need to be cleaned up:
+
+    lxc config unset storage.images_volume --target <name>
+    lxc config unset storage.backups_volume --target <name>
+    lxc storage volume delete local images --target <name>
+    lxc storage volume delete local backups --target <name>
+
+Any additional storage volumes belonging to this machine must also be deleted before removal without the `--force` flag.
```
roosterfish, line 21: Can we run those cleanup steps against the LXD API if the user provides `--force`? 
Having to provide `--force` but still requiring to delete resources upfront seems a bit too much.

doc/how-to/remove_machine.md
```diff
@@ -10,8 +10,29 @@ Before removing the machine, ensure that there are no LXD instances or storage v
...
 
 ```{caution}
 Removing a cluster member with `--force` will not attempt to perform any clean-up of the removed machine. All services will need to be fully re-installed before they can be re-initialised. Resources allocated to the MicroCloud like disks and network interfaces may need to be re-initialised as well.
 ```
+
+### Reducing the cluster to 1 machine
+
+When shrinking the cluster down to 1 machine, you must also clean up the Ceph monmap before proceeding, even when using the `--force` flag.
+
+    sudo microceph.ceph mon remove <name>
```
roosterfish, line 34: Is it worth having another test cover this bit?

roosterfish (changes requested):
Looking good only a few smaller comments.
Also nice find with the replication factor resize when scaling down.

doc/how-to/remove_machine.md
masnax: `--force` is basically "skip cleanup". You`d use `--force` if the node is not available for cleanup in the first place, like if the node is unreachable. To get below 3 nodes, MicroCeph requires us to pass `--force`, and we cant manage underlying Ceph/OVN cleanup, so regardless of any cleanup we do on LXD, the node would never be sufficiently "cleaned" anyway. 

Also I'm generally hesitant to delete user-created volumes & instances silently.

masnax: Also we try to be proactive, so if you run `microcloud remove` without `--force`, then everything except MicroCeph and MicroCloud will be removed cleanly. Then you would need to rerun `microcloud remove --force` for the same node to remove the last 2 services.

doc/how-to/remove_machine.md
masnax: I've added a test case for shrinking down to 1 node, which requires this, as well as checking that we can't remove the node even with `--force` until it is removed from MicroCeph's monmap.

Additionally, it seems a lot of the remove tests were incorrectly ignoring expected errors, so those are fixed too.

As well, I noticed a small bug with `microcloud status` where uninitialized services were being ignored because the snap was installed. This is fixed by also reporting the service as missing if it is installed, but its cluster member count is 0.

masnax: I just gave this a try by running `microcloud init` on a system with lxd 5.21 installed, and lxd 5.0 installed on a joiner. Indeed this is the error I see:
```
Error: Failed to join cluster: Failed to initialize member: Failed to initialize storage pools and networks: Failed to update local member project "default": Invalid project configuration key "features.networks.zones"
```

So it seems we eagerly try to update the projects table before doing any API/database version checking in `5.0`, thus getting back a non-descriptive error.

Still not sure why `snap info` listed `5.21`, but maybe I just failed to catch a `5.0` in there.

doc/how-to/remove_machine.md
```diff
@@ -10,8 +10,29 @@ Before removing the machine, ensure that there are no LXD instances or storage v
...
+
+## Reducing the cluster to 1 machine
+
+When shrinking the cluster down to 1 machine, you must also clean up the Ceph monmap before proceeding, even when using the `--force` flag.
+
+    sudo microceph.ceph mon remove <name>
+    sudo microceph cluster sql "delete from services where member_id = (select id from core_cluster_members where name='<name>') and service='mon'"
+    sudo microcloud remove <name> --force
+
+If the machine is no longer reachable and Ceph is no longer responsive, see the [Ceph documentation](https://docs.ceph.com/en/reef/rados/operations/add-or-rm-mons/#removing-monitors-from-an-unhealthy-cluster) for more recovery steps.
```
roosterfish, line 38: Worth already linking to the `squid` equivalent instead of `reef`?

roosterfish (changes requested):
Sorry haven't seen it in the last review, only one more comment in regards to the external link.

doc/how-to/remove_machine.md
roosterfish: thanks!

doc/how-to/remove_machine.md
masnax: Ah good catch! Fixed now.

PR #445, opened 2024-10-22
CLI: Remove redundant flags

roosterfish:
Whilst the `--auto` features was present having `--address`, `--wipe` and `--encrypt` made sense to influence what is happening in the auto mode.

Now after the auto mode is gone those flags have to be removed too.
This is a leftover.

Automating those settings can be done by using the preseed support.

### DIFF
--- a/cmd/microcloud/main_init.go
+++ b/cmd/microcloud/main_init.go
@@ -103,9 +103,6 @@ type initConfig struct {
 	// sessionTimeout is the duration to wait for the trust establishment session to complete.
 	sessionTimeout time.Duration
 
-	// wipeAllDisks indicates whether all disks should be wiped, or if the user should be prompted.
-	wipeAllDisks bool
-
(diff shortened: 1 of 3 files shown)


cmd/microcloud/main_init.go
```diff
@@ -154,15 +148,12 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
 	}
 
 	cfg := initConfig{
-		bootstrap:       true,
-		setupMany:       true,
-		address:         c.flagAddress,
-		wipeAllDisks:    c.flagWipeAllDisks,
-		encryptAllDisks: c.flagEncryptAllDisks,
```
masnax, line 161: These fields are no longer assigned on the struct, so I think we can remove them.

masnax (changes requested):

masnax (approved):

PR #446, opened 2024-10-22
Release MicroCloud 2.1 LTS

masnax:
Ideally, this should be merged after updating go module dependencies and finishing up the last few bugfixes.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,5 +1,21 @@
 // Package version provides shared version information.
 package version
 
-// Version is the current version of MicroCloud.
-const Version = "1.1"
+import (
+	"fmt"
+)
+
+// RawVersion is the current daemon version of MicroCloud.
+const RawVersion = "2.1"
+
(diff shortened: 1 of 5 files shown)


version/version.go
```diff
@@ -1,5 +1,21 @@
...
+
 // Version is the current version of MicroCloud.
-const Version = "1.1"
+const Version = "2.1"
+
+// LTS should be set if the current version is an LTS (long-term support) version.
+const LTS = true
+
+// LTSVersion appends "LTS" to the version string if MicroCloud is an LTS version.
+func LTSVersion() string {
```
tomponline, line 15: Why not just Version()? Seems odd to have a function called LTSVersion not always return the LTS suffix.

masnax: I've now renamed the raw version const to `RawVersion` and this helper is just `Version`.

roosterfish (approved):

PR #447, opened 2024-10-22
test/suites: Ensure ping is installed

masnax:
Sometimes cloud-init fails to install `ping` so try to install it manually as well.

### DIFF
--- a/test/suites/basic.sh
+++ b/test/suites/basic.sh
@@ -486,8 +486,8 @@ EOF
 "
 
   # Launch a container and VM with CEPH storage & OVN network.
-  lxc exec micro01 -- lxc init ubuntu-minimal:22.04 v1 -c limits.memory=512MiB -d root,size=3GiB --vm -s remote -n default
-  lxc exec micro01 -- lxc init ubuntu-minimal:22.04 c1 -c limits.memory=512MiB -d root,size=3GiB -s remote -n default
+  lxc exec micro01 -- lxc init ubuntu-minimal-daily:22.04 v1 -c limits.memory=512MiB -d root,size=3GiB --vm -s remote -n default
+  lxc exec micro01 -- lxc init ubuntu-minimal-daily:22.04 c1 -c limits.memory=512MiB -d root,size=3GiB -s remote -n default
 
   # Ensure proper storage pool and network selection by inspecting their used_by.
   for m in c1 v1 ; do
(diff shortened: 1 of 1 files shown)


test/suites/basic.sh
```diff
@@ -697,11 +697,18 @@ EOF
   IPV4_GW="$(lxc network get lxdbr0 ipv4.address | cut -d/ -f1)"
   IPV6_GW="$(lxc network get lxdbr0 ipv6.address | cut -d/ -f1)"
 
+  # Sometimes ping is not installed by cloud-init, so try to install it manually
+  for m in c1 c2 ; do
+    lxc exec micro01 -- lxc exec "${m}" -- apt-get install -y --no-install-recommends iputils-ping
```
roosterfish, line 702: When launching an `ubuntu-minimal:22.04`, it looks we also have to run `apt-get update` first. I fear cloud-init might have done it so the install works but maybe sometimes it's so slow that it didn't even update the package lists.

As part of https://github.com/canonical/microcloud/pull/436 the images are now pre-fetched on the host (runner) and `security.devlxd.images=true` is set on the `micro*` VMs.
Whilst you are at it can you use the `ubuntu-minimal-daily` instead to fetch the image directly from the host instead of downloading it again? I think the `basic` suite is the only other one that does these workload tests.

test/suites/basic.sh
```diff
@@ -697,11 +697,18 @@ EOF
...
+  done
+
   lxc exec micro01 -- lxc exec c1 -- ping -nc1 -w5 -4 "${IPV4_GW}"
   lxc exec micro01 -- lxc exec c2 -- ping -nc1 -w5 -4 "${IPV4_GW}"
   lxc exec micro01 -- lxc exec c1 -- ping -nc1 -w5 -6 "${IPV6_GW}"
   lxc exec micro01 -- lxc exec c2 -- ping -nc1 -w5 -6 "${IPV6_GW}"
   if [ "${SKIP_VM_LAUNCH}" != "1" ]; then
+    # Sometimes ping is not installed by cloud-init, so try to install it manually
+    lxc exec micro01 -- lxc exec "v1" -- apt-get install -y --no-install-recommends iputils-ping
```
roosterfish, line 711: As it's now always installed manually, I think we can also get rid of the install instruction in the cloud init file.

roosterfish (changes requested):
Good catch, only one remark regarding update of the package lists.

test/suites/basic.sh
masnax: Fixed.

test/suites/basic.sh
masnax: Fixed.

roosterfish (approved):

PR #449, opened 2024-10-24
Various print fixes

masnax:
* Corrects spacing for OVN underlay selection summary
* Corrects spacing for OSD selections if the recommended disks warning is rejected
* Mentions "OVN" in the underlay question
* Changes up the grammar of the system selection prompt
* Adds text feedback to the `preseed` command to display which systems are expected and give feedback once the initiator detects them, and vice versa. 
* Omits an extraneous ceph cluster network log that was showing up in the preseed command.


roosterfish (approved):
Thanks!

PR #450, opened 2024-10-25
Version: Set the .0 patch number for the 2.1.0 LTS

roosterfish:
Use the same LTS versioning scheme as LXD and include the .0 patch number for LTS versions.


tomponline (approved):
Ta

PR #455, opened 2024-10-28
Check MicroCloud & LXD's initialization status & don't restart LXD.

masnax:


cmd/microcloud/ask.go
```diff
@@ -19,12 +19,48 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	s, err := service.NewHandler("", "", stateDir, installedServices...)
+	if err != nil {
+		return err
+	}
+
+	return s.RunConcurrent("", "", func(s service.Service) error {
```
roosterfish, line 48: Worth returning an combined error from `RunConcurrent()` in case multiple services are (not) initialized? I think right now it only shows the first error which is encountered.

cmd/microcloud/join.go
```diff
@@ -42,6 +42,12 @@ func (c *cmdJoin) Run(cmd *cobra.Command, args []string) error {
 		return cmd.Help()
 	}
 
+	fmt.Println("Waiting for services to start ...")
```
roosterfish, line 45: Will this now ever reach the code down below?

```go
if status.Ready {
	return fmt.Errorf("MicroCloud is already initialized, run 'microcloud add' instead")
}
```

roosterfish (changes requested):

cmd/microcloud/ask.go
masnax: That might be too noisy, the errors will be listed hierarchically due to error wrapping, but they won't be related to each other.

cmd/microcloud/join.go
masnax: True, I'll remove it.

masnax: Fixed.

masnax:
Had to make some tweaks to `waitReady`. Since we need to wait for LXD to get through `daemon.start` and `daemon.activate` in the snap before its unix socket will appear, we couldn't easily track the difference between LXD taking a long time to start and LXD outright not being installed.

To get around that, I've added a 404 error to the proxy handlers to check for the unix socket, and in `IsInitialized` we check for this and return `Unix socket not found. Check if {service} is installed`

masnax:
> > Had to make some tweaks to `waitReady`. Since we need to wait for LXD to get through `daemon.start` and `daemon.activate` in the snap before its unix socket will appear, we couldn't easily track the difference between LXD taking a long time to start and LXD outright not being installed.
> > To get around that, I've added a 404 error to the proxy handlers to check for the unix socket, and in `IsInitialized` we check for this and return `Unix socket not found. Check if {service} is installed`
> 
> Do you use a sentinel error or api.StatusError?

There's a manual `os.Stat` for the unix socket. If that returns an error then we use `api.StatusError` to send a `404`. We use the `waitReady` implementation from LXD so the above will be repeatedly called for 10s. 

On the client side, after 10s, if we detect this `404`, we ignore upstream error and return a new error with the `check if service is installed` message. If we detect any other error then it will be padded through.

service/lxd.go
```diff
@@ -721,7 +754,7 @@ func (s *LXDService) Restart(ctx context.Context, timeoutSeconds int) error {
 }
 
 // waitReady repeatedly (500ms intervals) asks LXD if it is ready, up to the given timeout.
-func (s *LXDService) waitReady(ctx context.Context, c lxd.InstanceServer, timeoutSeconds int) error {
+func (s *LXDService) waitReady(ctx context.Context, c lxd.InstanceServer, timeoutSeconds time.Duration) error {
```
roosterfish, line 700: IIRC we talked about this last week, can't the timeout be set through the provided context from the outside?

cmd/microcloud/add.go
```diff
@@ -68,10 +68,6 @@ func (c *cmdAdd) Run(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("Failed to get MicroCloud status: %w", err)
 	}
 
-	if !status.Ready {
-		return fmt.Errorf("MicroCloud is uninitialized, run 'microcloud init' first")
```
roosterfish, line 71: In case MicroCloud is uninitialized, should we still add the hint to run `microcloud init` in the new error message?

roosterfish (changes requested):
There are still two manual checks for the MicroCloud status in `microcloud service list` and `microcloud status`.
I think we should also use `checkInitialized` in those cases?

service/lxd.go
masnax: Fixed.

cmd/microcloud/add.go
masnax: I've added a check on the CLI side that detects if the service is MicroCloud, and if we are not in preseed mode, it hints to use either `microcloud init` or `microcloud add` according to the expectation.

roosterfish: Thanks!

roosterfish (dismissed):
Looks good!

masnax:
> There are still two manual checks for the MicroCloud status in `microcloud service list` and `microcloud status`. I think we should also use `checkInitialized` in those cases?

* For `microcloud status`, we only care about the initialization status of MicroCloud. If any other service's initialization status is unexpected, this should be reported in the status output and not in the form of an error.

* Similarly for `MicroCloud service list` we want to list the initialized services so we only care about MicroCloud being initialized so we can query the other services.

Even if we do refactor `checkInitialized` to support an "Only MicroCloud" check, we would still need to call the status API to get the cluster name and address so it wouldn't simplify much.

service/lxd.go
```diff
@@ -756,18 +735,11 @@ func (s *LXDService) waitReady(ctx context.Context, c lxd.InstanceServer, timeou
...
-		case <-ctx.Done():
-			return fmt.Errorf("LXD is still not running after %ds timeout (%v)", timeoutSeconds, errLast)
-		}
-	} else {
-		<-finger
+	select {
+	case <-finger:
+		break
+	case <-ctx.Done():
+		return fmt.Errorf("Timed out waiting for LXD to start: %w", errLast)
```
roosterfish, line 742: There is one error in the pipeline showing ` Error: Timed out waiting for LXD to start: %!w(<nil>)`. Should we instead show the context's error using `context.Cause(ctx)`? 

Maybe within `waitReady` we can derive another context which can be cancelled by the goroutine using a custom error (lastError if present). You can use `WithCancelCause` when deriving the new context.

masnax: OK I've made it a bit more complicated but it eases up the UX a bit:

The `finger` channel is gone now. Instead there's 2 contexts:
The main context has a 1 minute deadline when the function is invoked, similar to before. There is now an additional `unixCtx` that waits up to 5 seconds for the LXD unix socket to appear. If this context's deadline is exceeded and we still haven't found the unix socket, the error returned from the API is used to cancel the main context.

The function blocks on the main context, and will return an error if the context cause is either `DeadlineExceeded`, or _not_ `Canceled`, which implies the issue was detecting the unix socket.

service/lxd.go
```diff
@@ -742,6 +785,14 @@ func (s *LXDService) waitReady(ctx context.Context, c lxd.InstanceServer, timeou
 
 			_, _, err := c.RawQuery("GET", "/internal/ready", nil, "")
 			if err != nil {
+				if api.StatusErrorCheck(err, http.StatusNotFound) {
```
roosterfish, line 731: In case `c.RawQuery` returns a non nil error which does not match `http.StatusNotFound` it will loop forever.
This should not be a problem as the outer func exits after the provided parent ctx gets cancelled but this will leak the routine as it never exits. 

Maybe to go routine can be structured like this to ensure it exits after the context is cancelled:

```go
go func() {
    for i := 0; ; i++ {
        select {
            case <-unixCtx.Done():
                return // Exit the routine
            default:
                // run the actual logic
        }
    }
}()
```

roosterfish (changes requested):

service/lxd.go
roosterfish: Ah, wanted to follow up on this thread, see https://github.com/canonical/microcloud/pull/455#discussion_r1834602451.

service/lxd.go
masnax: Not quite, `unixCtx` being done isn't sufficient to exit the goroutine, it's only a premature check in case we can't find the unix socket. 

Instead I think we can just check if `ctx.Err` is not nil and exit if so in each iteration.

masnax: Additionally, I just noticed the logger is useless since this is run on the client side anyway.

masnax: I've also set a default 1 minute timeout in the `waitReady` function itself in case a caller forgets to set one on the context. This is also mentioned in the description.

roosterfish (approved):

PR #463, opened 2024-11-05
doc: Account for pre-installed LXD in the tutorial

MggMuggins:
Requested by @escabo

### DIFF
--- a/doc/tutorial/get_started.md
+++ b/doc/tutorial/get_started.md
@@ -55,7 +55,11 @@ Complete the following steps to install and initialise LXD:
           sudo apt update
           sudo apt install snapd
 
-1. Enter the following command to install LXD:
+1. If LXD is already installed, enter the following command to update it:
+
+       sudo snap refresh lxd
+
+   Otherwise, enter the following command to install LXD:
 
        sudo snap install lxd
 


roosterfish (approved):

escabo (approved):

masnax:
I think we need to specify the channel with `--channel 5.21/stable` otherwise the refresh will keep tracking the pre-installed channel, which is usually `5.0/stable`.

roosterfish:
> I think we need to specify the channel with `--channel 5.21/stable`

Good point. 

Should we maybe tell the user to install the machines using `ubuntu:24.04` instead of `ubuntu:22.04`?
In this case LXD would not be there and have to be installed using `snap install lxd`.

masnax:
> > I think we need to specify the channel with `--channel 5.21/stable`
> 
> Good point.
> 
> Should we maybe tell the user to install the machines using `ubuntu:24.04` instead of `ubuntu:22.04`? In this case LXD would not be there and have to be installed using `snap install lxd`.

In the guide, probably yes. However we should still include the `refresh` line because MicroCloud is supposed to support `22.04` I believe.

roosterfish:
What I have learned in the past is that the tutorial(s) should "force" you into doing one exact thing without much if/else so if we say to install 24.04 we should assume that is what the user did an we can subsequently tell to install LXD.

masnax:
> What I have learned in the past is that the tutorial(s) should "force" you into doing one exact thing without much if/else so if we say to install 24.04 we should assume that is what the user did an we can subsequently tell to install LXD.

Yeah, I meant we should still have a line _somewhere_ in the docs that mentions if LXD is preinstalled, it should be refreshed to `5.21/stable`. This doesn't necessarily belong in the tutorial but maybe other places where we show that the user should run `snap install lxd`.

MggMuggins:
Anything we can do to decrease friction in the tutorial. My feeling is that users will try to diverge from the tutorial's requirements in various ways. If we can provide the real requirement (MicroCloud requires LXD 5.21), that will allow users to get past this without too much difficulty. Opened #468

PR #465, opened 2024-11-07
github: Use cache when DB download fails

hamistao:
This updates the Trivy workflow to cache the vulnerability database when the download is successful and, when it is not, use the latest cached database instead. Motivated by some failures seen on some recent runs: https://github.com/canonical/microcloud/actions/runs/11638102909/job/32412374433

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -30,16 +30,33 @@ jobs:
       - name: Install Trivy
         uses: canonical/lxd/.github/actions/install-trivy@main
 
+      - name: Download Trivy DB
+        id: db_download
+        run: trivy fs --download-db-only --cache-dir /home/runner/vuln-cache
+        continue-on-error: true
+
+      - name: Use previous downloaded database
+        if: ${{ steps.db_download.outcome == 'failure' }}
+        uses: actions/cache/restore@3624ceb22c1c5a301c8db4169662070a689d9ea8 # v4.1.1
+        with:
+          path: /home/runner/vuln-cache
+          key: trivy-latest-cache
+
       - name: Run Trivy vulnerability scanner
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/security.yml
```diff
@@ -49,7 +66,7 @@ jobs:
           ref: refs/heads/main
 
   trivy-snap:
-    name: Trivy vulnerability scanner - Snap
+    name: Trivy - Snap
```
roosterfish, line 69: Unrelated to this PR, but can we run the check for multiple snaps channels using a matrix build? 
I am thinking we could check `2/candidate` and `2/stable`. 

But we would probably need an option to group the results so they are not getting mixed up.
I have quickly checked the action's [docs](https://github.com/github/codeql-action/blob/v3/upload-sarif/action.yml) and it looks there is a `category` setting we could potentially use. 
Have you maybe tried something like this already?

hamistao: I did look into this `category` setting but it doesn't help filter the scan results on the security tab :/

What we do on LXD is we place results for each `{version}/stable` channel in its corresponding branch. So `5.21/stable` scan results are tagges as belonging to `stable-5.21` branch.

hamistao: @roosterfish  What do you think of placing `2/stable` results in `v2-edge` and `3/stable` in the main branch?

roosterfish: Sounds like a good plan.

roosterfish: Ah but we don't yet have `3/stable`. 
There will be a `2/stable` starting from next week we can check.
And going forward there will be only be a `3/edge` but I assume we can use this channel instead.

roosterfish: Maybe for now we can use `3/edge` in the `main` branch and `2/candidate` in the `v2-edge` until we release the LTS.

roosterfish (approved):

roosterfish:
Thanks!

PR #466, opened 2024-11-07
Doc: Add minimal preseed examples

roosterfish:
This PR adds minimal preseed examples for both initialization and growing of a MicroCloud.
In addition the preseed reference file now lists all the required keys.

### DIFF
--- a/doc/how-to/add_machine.md
+++ b/doc/how-to/add_machine.md
@@ -28,5 +28,5 @@ The preseed YAML file must use the following syntax:
 
 ```{literalinclude} preseed.yaml
 :language: YAML
-:emphasize-lines: 1-3,6-8,11-12,15-17,20,23-25,28-33,61-64,70,77-85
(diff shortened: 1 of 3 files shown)


doc/how-to/add_machine.md
```diff
@@ -28,5 +28,21 @@ The preseed YAML file must use the following syntax:
...
+You can use the following minimal preseed file to add another machine to an existing MicroCloud.
+In this case `micro01` takes over the role of the initiator.
+Multicast discovery is used to find the existing MicroCloud on the network:
+
+```yaml
+lookup_subnet: 10.0.0.0/24
+initiator: micro01
+session_passphrase: foo
+systems:
+- name: micro04
```
masnax, line 45: Should we also add an uplink interface & disks for local & remote storage to these examples?

They will need to be selected on the joining node, and joining a cluster that has the associated networks & storage pools set up will work, but will leave the storage pools & networks in an errored state.

roosterfish: Fixed.

masnax (approved):

PR #468, opened 2024-11-07
doc/tutorial: Explicitly require LXD 5.21

MggMuggins:
As discussed in #463

### DIFF
--- a/doc/tutorial/get_started.md
+++ b/doc/tutorial/get_started.md
@@ -31,7 +31,7 @@ It uses virtual machines in LXD, so you don't need any extra hardware to follow 
    We recommend a storage pool size of at least 40 GiB for that.
 ```
 
-Complete the following steps to install and initialise LXD:
+MicroCloud requires LXD version 5.21:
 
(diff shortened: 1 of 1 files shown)


doc/tutorial/get_started.md
```diff
@@ -57,11 +57,11 @@ Complete the following steps to install and initialise LXD:
 
 1. If LXD is already installed, enter the following command to update it:
 
-       sudo snap refresh lxd
+       sudo snap refresh lxd --channel=5.21/stable
 
    Otherwise, enter the following command to install LXD:
 
-       sudo snap install lxd
+       sudo snap install lxd --channel=5.21/stable
```
masnax, line 64: LTS will always be set to the default track so it shouldn't be necessary to include it for `snap install`.

masnax (changes requested):

doc/tutorial/get_started.md
tomponline: True. But if you want a specific lts series that is compatible it would be worth making it explicit.

PR #469, opened 2024-11-08
github: Use Canonical runners for scheduled system tests

roosterfish:
This PR splits the rather complex matrix system test into three groups.
The system test code is moved into a repo local action which is leveraged by each of the groups.
See an example of the restructured tests workflow [here](https://github.com/canonical/microcloud/actions/runs/12318431330?pr=469).

In addition the "instances" suite is now executed also on the Canonical runners when the workflow is triggered by schedule.

### DIFF
--- a/api/session_join.go
+++ b/api/session_join.go
@@ -70,11 +70,18 @@ func sessionJoinPost(sh *service.Handler) func(state state.State, r *http.Reques
 
 			// Prevent locking in case there isn't anymore an active consumer reading on the channel.
 			// This can happen if the initiator's websocket connection isn't anymore active.
+			// Wait up to 10 seconds for an active consumer.
+			// When the initiator returns the dice-generated passphrase, a joiner can go ahead and send
+			// its intent to join. If the initiator hasn't yet started to listen on join intents (too slow),
+			// the API might return an error as there isn't yet any active consumer.
+			ctx, cancel := context.WithTimeout(r.Context(), 10*time.Second)
+			defer cancel()
+
 			select {
 			case session.IntentCh() <- req:
 				return nil
-			default:
-				return fmt.Errorf("No active consumer for join intent")
+			case <-ctx.Done():
+				return fmt.Errorf("Timeout waiting for an active consumer of the join intent")
 			}
 		})
 
--- a/.github/actions/system-test/action.yml
+++ b/.github/actions/system-test/action.yml
@@ -1,0 +1,178 @@
+name: System test
+inputs:
+  github_runner:
+    description: Whether or not this test is executed on a GitHub runner.
+    required: true
+    default: true
+runs:
+  using: composite
+  steps:
+    - name: Performance tuning
+      uses: canonical/lxd/.github/actions/tune-disk-performance@main
+
+    - name: Reclaim some space
+      uses: canonical/lxd/.github/actions/reclaim-disk-space@main
+
+    - name: Reclaim some memory
+      shell: bash
+      run: |
+        set -eux
+
+        free -mt
+        sudo systemctl stop dpkg-db-backup.timer e2scrub_all.timer fstrim.timer logrotate.timer man-db.timer motd-news.timer update-notifier-download.timer update-notifier-motd.timer
+        sudo systemctl stop iscsid.socket multipathd.socket
+        sudo systemctl stop cron.service irqbalance.service multipathd.service networkd-dispatcher.service
+        free -mt
+
+    - name: Reclaim some memory (GitHub runners)
+      if: inputs.github_runner == 'true'
+      shell: bash
+      run : |
+        set -eux
+
+        free -mt
+        sudo systemctl stop phpsessionclean.timer
+        sudo systemctl stop mono-xsp4.service php8.1-fpm.service
+        free -mt
+
+    - name: Disable Docker (GitHub runners)
+      if: inputs.github_runner == 'true'
+      uses: canonical/lxd/.github/actions/disable-docker@main
+
+    - name: "Disable br_netfilter"
+      shell: bash
+      run: |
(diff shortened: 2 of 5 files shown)


roosterfish:
@masnax I did some tests regarding [this error](https://github.com/canonical/microcloud/actions/runs/11744653035/job/32720336321?pr=469#step:16:13645).

Unfortunately the timeout set in the MicroCeph [`GetConfig`](https://github.com/canonical/microceph/blob/main/microceph/client/configs.go#L38) client func is set to only 5s.
Based on your suggestion in the meeting earlier, can it be that right after forming the MicroCloud, the proxy is waiting for something in the cluster to settle before it can forward the request to MicroCeph's local unix socket?

I have bootstrapped a single node MicroCloud and fired requests to `/1.0/services/microceph` in parallel. 
Right around when the MicroCloud is bootstrapped I saw a delay in the response which could proof that something is going on there.

masnax:
> I have bootstrapped a single node MicroCloud and fired requests to `/1.0/services/microceph` in parallel. Right around when the MicroCloud is bootstrapped I saw a delay in the response which could proof that something is going on there.

Well this current failure is happening long before MicroCloud is bootstrapped, as it happens right after system discovery and before asking any setup questions. The only delay would be related to refreshing the truststore and waiting for the lock, but even that wouldn't happen on a single-node request as it all goes through the unix socket which skips truststore verification.

>  can it be that right after forming the MicroCloud, the proxy is waiting for something in the cluster to settle before it can forward the request to MicroCeph's local unix socket?

[this is the whole local proxy block in MicroCloud](https://github.com/canonical/microcloud/blob/a829e33a47559a2484118e30afdfe5f0d5de1bea/api/services_proxy.go#L113-L143) so it's definitely not waiting for anything here. 

Since it's a network request, there is the additional overhead of [authHandlerMTLS](https://github.com/canonical/microcloud/blob/a829e33a47559a2484118e30afdfe5f0d5de1bea/api/services_auth.go#L21-L51) pulling the truststore.

masnax:
> > I have bootstrapped a single node MicroCloud and fired requests to `/1.0/services/microceph` in parallel. Right around when the MicroCloud is bootstrapped I saw a delay in the response which could proof that something is going on there.
> 
> Well this current failure is happening long before MicroCloud is bootstrapped, as it happens right after system discovery and before asking any setup questions. In the bootstrap case, the only delay would be related to refreshing the truststore and waiting for the lock, but even that wouldn't happen on a single-node request as it all goes through the unix socket which skips truststore verification. There is also waiting for the listeners to restart post-bootstrap, to pick up the network address. 
> 
> > can it be that right after forming the MicroCloud, the proxy is waiting for something in the cluster to settle before it can forward the request to MicroCeph's local unix socket?
> 
> [this is the whole local proxy block in MicroCloud](https://github.com/canonical/microcloud/blob/a829e33a47559a2484118e30afdfe5f0d5de1bea/api/services_proxy.go#L113-L143) so it's definitely not waiting for anything here.
> 
> Since it's a network request, there is the additional overhead of [authHandlerMTLS](https://github.com/canonical/microcloud/blob/a829e33a47559a2484118e30afdfe5f0d5de1bea/api/services_auth.go#L21-L51) pulling the truststore.

roosterfish:
> Well this current failure is happening long before MicroCloud is bootstrapped, as it happens right after system discovery and before asking any setup questions

Mh it looks we can fix it by waiting for `microceph cluster bootstrap` to settle and only continue if it's done. 
I have added another commit that adds a wrapper function we can use throughout the test suite to wait until `microceph status` reports the single node cluster services are present. I found waiting for this condition looks to be enough.

masnax:
> > Well this current failure is happening long before MicroCloud is bootstrapped, as it happens right after system discovery and before asking any setup questions
> 
> Mh it looks we can fix it by waiting for `microceph cluster bootstrap` to settle and only continue if it's done. I have added another commit that adds a wrapper function we can use throughout the test suite to wait until `microceph status` reports the single node cluster services are present. I found waiting for this condition looks to be enough.

Is this something that can be checked over the API? Perhaps `microceph cluster bootsrap` shouldn't return until all its services are finished, or have a `ready` API that we can check against before sending requests.

roosterfish:
> Is this something that can be checked over the API? Perhaps `microceph cluster bootsrap` shouldn't return until all its services are finished

Issue logged in https://github.com/canonical/microceph/issues/473.

MggMuggins:
IIRC the trust store is held in memory in MicroCluster and synchronized (fanotify?); I think the recovery process doesn't use the in-memory synchronization because it's expected that it won't be accessing the trust store at the same time as any other thread. I wonder if this was an incorrect assumption. I'll try and take a look this afternoon.

api/session_join.go
```diff
@@ -70,11 +70,18 @@ func sessionJoinPost(sh *service.Handler) func(state state.State, r *http.Reques
 
 			// Prevent locking in case there isn't anymore an active consumer reading on the channel.
 			// This can happen if the initiator's websocket connection isn't anymore active.
+			// Wait up to 10 seconds for an active consumer.
+			// When the initiator returns the dice-generated passphrase, a joiner can go ahead and send
+			// its intent to join. If the initiator hasn't yet started to listen on join intents (too slow),
+			// the API might return an error as there isn't yet any active consumer.
+			timer := time.NewTimer(10 * time.Second)
```
masnax, line 77: In general we use contexts for timing. Any particular reason you went with a timer?

roosterfish: No particular reason, not sure why I went with a timer. I'll switch to using a context.

roosterfish: Fixed.

masnax (approved):

tomponline:
needs a rebase please

roosterfish:
@tomponline @masnax rebased.

tomponline (approved):

PR #472, opened 2024-11-08
Add LXD 5.0 upgrade test

masnax:
Technically, you could form a 1.1 MicroCloud with LXD `5.0` so we should test that upgrades will work okay.


masnax:
Looks like it's actually not possible! `5.0` never supported MicroOVN: https://github.com/canonical/lxd-pkg-snap/blob/bedf9beb1c5bf23cd68475349905511de0ad016d/snapcraft/commands/daemon.start#L503

PR #474, opened 2024-11-11
github: Scan proper snap channels

hamistao:
Updates the Trivy workflow to scan the `3/edge` and `2/stable` channels instead of just the `latest`.
Should need another update after the release of the LTS.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -72,13 +72,16 @@ jobs:
     needs: trivy-repo
     strategy:
       matrix:
-        version:
-          - "latest"
+        include:
+          - channel: "3/edge"
+            branch: "main"
+          - channel: "2/candidate"
+            branch: "v2-edge"
     steps:
       - name: Checkout
         uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
         with:
-          ref: ${{ (matrix.version == 'latest' && 'main') || format('stable-{0}', matrix.version) }}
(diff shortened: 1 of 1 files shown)


.github/workflows/security.yml
```diff
@@ -72,26 +72,29 @@ jobs:
     needs: trivy-repo
     strategy:
       matrix:
-        version:
-          - "latest"
+        include:
+          - channel: "3/edge"
```
roosterfish, line 76: `3/edge` doesn't yet exist. Should we comment it out for now?

.github/workflows/security.yml
```diff
@@ -72,26 +72,29 @@ jobs:
...
+            branch: "main"
+          - channel: "2/stable"
```
roosterfish, line 78: Maybe we can set this to `2/candidate` for now until `2/stable` is populated. We can then also uncomment `3/edge` in the same PR.

roosterfish (changes requested):

.github/workflows/security.yml
hamistao: It seems 3/edge does exist:

```
$ snap info microcloud | grep 3/edge
3/edge:           git-ba9113e   2024-11-12 (1125) 10MB -
```

roosterfish: You are right, my mistake, I meant `3/stable`.

roosterfish (approved):

roosterfish (approved):
Looks good!

PR #479, opened 2024-11-11
Multicast: Avoid leaking the context in tests (govet)

roosterfish:
Looks the most recent version of the linter now picks up potential context leaks.
This PR addresses this.


simondeziel (approved):
Thanks!

PR #481, opened 2024-11-11
Misc CI tweaks

simondeziel:
* Use `go.mod` for Go version detection
* Only `strip` binaries if no code coverage is needed
* Use some actions from LXD's repo
* Use ZFS as storage driver in the GHA runner/host
* Pin all actions
* Replace some greps with shell comparison


roosterfish:
Can be rebased now.

roosterfish (approved):
Thanks, looking good!

I saw it on all the other repos too, but what has happened that we started to pin down the actions? I understand it helps always pulling the same one but just to be sure I am not missing anything :)

PR #483, opened 2024-11-12
Update service handler address after init

masnax:
Adds thread safe helpers to get/set the service handler address. 

The address is updated in `PostBootstrap` and `PostJoin` after the MicroCloud daemon is initialized.

### DIFF
--- a/cmd/microcloudd/main.go
+++ b/cmd/microcloudd/main.go
@@ -133,6 +133,19 @@ func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
 		api.OVNProxy(s),
 	}
 
+	setHandlerAddress := func(url string) error {
+		addrPort, err := microTypes.ParseAddrPort(url)
+		if err != nil {
+			return err
+		}
+
+		if addrPort != (microTypes.AddrPort{}) {
+			s.SetAddress(addrPort.Addr().String())
+		}
+
+		return nil
+	}
+
 	dargs := microcluster.DaemonArgs{
 		Verbose:           c.global.flagLogVerbose,
 		Debug:             c.global.flagLogDebug,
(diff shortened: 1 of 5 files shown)


cmd/microcloudd/main.go
```diff
@@ -141,6 +141,16 @@ func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
 
 		PreInitListenAddress: "[::]:" + strconv.FormatInt(service.CloudPort, 10),
 		Hooks: &state.Hooks{
+			PostBootstrap: func(ctx context.Context, state state.State, initConfig map[string]string) error {
+				addrPort, err := microTypes.ParseAddrPort(s.Address())
```
roosterfish, line 145: Worth putting this in an anonymous function so that it can be used by both `PostBootstrap` and the `PostJoin` hook?

roosterfish (changes requested):
I like the idea of populating the address in the hooks! Only one small remark.

cmd/microcloudd/main.go
masnax: Fixed.

service/microcloud.go
```diff
@@ -66,44 +62,7 @@ func NewCloudService(name string, addr string, dir string) (*CloudService, error
...
-						PathPrefix: types.APIVersion,
-						Endpoints:  endpoints,
-					},
-				},
-			},
-		},
-	}
-
+func (s *CloudService) StartCloud(ctx context.Context, args microcluster.DaemonArgs) error {
 	return s.client.Start(ctx, args)
```
roosterfish, line 66: Mhh, in case the daemon gets restarted neither the `PostBootstrap` nor `PostJoin` hooks are getting called. 
Should we set the address here after starting the microcluster daemon?

masnax: Good point. I've moved the `Start` hook out into the daemon CLI package and now it also updates the handler address if the database is already open.

masnax: Had to make a small adjustment because `Database.IsOpen` returns "valid" errors in the case of upgrades.

roosterfish: Looks good! Probably StartCloud func can be removed now too as microcluster's Start can be called right away.

roosterfish (approved):
Looks good!

PR #484, opened 2024-11-13
Additional instance tests

masnax:
Adds two new instance connectivity tests, and splits out instance tests into their own file, as well as splitting out the connectivity checks into their own helper.

The new tests check that instances retain connectivity and can start with a CephFS filesystem attached, when MicroCloud is set up as a fully disaggregated cluster with 5 dedicated networks.

The second test is the same, except that it tests that there is no change when the cluster grows from 2->3 and an instance is launched both on the old members and new members.


roosterfish (dismissed):
Looking good!

masnax:
> Mh the pipeline seems to be consistently failing on the same error https://github.com/canonical/microcloud/actions/runs/11808466614/job/32907868404?pr=484#step:16:51172

Doesn't seem related to the changes here, except maybe the additional interface, but that's not even used in that test suite (`reset_systems 3 2 2`)

It does seem like LXD is either crashing or panicking though.

roosterfish:
> Doesn't seem related to the changes here

I restarted it a few times always with the same outcome. So I am wondering what change in the test suite might trigger this.

masnax:
Looks like this is still failing... not sure exactly what's going on and why it's always at the same point. 

The flow is:
```bash
  + lxc network delete microbr1
  Network microbr1 deleted
  + sudo pkill -9 -f 'dnsmasq .*microbr1'
  includes/microcloud.sh: line 979: 248823 Killed                  sudo pkill -9 -f "dnsmasq .*${network}"
  + true
  + lxc network create microbr1 ipv4.address=10.2.123.1/24 ipv4.dhcp=false ipv4.nat=true ipv6.address=fd42:2:1234:1234::1/64 ipv6.nat=true
  Network microbr1 created
  + lxc profile device add default eth2 nic network=microbr1 name=eth2
 Error: Put "http://unix.socket/1.0/profiles/default?project=microcloud-test": EOF
  ```

Initial network creation seems to work fine, but the network is listed as `UNAVAILABLE` and the interface is `DOWN`, and dnsmasq is up and running. Removing the network while it is in this state does not end the associated dnsmasq process.

@tomponline ever seen anything like this before?

It also seems to be extremely brittle for reproduction. Any changes to the flow of the test (like removing any preceding test case) causes the error to disappear. 

It does seem odd because there should be no lingering state in between tests. Each test deletes and re-creates the storage volumes used as block volumes, and does the same for networks. Then restores the VMs from snapshot.

**Comment by roosterfish** [comment]
Looks this PR had a collision with https://github.com/canonical/microcloud/pull/481. Requires a rebase now.

File: test/includes/microcloud.sh [hunk 1]
```diff
@@ -1050,7 +1050,7 @@ restore_system() {
 
     for n in $(seq 1 "${num_disks}") ; do
       disk="${name}-disk${n}"
-      lxc storage volume create zpool "${disk}" size=5GiB --type=block
+      lxc storage volume create zpool "${disk}" size=10GiB --type=block
```
roosterfish, line 1053: I'll rebase the Canonical runner PR after this got merged and see if it will cause any problems. IIRC the disks on those runners are a bit limited with 40GB (large flavor).

But as the disks aren't thick provisioned probably it will just work.

roosterfish (approved):
LGTM!

PR #485, opened 2024-11-13
github: Fix database cache update

hamistao:
A github cache can not be updated using the same key, so instead we update using a different key each time but use the same key to restore it.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -42,7 +42,7 @@ jobs:
           path: /home/runner/vuln-cache
           key: trivy-cache-${{ github.run_id }}
 
-      - name: Use previous downloaded database
+      - name: Use previously downloaded database instead
         if: ${{ steps.db_download.outcome == 'failure' }}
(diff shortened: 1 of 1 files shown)


roosterfish:
Can be rebased now.

.github/workflows/security.yml
```diff
@@ -53,10 +53,12 @@ jobs:
 
       - name: Cache Trivy vulnerability database
         if: ${{ steps.db_download.outcome == 'success' }}
-        uses: actions/cache/save@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
+        uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
         with:
           path: /home/runner/vuln-cache
-          key: trivy-latest-cache
+          key: trivy-cache-${{ github.run_id }}
```
roosterfish, line 59: What is the default lifetime of those caches? Should we set it to e.g. one week so that GitHub can safely delete them afterwards?

hamistao: The default (and as far as I know immutable) expiry time is of one week already so we should be good :)

roosterfish (approved):

PR #487, opened 2024-11-13
doc/howto: fix broken link to microceph docs

minaelee:
Link to https://canonical-microceph.readthedocs-hosted.com/en/latest/how-to/reef-upgrade/ is now https://canonical-microceph.readthedocs-hosted.com/en/latest/how-to/major-upgrade/ (using Intersphinx syntax `<microceph:how-to/major-upgrade>`).

### DIFF
--- a/doc/custom_conf.py
+++ b/doc/custom_conf.py
@@ -227,7 +227,7 @@ custom_tags = []
 if ('SINGLE_BUILD' in os.environ and os.environ['SINGLE_BUILD'] == 'True'):
     intersphinx_mapping = {
         'lxd': ('https://documentation.ubuntu.com/lxd/en/latest/', None),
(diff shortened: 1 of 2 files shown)


roosterfish (changes requested):
It looks this path has to be updated too https://github.com/canonical/microcloud/blob/main/doc/custom_conf.py#L230 to use `latest` instead of `reef-stable` (as the others).

I ran it locally and it should fix the pipeline.

roosterfish:
I assume the spell check doesn't clone the other repositories locally but instead consults the remote docs if the pages exist. And `how-to/major-upgrade` doesn't exist in the `reef-stable` docs.

minaelee:
> It looks this path has to be updated too https://github.com/canonical/microcloud/blob/main/doc/custom_conf.py#L230 to use `latest` instead of `reef-stable` (as the others).

Thanks! Updated.

Issue #374, opened 2024-08-16
Make sure we pull the correct version of the documentation


ru-fu:
If https://github.com/canonical/microcloud/pull/372 is merged, we'll pull in specific versions of the LXD/MicroCeph/MicroOVN documentation.
These versions can be specified in the Makefile.

They must be updated before the release to match the supported versions. Ideally, this should be done automatically based on our compatibility matrix.

roosterfish:
Can be closed as those version were pinned in https://github.com/canonical/microcloud/pull/490.

PR #490, opened 2024-11-14
Backports and pin down dependencies (v2-edge)

roosterfish:
Backports the entire delta from `main` into `v2-edge`.

Also pins down the Micro* deps for docs, pipeline and go.

Closes https://github.com/canonical/microcloud/issues/374.


roosterfish:
I have now backported that last minute doc changes into this PR. The test run before was already green.

.github/workflows/tests.yml
```diff
@@ -148,10 +148,10 @@ jobs:
         # Set of versions to use for the matrix tests.
         os: ["24.04"]
         go: ["1.22.x"]
-        microceph: ["latest/edge"]
-        microovn: ["latest/edge"]
-        lxd: ["5.21/edge"]
```
masnax, line 105: While we should have tests using the stable track, I think we should also keep testing against `/edge` to catch regressions early.

masnax (approved):

.github/workflows/tests.yml
roosterfish: I think it would be great if we run the same set of tests using the `x/edge` and `x/candidate` channels. 
Maybe let's do this as part of moving to the Canonical runners? It will be a lot of long running jobs.

roosterfish: Or run them on a daily/weekly schedule.

roosterfish: https://github.com/canonical/microcloud/issues/493

PR #495, opened 2024-11-15
doc/howto/initialise: Update initialise instructions

minaelee:
- Replace `node` with `cluster member` for consistent terminology
- Add pre-initialisation requirements to follow install steps & prep for any FDE encryption
- Update the Note annotation in the disk encryption step to advise what to do if disk wasn't prepped

### DIFF
--- a/doc/.custom_wordlist.txt
+++ b/doc/.custom_wordlist.txt
@@ -10,6 +10,7 @@ CIDR
 DCO
 DHCP
 DNS
+FDE
 GiB
 HWE
 intra
--- a/doc/how-to/initialise.md
+++ b/doc/how-to/initialise.md
@@ -8,7 +8,7 @@ You run the initialisation on one of the machines, and it configures the require
 ## Interactive configuration
 
 If you run the initialisation process in interactive mode (the default), you are prompted for information about your machines and how you want to set them up.
-The questions that you are asked might differ depending on your setup; for example, if you do not have the MicroOVN snap installed, you will not be prompted to configure your network, and if your machines don't have local disks, you will not be prompted to set up local storage.
+The questions that you are asked might differ depending on your setup. For example, if you do not have the MicroOVN snap installed, you will not be prompted to configure your network; if your machines don't have local disks, you will not be prompted to set up storage.
 
 The following instructions show the full initialisation process.
 
@@ -30,7 +30,7 @@ Complete the following steps to initialise MicroCloud:
 
 1. Select whether you want to set up more than one machine.
 
-   This allows you to create a MicroCloud using a single node.
+   This allows you to create a MicroCloud using a single cluster member.
    It will skip the {ref}`trust-establishment-session` if no more machines should be part of the MicroCloud.
 
    Additional machines can always be added at a later point in time.
@@ -67,7 +67,8 @@ Complete the following steps to initialise MicroCloud:
 1. Select whether you want to set up distributed storage (using MicroCeph).
 
    ```{note}
-   - To set up distributed storage, you need at least three additional disks on at least three different machines.
+   - You can set up distributed storage on a single cluster member.
+   - High availability requires a minimum of 3 cluster members, with 3 separate disks across 3 different cluster members.
    - The disks must not contain any partitions.
    - A disk that was previously selected for local storage will not be shown for distributed storage.
    ```
@@ -76,7 +77,7 @@ Complete the following steps to initialise MicroCloud:
 
    1. Select the disks that you want to use for distributed storage.
 
-      You must select at least three disks.
+      You must select at least one disk.
    1. Select whether you want to wipe any of the disks.
       Wiping a disk will destroy all data on it.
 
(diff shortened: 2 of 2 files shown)


doc/how-to/initialise.md
```diff
@@ -30,7 +30,7 @@ Complete the following steps to initialise MicroCloud:
 
 1. Select whether you want to set up more than one machine.
 
-   This allows you to create a MicroCloud using a single node.
+   This allows you to create a MicroCloud using a single cluster member.
```
roosterfish, line 39: In the docs we have node, cluster member and machine used on a regular basis.
Would it be the idea to start sticking with one of them?

doc/how-to/initialise.md
```diff
@@ -4,11 +4,17 @@
...
+- If you intend to use full disk encryption (FDE) on any cluster member, that member must meet the prerequisites listed on this page before initialisation: {doc}`microceph:explanation/full-disk-encryption`. 
+  - Follow only the instructions in the Prerequisites section on that page. Skip its Usage section; the MicroCloud initialisation process handles the disk encryption.
+
 (howto-initialise-interactive)=
 ## Interactive configuration
 
 If you run the initialisation process in interactive mode (the default), you are prompted for information about your machines and how you want to set them up.
-The questions that you are asked might differ depending on your setup; for example, if you do not have the MicroOVN snap installed, you will not be prompted to configure your network, and if your machines don't have local disks, you will not be prompted to set up local storage.
+The questions that you are asked might differ depending on your setup. For example, if you do not have the MicroOVN snap installed, you will not be prompted to configure your network; if your machines don't have local disks, you will not be prompted to set up local storage.
```
roosterfish, line 17: "to setup local and/or remote storage."? Both questions won't be asked if there aren't any disks available.

doc/how-to/initialise.md
```diff
@@ -84,8 +90,10 @@ Complete the following steps to initialise MicroCloud:
       Encrypting a disk will store the encryption keys in the Ceph key ring inside the Ceph configuration folder.
 
       ```{note}
-      Encryption requires a kernel with `dm_crypt` enabled.
-      See {doc}`how full disk encryption works <microceph:explanation/full-disk-encryption>` in the MicroCeph documentation for more information.
+      Disks to be encrypted require a kernel with `dm_crypt` enabled. See {doc}`how full disk encryption works <microceph:explanation/full-disk-encryption>` in the MicroCeph documentation for more information.
+
+      If you have not enabled `dm_crypt` on a cluster member disk that you want to encrypt, open a separate terminal and enable it before you continue. If the disk is on the cluster member on which you ran `microcloud init`, stop the initialisation process and start it again after enabling `dm_crypt`.
```
roosterfish, line 95: Isn't `dm_crypt` enabled per cluster member and not disk? 

Even if it's the same cluster member you can keep staying in the init process and launch another terminal to enable `dm_crypt`. You don't have to restart the init process.
The only requirement is that you are still in the questionnaire and MicroCloud hasn't yet started to form and configure the clusters.

roosterfish (changes requested):

doc/how-to/initialise.md
minaelee: Yes, discussed and approved by LXD management team, to reduce confusion where the words `node`, `cluster member`, and `host` are used interchangeably & inconsistently. `machine` can also be a culprit, but it does not need to be replaced in some cases, as in `virtual machine` or `For production environments, a minimum of 3 physical machines is required as cluster members.` (You wouldn't say "a minimum of 3 cluster members is required as cluster members."

Let's use `cluster member` going forward for `node`, `member`, `host`, as well as `machine` if it makes sense. `member(s)` can be used if there's a clear antecedent (e.g. "You can use the `microcloud cluster` command to show information about the cluster members, or to remove specific **members**."). LXD's documentation already conforms to this, but MicroCloud's is inconsistent presently.

roosterfish: Thanks for the clarification, sounds good to me.

doc/how-to/initialise.md
minaelee: Do you mean it won't ask about remote storage either if you don't have any local disks?

doc/how-to/initialise.md
minaelee: I just noticed I need to fix `dm_crypt `to `dm-crypt` as well

doc/how-to/initialise.md
minaelee: Got it, I removed the word "local" so it's just "you will not be prompted to set up storage."

doc/how-to/initialise.md
minaelee: Fixed all above concerns

doc/.custom_wordlist.txt
```diff
@@ -1,45 +0,0 @@
-ACL
```
roosterfish: We tend to keep the history clean as there is no squashing performed to the commits when we merge the PR.

Can you please remove this file from the commit entirely as it's anyway added back in the commit after?

roosterfish (changes requested):
Looks good, just another small comment regarding the commit history.

doc/.custom_wordlist.txt
minaelee: Done :)

roosterfish (approved):

PR #498, opened 2024-11-15
GitHub: Let trivy scan the `2/stable` snap channel

roosterfish:
Now after we have released `2/stable` we can move the trivy scans that were previously tracking `2/candidate`.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -75,7 +75,7 @@ jobs:
         include:
           - channel: "3/edge"
             branch: "main"
-          - channel: "2/candidate"
+          - channel: "2/stable"
             branch: "v2-edge"
     steps:
       - name: Checkout


hamistao (approved):
Thanks for this!

I believe this won't change the workflow configurations as previously happened when using a matrix on it. Have you tried it on your fork to check? @roosterfish

roosterfish:
> Have you tried it on your fork to check? @roosterfish

No I haven't but I would assume it again adds another configuration as the jobs signature is different. 

Right now it says only `branch:main/channel:3/edge` in the title. I am actually interested what will happen now to narrow it down a bit more what is causing those different configs.

hamistao:
> Right now it says only branch:main/channel:3/edge in the title. I am actually interested what will happen now to narrow it down a bit more what is causing those different configs.

So am I. It is definetely sensitive to using a matrix strategy or not, but now we will find out if the values used in the matrix change the config.

PR #502, opened 2024-11-18
Set `user.microcloud` when bootstrapping LXD

masnax:
CC @edlerd @mas-who

### DIFF
--- a/test/suites/upgrade.sh
+++ b/test/suites/upgrade.sh
@@ -3,6 +3,22 @@
 test_upgrade() {
   reset_systems 4 2 3
 
+  # Clear out the debug binaries as we want to test upgrades from published versions.
+  for i in $(seq -f "%02g" 1 4) ; do
+    name="micro${i}"
+    if [ -n "${MICROCLOUD_DEBUG_PATH}" ] || [ -n "${MICROCLOUDD_DEBUG_PATH}" ]; then
+      lxc exec "${name}" -- rm -f /var/snap/microcloud/common/microcloud.debug || true
+      lxc exec "${name}" -- rm -f /var/snap/microcloud/common/microcloudd.debug || true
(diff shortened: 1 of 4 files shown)


test/suites/upgrade.sh
```diff
@@ -150,6 +150,7 @@ ovn:
     # Fourth upgrade MicroCloud.
     for m in micro01 micro02 micro03; do
       lxc exec "${m}" -- snap refresh microcloud --channel "${microcloud_target}"
+      set_debug_binaries "${m}"
```
roosterfish, line 153: Aren't those already copied to the respective location? Even though MicroCloud 1 snap is ignoring them the binaries should still have been copied to the `/var/snap/microcloud/common/` path and AFAIK the `snap refresh` isn't modifying this.

roosterfish (changes requested):
Looks good, only one comment to the test suite.

test/suites/upgrade.sh
roosterfish: Should get set [here](https://github.com/canonical/microcloud/blob/main/test/includes/microcloud.sh#L1277) already.

roosterfish: Yeah good point. Maybe adding a condition in the `setup_system` to only inject if MicroCloud >= 2? Same as with the `dm-crypt` stuff.

masnax: Agreed, except instead of >=2 it would be only inject if `/edge`, and skip it `/stable`.

masnax: Actually now that I'm thinking about it, silently having cases that skip the debug binary can be very destructive if you forget about it. 

Might just be better to Manually delete the debug binaries in the upgrade tests.

masnax:
I've also switched to use the version string instead of "true" as the value for the `user.microcloud` key.

roosterfish (approved):

PR #503, opened 2024-11-19
doc/tutorial: tutorial improvements

minaelee:
- fix typo of `micro01`..`04` to `micro1`..`4` in several places
- fix typo of `microcloud init` to `microcloud join` in output for joining members
- use `cluster member` when appropriate
- simplify & clarify text
- reduce parentheses and future tense
- note that VMs should not be used in production
- remove mention of needing minimum 3 VMs
- update screengrabs of the init and join processes

### DIFF
--- a/doc/tutorial/get_started.md
+++ b/doc/tutorial/get_started.md
@@ -267,9 +267,9 @@ Complete the following steps:
    1. Select `yes` to select more than one cluster member.
    1. As the address for MicroCloud's internal traffic, select the listed IPv4 address.
    1. Copy the session passphrase.
-   1. Head to the other servers (`micro02`, `micro03`, and `micro04`) and start the join process:
+   1. Head to the other servers (`micro2`, `micro3`, and `micro4`) and start the join process:
 
-          lxc exec micro02 microcloud join
+          lxc exec micro2 microcloud join
 
          ```{tip}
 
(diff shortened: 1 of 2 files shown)


doc/tutorial/get_started.md
```diff
@@ -267,48 +264,50 @@ Complete the following steps:
...
+       lxc exec micro2 microcloud join
 
-         ```{tip}
+   ```{tip}
+   Open three additional terminals to run the commands concurrently.
+   ```
 
-         Open up three additional terminals to run the commands concurrently.
-         ```
+   In each joining cluster member, select the address for MicroCloud's internal traffic that was chosen on `micro1`.
```
roosterfish, line 276: "from the same subnet that was chosen on `micro01`"?

roosterfish: Maybe I am wrong but the sentence sounds like you should select the same address for MicroCloud's internal traffic that was already chosen on `micro1`.
But in fact it's a different address in the same subnet.

minaelee: You are totally right. I updated it to use the same wording as on the initiating cluster member:
>In each joining cluster member, select the listed IPv4 address for MicroCloud's internal traffic.

I also added to the commit an updated screengrab of the join process that shows the address options, to help make it more clear. Thanks!

roosterfish: Thx!

roosterfish (approved):

PR #515, opened 2024-11-27
Interfaces: Allow selection of bridges when using network segregation

roosterfish:
This fixes a situation in which you have bridges configured on your systems with addresses that should be used either for Ceph public/internal or OVN underlay traffic. 

See this example netplan configuration:

```yaml
# netplan
network:
  version: 2
  bridges:
    br-ceph:
      interfaces: [enp5s0]
      addresses:
      - 1.2.3.4/24
      parameters:
        forward-delay: 0
        priority: 0
        stp: false

  ethernets:
    enp5s0: {}
```

Technically there isn't anything hindering you from selecting `br-ceph` as OVN uplink but also use it for Ceph public/internal traffic as well as OVN underlay.

Those bridges were falsely filtered out and used solely as candidates for OVN uplink interfaces but can now also be used for network segregation.

### DIFF
--- a/service/lxd.go
+++ b/service/lxd.go
@@ -557,18 +557,45 @@ func (s LXDService) GetNetworkInterfaces(ctx context.Context, name string, addre
 	uplinkInterfaces := map[string]api.Network{}
 	dedicatedInterfaces := map[string]DedicatedInterface{}
 	for _, network := range networks {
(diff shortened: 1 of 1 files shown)


masnax (approved):
LGTM!

PR #517, opened 2024-11-27
Test: Consistently use 2GiB for container root vol

roosterfish:
Prevent errors like https://github.com/canonical/microcloud/actions/runs/12029208869/job/33534087228?pr=469


roosterfish:
Those errors have started to appear in lots of places, always a bit of a different error but all related to no more space available, e.g. https://github.com/canonical/microcloud/actions/runs/12049316675/job/33599362221?pr=515.

Not sure if it's related to the recent runner switch but when launching containers in the test suite I think it's a good idea to always use `2GiB` for the root vol which is already done for some of the containers.

masnax (approved):

Issue #520, opened 2024-12-02
Aborted setup doesnt terminate `microcloud join` 


tomponline:
```
root@v1:~# snap list
Name        Version                 Rev    Tracking       Publisher   Notes
core22      20241001                1663   latest/stable  canonical✓  base
core24      20240920                609    latest/stable  canonical✓  base
lxd         5.21.2-084c8c8          31214  5.21/stable    canonical✓  -
microcloud  2.1.0-3e8b183           1144   2/stable       canonical✓  -
microovn    24.03.2+snapa2c59c105b  667    24.03/stable   canonical✓  -
snapd       2.66.1                  23258  latest/stable  canonical✓  snapd
```

```
root@v1:~# microcloud init
Waiting for services to start ...
Do you want to set up more than one cluster member? (yes/no) [default=yes]: 
Select an address for MicroCloud's internal traffic:

 Using address "10.19.164.17" for MicroCloud

MicroCeph not found. Continue anyway? (yes/no) [default=yes]: 
Use the following command on systems that you want to join the cluster:

 microcloud join

When requested enter the passphrase:

 oftentimes rupture saxophone buzzard

Verify the fingerprint "106a729df5d6" is displayed on joining systems.
Waiting to detect systems ...

 Selected "v2" at "10.19.164.247"
 Selected "v1" at "10.19.164.17"

Gathering system information ...
Would you like to set up local storage? (yes/no) [default=yes]: 
Select exactly one disk from each cluster member:

Select which disks to wipe:

 Using "/dev/disk/by-id/scsi-SQEMU_QEMU_HARDDISK_lxd_vol1" on "v1" for local storage pool
 Using "/dev/disk/by-id/scsi-SQEMU_QEMU_HARDDISK_lxd_vol1" on "v2" for local storage pool

Some systems are ineligible for distributed networking, which requires either an interface with no IPs assigned or a bridge. Continue anyway? (yes/no) [default=yes]: no
Error: User aborted
```

on joining member `microcloud join` is hanging:

```
root@v2:~# microcloud join
Waiting for services to start ...
Select an address for MicroCloud's internal traffic:

 Using address "10.19.164.247" for MicroCloud

Verify the fingerprint "f133cfdb303c" is displayed on the other system.
Specify the passphrase for joining the system: oftentimes rupture saxophone buzzard
Searching for an eligible system ...

 Found system "v1" at "10.19.164.17" using fingerprint "106a729df5d6"

[...]

PR #525, opened 2024-12-03
microcloud: terminate a session in joiner members when setup is aborted on initiator

gabrielmougard:
closes https://github.com/canonical/microcloud/issues/520


service/microcloud.go
```diff
@@ -374,3 +374,28 @@ func (s *CloudService) StartSession(ctx context.Context, role string, sessionTim
...
+		return err
+	}
+
+	client, err = cloudClient.UseAuthProxy(client, types.MicroCloud, cloudClient.AuthConfig{})
+	if err != nil {
+		return err
+	}
+
+	data := types.SessionStopPut{Reason: stopMsg}
+	err = client.Query(ctx, "PUT", types.APIVersion, api.NewURL().Path("session", "stop"), data, nil)
```
roosterfish, line 395: Can you please move this into `client/client.go` as it's done for `StartSession`.

api/session_stop.go
```diff
@@ -0,0 +1,52 @@
...
+	return rest.Endpoint{
+		AllowedBeforeInit: true,
+		Name:              "session/stop",
+		Path:              "session/stop",
+
+		Put: rest.EndpointAction{Handler: authHandlerMTLS(sh, sessionStopPut(sh))},
+	}
+}
+
+// sessionStopPut receives a stop request from an initiator member and stops the current session.
```
roosterfish, line 28: Theoretically the stop endpoint could also be invoked on the initiator so I would not be too specific in the comment and try to be generic by saying that is just stops the current session.

roosterfish (changes requested):
Thanks for the proposal. 

Could we also stop the session on the joiners if something doesn't go well in case of preseed?

The stop endpoint can be used to quit the session prematurely (with a reason/error), but may also be leveraged to notify the joiner when the creation of the MicroCloud is complete (graceful exit).  
I saw you have already extended the reverter to call the stop on every joiner in case of error but can we also extend this a bit to let the joiners wait until the initiator formed the cluster and is done?
IIRC there is already logic in the session stop for gracefully exiting out in case no reason is given.
In case of preseed we just let the joiner exit, see https://github.com/canonical/microcloud/blob/main/cmd/microcloud/preseed.go#L254.
In case of interactive we print a message here https://github.com/canonical/microcloud/blob/main/cmd/microcloud/ask.go#L1729 and let the joiner exit out.
It would be great if those could wait until the initiator is actually done setting up MicroCloud.

api/session_stop.go
gabrielmougard: fixed

service/microcloud.go
gabrielmougard: fixed

gabrielmougard:
@roosterfish it should be updated

client/client.go
```diff
@@ -44,6 +44,20 @@ func StartSession(ctx context.Context, c *client.Client, role string, sessionTim
 	return conn, nil
 }
 
+// StopJoinerSession is called from the initiator member to stop a joiner session.
```
roosterfish, line 47: Don't want to be picky but an initiator is only a member in case you grow the MicroCloud. Before the initial cloud is formed, there are only initiators and joiners but no members.

So I would suggest to just use initiator.

client/client.go
```diff
@@ -44,6 +44,20 @@ func StartSession(ctx context.Context, c *client.Client, role string, sessionTim
...
+func StopJoinerSession(ctx context.Context, remoteClient *client.Client, stopMsg string) error {
```
roosterfish, line 48: Can we please name this `c *client.Client` to stay consistent with the other funcs?

roosterfish: The func should rather be called `StopSession` as it calls the `session/stop` endpoint which doesn't have any notation of a joiner. You could potentially also call the endpoint on an initiator.

cmd/microcloud/main_init.go
```diff
@@ -214,13 +214,43 @@ func (c *initConfig) RunInteractive(cmd *cobra.Command, args []string) error {
...
 		}
+
+		reverter = revert.New()
+		defer reverter.Fail()
+
+		reverter.Add(func() {
+			// Stop each joiner member session.
+			cloud := s.Services[types.MicroCloud].(*service.CloudService)
+			for peer, system := range c.systems {
+				if system.ServerInfo.Name == "" || system.ServerInfo.Name == c.name {
```
roosterfish, line 233: Is there a situation in which `system.ServerInfo.Name` is empty? In my understanding this should not happen.
 
But I have seen this pattern now multiple times across the codebase, maybe @masnax can give some clarification on this :)

cmd/microcloud/main_init.go
```diff
@@ -214,13 +214,43 @@ func (c *initConfig) RunInteractive(cmd *cobra.Command, args []string) error {
...
+					continue
+				}
+
+				if system.ServerInfo.Address == "" {
```
roosterfish, line 237: Have you found a situation in which the address isn't provided? In any case it should not `return` but instead `continue` to try stopping the session on the other `system`s too.

cmd/microcloud/main_init.go
```diff
@@ -214,13 +214,43 @@ func (c *initConfig) RunInteractive(cmd *cobra.Command, args []string) error {
...
+					logger.Error("No joiner address provided to stop the session")
+					return
+				}
+
+				remoteClient, err := cloud.RemoteClient(system.ServerInfo.Certificate, util.CanonicalNetworkAddress(system.ServerInfo.Address, service.CloudPort))
+				if err != nil {
+					logger.Error("Failed to create remote client", logger.Ctx{"address": system.ServerInfo.Address, "error": err})
+					return
```
roosterfish, line 245: `continue`?

cmd/microcloud/main_init.go
```diff
@@ -214,13 +214,43 @@ func (c *initConfig) RunInteractive(cmd *cobra.Command, args []string) error {
...
+				}
+
+				err = cloudClient.StopJoinerSession(context.Background(), remoteClient, "Initiator cluster member aborted")
```
roosterfish, line 248: Can we rather send "Initiator aborted the setup"? See the comment on initiator/member from before.

roosterfish (changes requested):
Thanks for the updates, looking good, only a few more smaller comments to make it more robust.

client/client.go
gabrielmougard: fixed

cmd/microcloud/main_init.go
gabrielmougard: fixed

cmd/microcloud/main_init.go
gabrielmougard: fixed

roosterfish (approved):
LGTM! But I'll wait for Max to reply on https://github.com/canonical/microcloud/pull/525#discussion_r1914520989 so that we don't miss anything.

By the way can we have small test that checks the stop session behavior?

gabrielmougard:
> LGTM! But I'll wait for Max to reply on [#525 (comment)](https://github.com/canonical/microcloud/pull/525#discussion_r1914520989) so that we don't miss anything.
> 
> By the way can we have small test that checks the stop session behavior?

sure, I can add that :)

gabrielmougard:
@roosterfish test added

cmd/microcloud/main_init.go
masnax: I believe that prior to the secure key exchange mechanism, we would populate the `InitSystem` map with existing cluster members who did not have populated `ServerInfo` structs. as they were not detected during system lookup and are instead implicitly included in the setup (for example when growing MicroCloud or re-using an existing MicroCeph/MicroOVN into MicroCloud). 

I'm not sure if the changes made with the new lookup mechanisms make this unnecessary now.

roosterfish: AFAIK this hasn't been touched and the code tells this is still happening https://github.com/canonical/microcloud/blob/main/cmd/microcloud/add.go#L160. But there is always a name being set so I wonder if we can omit the first half of the condition.

@gabrielmougard have you checked stopping the session of joiners when growing the MicroCloud?

roosterfish:
> @roosterfish test added

Thanks, looking good. I like the idea of having no uplink ports to let the initiator exit.

cmd/microcloud/main_init.go
gabrielmougard: No. Should I add another test case for that ?

roosterfish: Would be great, it will catch any regression when we try to get rid of the `system.ServerInfo.Name == ""` check.

gabrielmougard: @roosterfish updated :)

roosterfish: The test seems to be failing. I have the suspicion that OVN networking is now skipped when adding the joiners as there wasn't an uplink interface when bootstrapping the single node.

roosterfish (approved):
Thanks, LGTM!

PR #527, opened 2024-12-06
cmd/microcloud: Do not iterate over a map for the questionnaire

masnax:
Fixes pipeline errors expecting the questions to always be in a preset order.


roosterfish (approved):
Thanks, nice find!

PR #540, opened 2024-12-10
Preseed: Always create remote pools across all members

roosterfish:
Fixes a situation in which the `remote` storage pool cannot be created on LXD because MicroCloud only prepares the storage pool on nodes that supply a disk.


masnax (approved):
LGTM! Although this made me realize there's a bug with reconciling the explicit disk definitions with the filtered disk definitions.

If you define explicit disks for each system, then the filter will attempt to search an empty list of storage resources, and return an error claiming that no disk could be found with the correct filter.

We should instead check that if all systems have specified a disk explicitly, that the filter is redundant and should not be used. However, if only some (or none) of systems have specified a disk explicitly, then the filter should only find disks for the unspecified systems.

PR #545, opened 2024-12-13
Fix storage filter restrictions

masnax:
The storage selection filtering mechanism in preseed is meant as an alternative to the explicitly defined disk set.

If all systems have supplied disks explicitly for ceph, then also using the filter will now return a descriptive error, instead of simply reporting that no more disks could be found. The same is true for zfs.

Disk filtering will apply for any system that did not explicitly define disks. That is to say, as long as one system has not explicitly defined disks, then including a filter will be valid. This is true for ceph/zfs respectively.


cmd/microcloud/preseed.go
```diff
@@ -903,31 +903,39 @@ func (p *Preseed) Parse(s *service.Handler, c *initConfig, installedServices map
...
-		if setupStorage {
-			continue
+		if len(system.Storage.Ceph) == 0 {
+			checkFilterCeph = true
 		}
+	}
 
-		cert := system.ServerInfo.Certificate
+	if !checkFilterCeph && len(p.Storage.Ceph) > 0 {
+		return nil, fmt.Errorf("Ceph disk filter cannot be used. All systems have explicitly specified disks")
```
roosterfish, line 922: I think it's also worth putting a comment in the reference preseed file in the docs to indicate you can either explicitly or implicitly specific disks.

cmd/microcloud/preseed.go
```diff
@@ -903,31 +903,39 @@ func (p *Preseed) Parse(s *service.Handler, c *initConfig, installedServices map
...
-		// Fetch system resources from LXD to find disks if we haven't directly set up disks.
-		allResources[peer], err = s.Services[types.LXD].(*service.LXDService).GetResources(context.Background(), peer, system.ServerInfo.Address, cert)
-		if err != nil {
-			return nil, fmt.Errorf("Failed to get system resources of peer %q: %w", peer, err)
+	if !checkFilterZFS && len(p.Storage.Local) > 0 {
+		return nil, fmt.Errorf("Local disk filter cannot be used. All systems have explicitly specified a disk")
+	}
+
+	allResources := map[string]*lxdAPI.Resources{}
+	if checkFilterZFS || checkFilterCeph {
```
roosterfish, line 930: If you only check here if any of `checkFilterZFS` or `checkFilterCeph` is `true` then you will also fetch disks for systems which have explicitly defined disks.

I assume that is the error you are seeing in the preseed suite.

roosterfish (changes requested):

roosterfish:
Merging the other preseed related PR has now caused a conflict. Requires rebase.

cmd/microcloud/preseed.go
masnax: Hmm, right, I've switched these to maps now, so we can selectively compile resources, and I've split the resource maps to be separate for ZFS and Ceph.

roosterfish (approved):
LGTM!

PR #546, opened 2024-12-13
Backport security fixes and dependency updates to v1 (v1-edge)

masnax:


roosterfish (commented):
Some of the commits don't allow tracing back the original change/commit. Can you use `cherry-pick -x` for them?

masnax:
> Some of the commits don't allow tracing back the original change/commit. Can you use `cherry-pick -x` for them?

Tracking with cherry picks wouldn't be all that easy here because we're taking bits and pieces that have been iterated on for a year. I can cherry pick the last commit that is "thematically similar" but since we're closing this branch by mid-year, I don't think it would be all that helpful.

roosterfish (approved):
Thanks, fine for me!

Issue #566, opened 2024-12-23
Device `cdrom` should not be listed as a deployment target


hd9:
When initializing microcloud, if CD-ROM is available, Microcloud suggests it as a viable deployment target, which doesn't make sense.

```
root@hpe: ~# microcloud init
Waiting for services to start ...
Do you want to set up more than one cluster member? (yes/no) [default=yes]: no
Using address "10.0.1.9" for MicroCloud
Gathering system information ...
Would you like to set up local storage? (yes/no) [default=yes]: 
Select exactly one disk from each cluster member:
Space to select; enter to confirm; type to filter results.
Up/down to move; right to select all; left to select none.
       +----------+-----------------------+-----------+-------+-----------------------------------------------------------+
       | LOCATION |         MODEL         | CAPACITY  | TYPE  |                           PATH                            |
       +----------+-----------------------+-----------+-------+-----------------------------------------------------------+
   [ ]  | hpe      | CT1000P3PSSD8         | 931.51GiB | nvme  | /dev/disk/by-id/nvme-eui.0000000000078f6d8a               |
>  [ ]  | hpe      | hp      DVD-RAM GH82N | 0B        | cdrom | /dev/disk/by-id/ata-hp_DVD-RAM_GH82N_234CD040074          |
       +----------+-----------------------+-----------+-------+-----------------------------------------------------------+
```

Ideally, only writeable devices should be listed in this screen.

roosterfish:
Hi, thanks for reporting this bug.

PR #576, opened 2025-01-07
Service: Exlude disks of type `cdrom`

roosterfish:
When creating local and remote storage pools the list of disks should not contain any cdrom drive.

Fixes https://github.com/canonical/microcloud/issues/566

### DIFF
--- a/service/system_information.go
+++ b/service/system_information.go
@@ -106,6 +106,11 @@ func (sh *Handler) CollectSystemInformation(ctx context.Context, connectInfo mul
 				continue
 			}
 
+			// Exclude cdrom drives as viable storage disk.
+			if disk.Type == "cdrom" {
+				continue
+			}
+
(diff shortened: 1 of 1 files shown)


roosterfish:
I was testing this with https://en.wikipedia.org/wiki/CDemu. My first though was just filter out read-only drivers but as you could potentially use the drive to write to the CD the rom drive itself isn't ro.

roosterfish:
@masnax it looks after `1/stable` got updated the preseed cmd fails (e.g. https://github.com/canonical/microcloud/actions/runs/12649784309/job/35247023464?pr=576). 

Did a bit of digging and it looks we are missing at least the first commit of the `latest-candidate` [cherry picks](https://github.com/canonical/microcloud-pkg-snap/blob/latest-candidate/snapcraft.yaml#L102) in the packaging repo in the `v1-candidate` branch.

masnax (approved):

PR #583, opened 2025-01-13
test/suites/upgrade: Use the respective service's candidate channels

roosterfish:
Found this whilst reviewing https://github.com/canonical/microcloud/pull/581.

To catch error more early in the process use the candidate channels instead of stable.


roosterfish:
Mh we should also check if the used MicroCloud snap channel variable is >= 2/edge which would still allow running the check for `user.microcloud` before we have `2.1.1`.

roosterfish:
@masnax ready.

masnax (approved):

PR #587, opened 2025-01-13
Test: check 1/candidate works

masnax:


roosterfish:
I have restarted the workflow on `main` after `1/candidate` got updated and its all green https://github.com/canonical/microcloud/commit/48e29e6a37d5e5ef31497954c33df978da79f80d. So I suspect the same outcome here :)

PR #602, opened 2025-02-02
doc: add how-to on safely shutting down a cluster member

minaelee:
- Add how-to on shutting down cluster members
- Add to nav sidebar
- Add link to "Work with MicroCloud" page

Thanks to @simondeziel for testing and the commands to override services start/stop sequence.

### DIFF
--- a/doc/how-to/shutdown_machine.md
+++ b/doc/how-to/shutdown_machine.md
@@ -1,0 +1,38 @@
+(howto-shutdown)=
+# How to shut down a machine
+
+## Stop all instances on the cluster member
+
+Before you can shut down a machine that is a MicroCloud cluster member, first stop all LXD instances hosted on it:
+
+```
+lxc stop --all
+```
+
+## Enforce services shutdown and restart order
+
+During the shutdown process of a MicroCloud cluster member, the LXD service must stop _before_ the MicroCeph and MicroOVN services. At restart, the LXD service must start _after_ MicroCeph and MicroOVN. This order ensures that LXD does not run into issues due to unavailable storage or networking services.
+
+To enforce this shutdown and restart order, create a configuration file in each cluster member's `/etc/systemd/system/snap.lxd.daemon.service.d` directory to override the behaviour of `snap.lxd.daemon.service`. To simplify creating the directory and configuration file, you can copy and paste the following commands into each cluster member:
+
+```
+# Create the directory if it doesn't exist
+sudo mkdir -p /etc/systemd/system/snap.lxd.daemon.service.d
+
+# Create the configuration file
+cat << EOF | sudo tee /etc/systemd/system/snap.lxd.daemon.service.d/lxd-shutdown.conf
+# Makes sure the LXD daemon stops before Ceph/OVN and restarts after Ceph/OVN
+[Unit]
+After=snap.microceph.daemon.service
+After=snap.microovn.daemon.service
+EOF
+
+# Reload systemd daemon
+sudo systemctl daemon-reload
+```
+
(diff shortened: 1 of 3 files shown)


doc/how-to/shutdown_machine.md
```diff
@@ -0,0 +1,38 @@
+(howto-shutdown)=
+# How to shut down a machine
+
+## Stop all instances on the cluster member
+
+Before you can shut down a machine that is a MicroCloud cluster member, first stop all LXD instances hosted on it:
+
+```
+lxc stop --all
```
roosterfish, line 9: Worth mentioning you can also perform `lxc move <instance> --target <other cluster member>` to migrate the instances away from the member instead of just shutting them down?

roosterfish (changes requested):

doc/how-to/shutdown_machine.md
```diff
@@ -0,0 +1,38 @@
...
+During the shutdown process of a MicroCloud cluster member, the LXD service must stop _before_ the MicroCeph and MicroOVN services. At restart, the LXD service must start _after_ MicroCeph and MicroOVN. This order ensures that LXD does not run into issues due to unavailable storage or networking services.
+
+To enforce this shutdown and restart order, create a configuration file in each cluster member's `/etc/systemd/system/snap.lxd.daemon.service.d` directory to override the behaviour of `snap.lxd.daemon.service`. To simplify creating the directory and configuration file, you can copy and paste the following commands into each cluster member:
+
+```
+# Create the directory if it doesn't exist
+sudo mkdir -p /etc/systemd/system/snap.lxd.daemon.service.d
+
+# Create the configuration file
+cat << EOF | sudo tee /etc/systemd/system/snap.lxd.daemon.service.d/lxd-shutdown.conf
```
roosterfish, line 23: @tomponline @simondeziel can we ship such a file by default in the LXD snap?

simondeziel: Thanks for asking, this made me check my assumption and turns out we do have the ability to set the `After=` clause according to https://snapcraft.io/docs/services-and-daemons.

I'll give it a try and will report back.

simondeziel: Sigh, too much enthusiasm it seems as I didn't read till the end:
> after An ordered list of applications the daemon is to be started after. **Applications must be part of the same snap.**

simondeziel: Thinking some more, we could possibly use LXD's privs to inject the workaround at runtime.

doc/how-to/shutdown_machine.md
minaelee: Definitely. I updated, thanks!

doc/how-to/shutdown_machine.md
roosterfish: That sounds great, thanks for checking this! In the meantime should we still merge this "workaround"?

roosterfish (approved):

doc/how-to/shutdown_machine.md
simondeziel: Yes please.

simondeziel: https://github.com/canonical/lxd-pkg-snap/pull/710

simondeziel: > In the meantime should we still merge this "workaround"?

https://github.com/canonical/lxd-pkg-snap/pull/710 is for `latest/edge` and this will need to be backported to `5.21/stable` before we can think about removing the doc workaround.

Issue #603, opened 2025-02-03
Ordering of "Commencing cluster join of the remaining services" not consistent


tomponline:
first joiner:

```
Commencing cluster join of the remaining services (LXD, MicroOVN, MicroCeph)
```

second joiner

```
Commencing cluster join of the remaining services (LXD, MicroCeph, MicroOVN)
```

Looks like there's a map being iterated in random order somewhere.

roosterfish:
Ty for reporting, fixed in https://github.com/canonical/microcloud/pull/604

PR #604, opened 2025-02-03
Cmd: Sort remaining services in joiners message

roosterfish:
Fixes https://github.com/canonical/microcloud/issues/603

Fixes a situation in which multiple joiners might see a list of services appear using a different ordering.

### DIFF
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -1685,6 +1685,8 @@ func (c *initConfig) askJoinConfirmation(gw *cloudClient.WebsocketGateway, servi
 	}
 
 	if len(servicesStr) > 0 {
+		slices.Sort(servicesStr)
+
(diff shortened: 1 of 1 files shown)


cmd/microcloud/ask.go
```diff
@@ -1675,20 +1675,25 @@ func (c *initConfig) askJoinConfirmation(gw *cloudClient.WebsocketGateway, servi
...
+		servicesStr = append(servicesStr, string(serviceType))
 	}
 
-	if len(newServices) > 0 {
-		var servicesStr []string
-		for service := range newServices {
-			servicesStr = append(servicesStr, string(service))
-		}
+	if len(servicesStr) > 0 {
+		slices.SortFunc(servicesStr, func(a string, b string) int {
```
tomponline, line 1688: https://pkg.go.dev/sort#Strings ?

roosterfish: Good point, much more convenient. It calls `slices.Sort` under the hood so I'll go with this one instead.

tomponline (approved):

PR #504, opened 2024-11-19
Service: Set corret name of `instances_migration_stateful` extension (v2-edge)

roosterfish:
Signed-off-by: Julian Pelizäus <julian.pelizaeus@canonical.com>
(cherry picked from commit 2f4342c98f4470bd27698fbeed8c5899d7f4e44a)

PR #609, opened 2025-02-05
doc: keep instances running during update/upgrade

minaelee:
This PR adds a section in the [How to update and upgrade](https://canonical-microcloud.readthedocs-hosted.com/en/latest/microcloud/how-to/update_upgrade/) docs page to explain that instances can be left running during and update/upgrade but the storage/network services might briefly be affected. It also explains how to use cluster evacuate to migrate instances away before an update/upgrade, including live migration for pre-configured VMs, and how to use cluster restore to migrate them back afterward.

### DIFF
--- a/doc/how-to/update_upgrade.md
+++ b/doc/how-to/update_upgrade.md
@@ -14,6 +14,21 @@ See the following backup guides for each of the snaps:
 
 In case of error, see {ref}`howto-recover` for troubleshooting details.
 
+(howto-update-upgrade-running-instances)=
+## Keep instances running during an update or upgrade
+
+You can update or upgrade MicroCloud and its dependency snaps on a cluster member while it is hosting running instances. However, when updating or upgrading the MicroCeph or MicroOVN snaps, storage and network services on that member might be briefly affected. For example, the instances might experience temporary packet loss.
+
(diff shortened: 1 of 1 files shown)


doc/how-to/update_upgrade.md
```diff
@@ -14,6 +14,21 @@ See the following backup guides for each of the snaps:
...
 
+(howto-update-upgrade-running-instances)=
+## Keep instances running during an update or upgrade
+
+You can update or upgrade MicroCloud and its dependency snaps on a cluster member while it is hosting running instances. However, when updating or upgrading the MicroCeph or MicroOVN snaps, storage and network services on that member might be briefly affected. For example, the instances might experience temporary packet loss.
+
+To ensure that the update or upgrade does not affect workloads on running instances at all, use the following live-migration approach:
+
+- Use virtual machines (VMs) instead of system containers for crucial workloads; in general, containers cannot be live-migrated.
+- Each VM must be pre-configured for live migration. See: {ref}`lxd:live-migration-vms` for information on the required configurations.
```
roosterfish, line 25: We have [fixed](https://github.com/canonical/microcloud/pull/504) this already in `v2-edge` and it should be configured by default starting with the next MicroCloud release if you create a new cluster.
But for now I think it makes sense to explicitly document this.

roosterfish (approved):

PR #630, opened 2025-02-10
Add support for connection to cluster manager

edlerd:
# Done
- extend database with remotes and remote-configuration tables
- add API endpoints to configure, show and delete cluster manager configuration
- add CLI commands for the new API endpoints
- add go routine that collects info from LXD and regularly sends it to cluster manager

Open issues:
- retry logic for sending updates to cluster manager
- make the sending of updates resilient to timeouts or downtimes


database/remotes.mapper.go
```diff
@@ -0,0 +1,328 @@
+package database
+
+// The code below was generated by lxd-generate - DO NOT EDIT!
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"github.com/canonical/microcluster/v2/cluster"
```
edlerd, line 10: Manually fixed this one import. Any hints how to `make update-schema` create the file with the right import directly?

database/remotes_table.go
```diff
@@ -0,0 +1,44 @@
...
+//go:generate mapper stmt -d github.com/canonical/microcluster/v2/cluster -e Remote update table=remotes
+//
+//go:generate mapper method -i -d github.com/canonical/microcluster/v2/cluster -e Remote GetMany table=remotes
+//go:generate mapper method -i -d github.com/canonical/microcluster/v2/cluster -e Remote GetOne table=remotes
+//go:generate mapper method -i -d github.com/canonical/microcluster/v2/cluster -e Remote ID table=remotes
+//go:generate mapper method -i -d github.com/canonical/microcluster/v2/cluster -e Remote Exists table=remotes
+//go:generate mapper method -i -d github.com/canonical/microcluster/v2/cluster -e Remote Create table=remotes
+//go:generate mapper method -i -d github.com/canonical/microcluster/v2/cluster -e Remote Update table=remotes
+//go:generate mapper method -i -d github.com/canonical/microcluster/v2/cluster -e Remote DeleteOne-by-Name table=remotes
+
```
edlerd, line 20: I tried to also create queries for the `RemoteConfigTable`, but adding
```
//go:generate mapper stmt -d github.com/canonical/microcluster/v2/cluster -e RemoteConfigTable objects table=remotes_config
```
here causes a segfaul in `make update-schema`:
```
go generate ./...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x75a7d3]

goroutine 1 [running]:
github.com/canonical/lxd/lxd/db/generate/db.(*Field).IsScalar(...)
        /home/edlerd/dev/open-source/lxd/lxd/db/generate/db/mapping.go:275
github.com/canonical/lxd/lxd/db/generate/db.(*Field).OrderBy(0x0, 0xc0000cf4a0, {0x7ffc38407a4e, 0xe})
        /home/edlerd/dev/open-source/lxd/lxd/db/generate/db/mapping.go:363 +0x133
github.com/canonical/lxd/lxd/db/generate/db.(*Stmt).objects(0xc0002e7040, 0xc0000b29f0)
        /home/edlerd/dev/open-source/lxd/lxd/db/generate/db/stmt.go:143 +0x530
github.com/canonical/lxd/lxd/db/generate/db.(*Stmt).Generate(0xc0002e7040, 0xc0000b29f0)
        /home/edlerd/dev/open-source/lxd/lxd/db/generate/db/stmt.go:84 +0x165
github.com/canonical/lxd/lxd/db/generate/file.Append({0x7ffc38407a2e?, 0x2c?}, {0x7ffc384079e4, 0x11}, {0x9edfe8, 0xc0002e7040}, 0x40?)
        /home/edlerd/dev/open-source/lxd/lxd/db/generate/file/write.go:85 +0xf4
main.newDbMapperStmt.func1(0xc000150400?, {0xc000142d80?, 0x4?, 0x90caea?})
        /home/edlerd/dev/open-source/lxd/lxd/db/generate/db.go:110 +0x106
github.com/spf13/cobra.(*Command).execute(0xc00019ef08, {0xc000142d00, 0x8, 0x8})
        /home/edlerd/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:985 +0xaca
github.com/spf13/cobra.(*Command).ExecuteC(0xc00019e008)
        /home/edlerd/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117 +0x3ff
github.com/spf13/cobra.(*Command).Execute(...)
        /home/edlerd/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041
main.main()
        /home/edlerd/dev/open-source/lxd/lxd/db/generate/main.go:9 +0x19
database/remotes_table.go:21: running "lxd-generate": exit status 2
make: *** [Makefile:91: update-schema] Error 1
```

cmd/microcloudd/main.go
```diff
@@ -229,6 +233,12 @@ func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
 					logger.Error("Failed to update LXD configuration on start", logger.Ctx{"error": err})
 				}
 
+				// todo: is this the right way to trigger the update?
+				go func() {
+					api.SendClusterManagerStatusMessageTask(s, state)
+					time.Sleep(30 * time.Second)
+				}()
+
```
edlerd, line 241: Should this be done with a different approach, or is this OK here?

roosterfish:
The code tests seem to be failing due to https://github.com/canonical/microcloud/pull/677.
Potential fix in https://github.com/canonical/microcloud/pull/679.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+		result.CPULoad5 = fmt.Sprintf("%.2f", cpuLoad5)
+		result.CPULoad15 = fmt.Sprintf("%.2f", cpuLoad15)
+	}
+
+	return nil
+}
+
+// SendClusterManagerStatusMessageTask starts a dedicated go routine, that sends the cluster manager status message.
+func SendClusterManagerStatusMessageTask(ctx context.Context, sh *service.Handler, s state.State) {
+	go func(ctx context.Context, sh *service.Handler, s state.State) {
```
roosterfish, line 667: It's important to close the `for` loop when the outer context gets cancelled. Therefore I would highly recommend a ticker with channel:

```go
d := new time.NewTicker(60*time.Second)
defer ticker.Stop()

for {
    select {
        case <-d.C:
           // send, you can also do d.Reset(<new duration>)
        case <-ctx.Done():
           return // exit the loop and close the go routine
    }
}
```

api/services_proxy.go
```diff
@@ -113,6 +114,31 @@ func lxdHandler(s state.State, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	// Special case for metrics endpoint, that is not responding with JSON.
+	if r.URL.Path == "/1.0/metrics" {
+		return response.ManualResponse(func(w http.ResponseWriter) error {
+			defer resp.Body.Close()
```
roosterfish, line 120: Isn't the body closed automatically?

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,366 @@
...
+
+	"github.com/canonical/microcloud/microcloud/api/types"
+	"github.com/canonical/microcloud/microcloud/database"
+)
+
+type cmdClusterManger struct {
+	common *CmdControl
+}
+
+func (c *cmdClusterManger) Command() *cobra.Command {
```
roosterfish, line 21: Missing doc string.

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,366 @@
...
+	cmd := &cobra.Command{}
+	cmd.Use = "cluster-manager"
+	cmd.Short = "Manage cluster-manager connection"
+
+	// Join
+	clusterManagerJoinCmd := cmdClusterManagerJoin{common: c.common, alias: c}
+	cmd.AddCommand(clusterManagerJoinCmd.command())
+
+	// Show
+	clusterManagerShowCmd := cmdClusterManagerShow{common: c.common, alias: c}
```
roosterfish, line 31: Would be great to have a "last status" field or something equivalent to check if the last update interval succeeded or if it returned an error.

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,366 @@
...
+
+func (c *cmdClusterManagerJoin) run(_ *cobra.Command, args []string) error {
+	if len(args) != 1 {
+		return fmt.Errorf("Expected exactly one argument with the token")
+	}
+
+	token := args[0]
+
+	if token == "" {
+		return fmt.Errorf("Missing token")
```
roosterfish, line 81: `Token cannot be empty` ?

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,366 @@
...
+		Token: token,
+	}
+
+	apiClient, err := getApiClient(c.common)
+	if err != nil {
+		return err
+	}
+
+	var clusterManager *types.ClusterManager
+	err = apiClient.QueryStruct(context.Background(), "POST", "1.0", api.NewURL().Path("cluster-manager"), payload, &clusterManager)
```
roosterfish, line 94: Can you use those client funcs also for the HTTP requests done by the cluster manager client?

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,366 @@
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Successfully joined cluster manager." + "\n")
```
roosterfish, line 99: `Println()` saves you from doing the `\n` at the end.

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+
+	var clusterManager *types.ClusterManager
+	err = apiClient.QueryStruct(context.Background(), "GET", "1.0", api.NewURL().Path("cluster-manager"), nil, &clusterManager)
+	if err != nil {
+		return err
+	}
+
+	if len(clusterManager.Addresses) == 0 {
+		fmt.Printf("cluster manager not configured" + "\n")
```
roosterfish, line 134: `Println()`

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,366 @@
...
+		return nil
+	}
+
+	fmt.Printf(
+		"cluster manager configuration:\n addresses"+": %s\n fingerprint"+": %s\n update-interval-seconds"+": %s\n",
```
roosterfish, line 139: Can we rather output valid YAML? IMHO makes it easier for parsing. So having a top level `config:` (as in LXD) or output a struct which is marshalled by the yaml package itself.

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,366 @@
...
+		return err
+	}
+
+	var clusterManager *database.ClusterManager
+	err = apiClient.QueryStruct(context.Background(), "DELETE", "1.0", api.NewURL().Path("cluster-manager"), nil, &clusterManager)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("cluster manager configuration cleared" + "\n")
```
roosterfish, line 177: `Println()`

roosterfish: We tend to always start messages with uppercase.

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,366 @@
...
+// Set.
+type cmdClusterManagerSet struct {
+	common *CmdControl
+	alias  *cmdClusterManger
+}
+
+func (c *cmdClusterManagerSet) command() *cobra.Command {
+	cmd := &cobra.Command{}
+	cmd.Use = "set"
+	cmd.Short = "Set specific cluster manager configuration key. Available keys are: addresses, fingerprint and update-interval-seconds."
```
roosterfish, line 245: I would rather put this in some documentation then having to list all of the config keys in the description.

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,366 @@
...
+	cmd.Example = `lxc cluster-manager set addresses example.com:8443`
+
+	cmd.RunE = c.run
+
+	return cmd
+}
+
+func (c *cmdClusterManagerSet) run(_ *cobra.Command, args []string) error {
```
roosterfish, line 253: In LXD you can set multiple config keys at once. Think that would also make sense in this case.

cmd/microcloudd/main.go
```diff
@@ -175,6 +179,8 @@ func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
 				return setHandlerAddress(state.Address().URL.Host)
 			},
 			OnStart: func(ctx context.Context, state state.State) error {
+				api.SendClusterManagerStatusMessageTask(ctx, s, state)
```
roosterfish, line 182: Microcluster cancels this context during shutdown which will then cause the go routine (commented above) to exit cleanly.

database/cluster_manager_schema.go
```diff
@@ -0,0 +1,45 @@
...
+	"database/sql"
+
+	"github.com/canonical/lxd/lxd/db/schema"
+)
+
+// SchemaExtensions is a list of schema extensions that can be passed to the MicroCluster daemon.
+// Each entry will increase the database schema version by one, and will be applied after internal schema updates.
+var SchemaExtensions = []schema.Update{
+	clusterManagerTable,
+	clusterManagerConfigTable,
```
roosterfish, line 15: We can do this in a single schema update. This helps to reason that both changes belong together.

database/cluster_manager_schema.go
```diff
@@ -0,0 +1,45 @@
...
+	return err
+}
+
+func clusterManagerConfigTable(ctx context.Context, tx *sql.Tx) error {
+	stmt := `
+CREATE TABLE cluster_manager_config (
+    id                  INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+    cluster_manager_id  INTEGER NOT NULL,
+    field               TEXT NOT NULL,
+    value               TEXT NOT NULL
```
roosterfish, line 38: Please add a foreign key relation to the `cluster_manager` table.
It should also have a delete on cascade to it will get rid of the config keys when deleting the referenced cluster manager.

This also makes the removal of the cluster manager easier as you don't have to manually cleanup the config keys to. I saw this in the function above.

roosterfish (changes requested):
Thanks for this! I was able to deploy the cluster manager using the steps provided in the docs, had some trouble with authentication but managed to get in.

Connecting MicroCloud and the manager worked out of the box! I have left some comments below in regards to the overall structure and design.

Something I couldn't debug was that my MicroCloud seemed to never send a status after the initial join even though the link to cluster manager was still setup. 
In such situations it would be very useful to be able to check the status of the last attempt. Left a comment for this.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+	"github.com/canonical/lxd/shared/version"
+	"github.com/canonical/microcluster/v2/rest"
+	"github.com/canonical/microcluster/v2/state"
+
+	"github.com/canonical/microcloud/microcloud/api/types"
+	"github.com/canonical/microcloud/microcloud/database"
+	"github.com/canonical/microcloud/microcloud/service"
+)
+
+var updateIntervalField = "UpdateInterval"
```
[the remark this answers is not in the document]
edlerd: yes, agreed.

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,366 @@
...
+	return cmd
+}
+
+// Join.
+type cmdClusterManagerJoin struct {
+	common *CmdControl
+	alias  *cmdClusterManger
+}
+
+func (c *cmdClusterManagerJoin) command() *cobra.Command {
```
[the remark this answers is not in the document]
edlerd: True this is a flaw. Changing the logic to
1. ensure cluster manager is not configured
2. send the post request to cluster manger
3. store config in the local database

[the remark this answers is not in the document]
edlerd: This has a benefit of first ensuring the token is valid by sending the request to cluster manager.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+
+// ClusterManagerCmd represents the manage cluster manager configuration.
+var ClusterManagerCmd = func(sh *service.Handler) rest.Endpoint {
+	return rest.Endpoint{
+		Path: "cluster-manager",
```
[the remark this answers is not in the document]
edlerd: The tables do not have a name, only an ID. I think we shouldn't expose the internal id on the endpoint level.

We don't need a name at this point, only an ID, should we introduce a name just for the endpoint extensibility? Seems like overly complicating it to me. Wdyt?

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+		return response.InternalError(err)
+	}
+
+	if clusterManager.Addresses == "" {
+		return response.SyncResponse(true, types.ClusterManager{})
+	}
+
+	var updateInterval string
+	if len(updateIntervalConfig) > 0 {
+		updateInterval = updateIntervalConfig[0].Value
```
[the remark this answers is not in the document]
edlerd: I think `ClusterManagerConfig` is the public API structure. It should be explicit and list all supported keys. We flatten out the structure to the outside world. Internally, we store config as an arbitrary list, which seems fine, because it is in our control. But we should not expose the config as an arbitrary list. it might contain internal data in the future, so keeping that controlled is beneficial.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+func clusterManagerPost(sh *service.Handler) func(state state.State, r *http.Request) response.Response {
+	return func(state state.State, r *http.Request) response.Response {
+		args := types.ClusterManagerPost{}
+		err := json.NewDecoder(r.Body).Decode(&args)
+		if err != nil {
+			return response.BadRequest(err)
+		}
+
+		if args.Token == "" {
+			return response.BadRequest(fmt.Errorf("No token provided"))
```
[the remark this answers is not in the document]
edlerd: Good hint, updating also other occurrences to use it.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+			Fingerprint: joinToken.Fingerprint,
+		}
+
+		updateIntervalConfig := database.ClusterManagerConfig{
+			Field: updateIntervalField,
+			Value: "60",
+		}
+
+		err = state.Database().Transaction(context.Background(), func(ctx context.Context, tx *sql.Tx) error {
+			existingId, err := loadClusterManagerId(state)
```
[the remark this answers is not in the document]
edlerd: Returning a list complicates the callers. We need to distinguish the cases of 1. no value configured 2. some value configured.

We could just throw an error in `loadClusterManagerId` if it is not configured, instead of checking for -1.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+	url := "https://" + address + "/1.0/remote-cluster"
+	req, err := http.NewRequest("POST", url, bytes.NewReader(reqBody))
+	if err != nil {
+		return err
+	}
+
+	// Sign the payload with a hmac, using the secret from the join token.
+	mac := hmac.New(sha256.New, []byte(joinToken.Secret))
+	mac.Write(reqBody)
+	req.Header.Set("X-CLUSTER-SIGNATURE", base64.StdEncoding.EncodeToString(mac.Sum(nil)))
```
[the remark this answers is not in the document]
edlerd: sure, adopted those ones.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+
+	content := new(bytes.Buffer)
```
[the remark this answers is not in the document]
edlerd: This is only for more verbose error reporting. I'll move this into the error code branch.

api/types/cluster_manager.go
```diff
@@ -0,0 +1,23 @@
...
+	// Example: example.com:8443
+	Addresses []string `json:"addresses" yaml:"addresses"`
+
+	// Fingerprint of the cluster manager server certificate
+	// Example: 90fedb21cda5ac6a45a878c151e6ac8fe16b4b723e44669fd113e4ea07597d83
+	Fingerprint *string `json:"fingerprint" yaml:"fingerprint"`
+
+	// Interval in seconds to send status messages to the cluster manager
+	// Example: 60
+	UpdateInterval *string `json:"update_interval" yaml:"update_interval"`
```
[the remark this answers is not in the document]
edlerd: It is to distinguish the "field is empty" and "field is not set" cases. We need to distinguish those for the put endpoint and when using the `microcloud cluster-manager set` commands. A user might set a field to empty string or set another field. With pointers, we can nicely distinguish the two cases.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+	var address string
+	var remoteCert *x509.Certificate
+	var err error
+
+	if len(addresses) == 0 {
+		return nil, "", "", fmt.Errorf("No cluster manager addresses provided.")
+	}
+
+	// fetch remote cert and pick the first address that responds without error
+	for _, address = range addresses {
```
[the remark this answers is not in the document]
edlerd: Just because an address is up on join, doesn't mean on a later status request time, which might be minutes, hours or days later, the target address is still up.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+
+	// verify remote cert
+	remoteFingerprint := shared.CertFingerprint(remoteCert)
+	if !strings.EqualFold(remoteFingerprint, expectedFingerprint) {
+		return nil, "", "", fmt.Errorf("Invalid cluster manager certificate fingerprint, expected %s, got %s", expectedFingerprint, remoteFingerprint)
+	}
+
+	// get local cert
+	cloud := sh.Services[types.MicroCloud].(*service.CloudService)
+	localCert, err := cloud.ServerCert()
```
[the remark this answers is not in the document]
edlerd: Good catch! Adjusted this bit.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+	CPUTotalCount     int64                `json:"cpu_total_count"`
+	CPULoad1          string               `json:"cpu_load_1"`
+	CPULoad5          string               `json:"cpu_load_5"`
+	CPULoad15         string               `json:"cpu_load_15"`
+	MemoryTotalAmount int64                `json:"memory_total_amount"`
+	MemoryUsage       int64                `json:"memory_usage"`
+	DiskTotalSize     int64                `json:"disk_total_size"`
+	DiskUsage         int64                `json:"disk_usage"`
+	MemberStatuses    []StatusDistribution `json:"member_statuses"`
+	InstanceStatuses  []StatusDistribution `json:"instance_status"`
```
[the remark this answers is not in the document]
edlerd: I don't understand what you mean here? Can you please clarify?

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+	Metrics           string               `json:"metrics"`
+	UiUrl             string               `json:"ui_url"`
+}
+
+func sendClusterManagerStatusMessage(ctx context.Context, sh *service.Handler, s state.State) time.Duration {
+	logger.Debug("Running sendClusterManagerStatusMessage")
+	var nextUpdate time.Duration = 0
+
+	cloud := sh.Services[types.MicroCloud].(*service.CloudService)
+	isInitialized, err := cloud.IsInitialized(context.Background())
```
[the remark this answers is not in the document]
edlerd: ah yes, we can use that one.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+	return nil
+}
+
+func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *ClusterManagerStatusPost) error {
+	lxdMembers, err := lxdClient.GetClusterMembers()
+	if err != nil {
+		return fmt.Errorf("Failed to get LXD cluster members: %w", err)
+	}
+
+	result.UiUrl = lxdMembers[0].URL
```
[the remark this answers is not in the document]
edlerd: good idea.

database/cluster_manager_schema.go
edlerd: compacted to a single function.

cmd/microcloud/cluster_manager.go
edlerd: yes, changing this to yaml format.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+	}
+
+	url := "https://" + address + "/1.0/remote-cluster"
+	req, err := http.NewRequest("POST", url, bytes.NewReader(reqBody))
+	if err != nil {
+		return err
+	}
+
+	// Sign the payload with a hmac, using the secret from the join token.
+	mac := hmac.New(sha256.New, []byte(joinToken.Secret))
```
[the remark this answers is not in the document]
edlerd: adjusted to the tools, also updated cluster manager in https://github.com/canonical/microcloud-cluster-manager/pull/165

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+	}
+
+	logger.Debug("Done sending status message to cluster manager")
+	return nextUpdate
+}
+
+func enrichInstanceMetrics(lxdClient lxd.InstanceServer, result *ClusterManagerStatusPost) error {
+	instanceFrequencies := make(map[string]int64)
+
+	instanceList, err := lxdClient.GetInstances(api.InstanceTypeAny)
```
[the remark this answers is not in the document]
edlerd: good point, we should use `GetInstancesAllProjects` instead.

api/cluster_manager.go
edlerd: I think this is slightly simpler:
```
			if ctx.Done() != nil {
				return
			}
```
We don't want to send every 60 seconds, gathering data, etc might take time. So I think a ticker is a bit different as it doesn't account for the time taken collecting and sending data.

database/cluster_manager_schema.go
edlerd: Ah right, good idea!

api/services_proxy.go
edlerd: I think this is always needed, see [this discussion](https://forum.golangbridge.org/t/is-it-necessary-to-close-request-body/37519).

cmd/microcloud/cluster_manager.go
edlerd: We can store a lastSuccessTime and lastErrorTime in the config table, but not sure if that is a good idea.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+		if err != nil {
+			return response.InternalError(err)
+		}
+
+		return response.SyncResponse(true, nil)
+	}
+}
+
+func sendRequestToClusterManager(sh *service.Handler, joinToken *api.ClusterMemberJoinToken) error {
+	client, publicKey, address, err := NewClusterManagerClient(sh, joinToken.Addresses, joinToken.Fingerprint)
```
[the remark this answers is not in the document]
edlerd: Good idea, extracted a client file with the three methods we use.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+			return response.InternalError(err)
+		}
+
+		url := "https://" + address + "/1.0/remote-cluster"
+		req, err := http.NewRequest("DELETE", url, nil)
+		if err != nil {
+			return response.InternalError(err)
+		}
+
+		resp, err := client.Do(req)
```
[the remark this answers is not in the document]
edlerd: moved to a reusable function and dedicated client file

cmd/microcloud/cluster_manager.go
edlerd: I think it is a UX improvement to directly show it on the CLI.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return clusterManager, updateIntervalConfig, nil
+}
+
+// StatusDistribution represents the status distribution of items.
+type StatusDistribution struct {
```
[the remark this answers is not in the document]
edlerd: Again not sure, this is types for communicating with cluster manager, building the payload for it.

api/cluster_manager.go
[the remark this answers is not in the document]
edlerd: I misunderstood you before. Yes, this picking an address should be closer to sending the request. With moving the logic into a dedicated client file, this should be resolved.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
...
+			}
+
+			clusterManagerId, err := database.CreateClusterManager(ctx, tx, clusterManager)
+			if err != nil {
+				return err
+			}
+
+			updateIntervalConfig.ClusterManagerID = clusterManagerId
+
+			_, err = database.CreateClusterManagerConfig(ctx, tx, updateIntervalConfig)
```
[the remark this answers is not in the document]
edlerd: This and some other comments below on the name -- I would rather keep as is. I think supporting a name is complicating things today for a future extension we might never need.

cmd/microcloud/cluster_manager.go
edlerd: Not sure if that is really needed. Maybe keep it simple?

cmd/microcloud/cluster_manager.go
edlerd: Not sure what you mean here. The cluster manager client can't use the functions, they will connect to microcloud. But it needs to connect to a remote cluster manager server.

roosterfish:
You can rebase this one to get the linter fix.

edlerd:
All issues mentioned above should be resolved. @roosterfish please give this another pass of review.

edlerd:
Relying now on a simpler token verification without needing the hmac. Tested locally and is ready for another review.

roosterfish:
Based on the previous discussion to enable this via a feature flag, I think what we can proceed doing is using the `snap set` command to set/unset a specific feature flag for cluster manager:

`snap set microcloud feature.cluster-manager=true`

Based on this, the snap can then export an environment variable e.g. `FEATURE_CLUSTER_MANAGER=true` which the `microcloud` CLI can interpret and which will lead to adding the `cluster-manager` subcommand.
I think it's fine always have the API endpoints exposed whether or not the feature flag is set/unset but at least the CLI will hide it if not activated.
This is how `daemon.debug` is set in LXD https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft/commands/daemon.start#L724.
We would require a `configure` hook which sets the value https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft/hooks/configure#L46.

roosterfish:
> Based on the previous discussion to enable this via a feature flag, I think what we can proceed doing is using the `snap set` command to set/unset a specific feature flag for cluster manager:

Meaning we would make `app.AddCommand(cmdClusterManager.Command())` conditional whether or not `FEATURE_CLUSTER_MANAGER=true` is set.

edlerd:
> Meaning we would make `app.AddCommand(cmdClusterManager.Command())` conditional whether or not `FEATURE_CLUSTER_MANAGER=true` is set.

I think this is the only thing we want to drive: Enable or disable the commands in the CLI. So we don't need the snap for that. We can keep it simple and rely on an environment variable for that. I added a commit for just that above.

roosterfish:
> I think this is the only thing we want to drive: Enable or disable the commands in the CLI. So we don't need the snap for that.

As discussed in chat let's also have the feature flag in the snap so we can expose the env var in the respective CLI https://github.com/canonical/microcloud-pkg-snap/blob/latest-edge/snapcraft/commands/microcloud.

api/cluster_manager.go
```diff
@@ -0,0 +1,677 @@
+package api
```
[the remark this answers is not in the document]
roosterfish: I see the commits show the iterative process after all the reviews of the PR/spec. But can you please use the same package based commit messages and introduce the changes together commit by commit? This way we don't add stuff and delete it later in another commit within the same PR.

edlerd:
> As discussed in chat let's also have the feature flag in the snap so we can expose the env var in the respective CLI https://github.com/canonical/microcloud-pkg-snap/blob/latest-edge/snapcraft/commands/microcloud.

Yes, agreed, added the feature flag to the snap configuration in https://github.com/canonical/microcloud-pkg-snap/pull/65

api/cluster_manager.go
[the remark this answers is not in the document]
edlerd: I think this change only makes sense with all features, so would squash all commits into a single one. There seems to be no advantage in splitting this into multiple commits for merging.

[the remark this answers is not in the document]
edlerd: Split the commits accordingly

api/cluster_manager.go
```diff
@@ -0,0 +1,263 @@
...
+
+// clusterManagerPut updates the cluster manager configuration.
+func clusterManagerPut(state state.State, r *http.Request) response.Response {
+	args := types.ClusterManagerPut{}
+	err := json.NewDecoder(r.Body).Decode(&args)
+	if err != nil {
+		return response.BadRequest(err)
+	}
+
+	clusterManager, updateIntervalConfig, err := database.LoadClusterManagerConfig(state, r.Context())
```
roosterfish, line 168: Same here about the variable name `updateIntervalConfig`.

api/cluster_manager.go
```diff
@@ -0,0 +1,263 @@
...
+			if err != nil {
+				return err
+			}
+		}
+
+		if args.UpdateInterval == nil {
+			return nil
+		}
+
+		if *args.UpdateInterval == "" && len(updateIntervalConfig) > 0 {
```
roosterfish, line 195: I wasn't commenting on this if/else in the first review as I hoped the complexity would reduce in this section.

Can't this logic be inside `UpdateClusterManagerConfig` and be generic for all kinds of future config keys?
Even though the struct is named `ClusterManagerConfig` right now its purpose is only about tracking the update interval.
We should make this as generic as possible right from the beginning to be able to easily add new config keys in the future.

I would propose a `UpdateClusterManagerConfig(<name>, map[string]string)` which works for both the interval and future config keys.

api/cluster_manager.go
```diff
@@ -0,0 +1,263 @@
...
+// clusterManagerDelete clears the cluster manager configuration.
+func clusterManagerDelete(sh *service.Handler) func(state state.State, r *http.Request) response.Response {
+	return func(state state.State, r *http.Request) response.Response {
+		clusterManager, _, err := database.LoadClusterManagerConfig(state, r.Context())
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		err = state.Database().Transaction(context.Background(), func(ctx context.Context, tx *sql.Tx) error {
+			err = database.DeleteClusterManager(ctx, tx, clusterManager.ID)
```
roosterfish, line 238: Let's also use the name of the cluster manager here instead of the ID.

api/services_proxy.go
roosterfish: Good catch. But can you please move this out of the `if r.URL.Path == "/1.0/metrics" {` as it should also defer if the path is different.

api/types/cluster_manager.go
```diff
@@ -0,0 +1,82 @@
...
+	CPULoad5          string               `json:"cpu_load_5"`
+	CPULoad15         string               `json:"cpu_load_15"`
+	MemoryTotalAmount int64                `json:"memory_total_amount"`
+	MemoryUsage       int64                `json:"memory_usage"`
+	DiskTotalSize     int64                `json:"disk_total_size"`
+	DiskUsage         int64                `json:"disk_usage"`
+	MemberStatuses    []StatusDistribution `json:"member_statuses"`
+	InstanceStatuses  []StatusDistribution `json:"instance_statuses"`
+	Metrics           string               `json:"metrics"`
+	UiUrl             string               `json:"ui_url"`
```
roosterfish, line 74: `UIURL`?

client/cluster_manager_client.go
```diff
@@ -0,0 +1,186 @@
...
+		ClusterCertificate: string(clusterCert.PublicKey()),
+		Token:              encodedToken,
+	}
+
+	reqBody, err := json.Marshal(payload)
+	if err != nil {
+		return err
+	}
+
+	req, err := c.craftRequest("POST", "/1.0/remote-cluster", bytes.NewReader(reqBody))
```
roosterfish, line 46: Worth having a constant for `/1.0/remote-cluster`?

client/cluster_manager_client.go
```diff
@@ -0,0 +1,186 @@
...
+func (c *ClusterManagerClient) getHTTPClient(clusterCert *shared.CertInfo) (*http.Client, string, error) {
+	client := &http.Client{}
+
+	var address string
+	var remoteCert *x509.Certificate
+	var err error
+
+	addresses := strings.Split(c.config.Addresses, ",")
+	if len(addresses) == 0 {
+		return nil, "", errors.New("No cluster manager addresses.")
```
roosterfish, line 139: Please remove `.` at the end

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,350 @@
...
+
+type cmdClusterManager struct {
+	common *CmdControl
+}
+
+// Command returns the subcommand to manage cluster manager configuration.
+func (c *cmdClusterManager) Command() *cobra.Command {
+	cmd := &cobra.Command{}
+	cmd.Use = "cluster-manager"
+	cmd.Short = "Manage cluster-manager connection"
```
roosterfish, line 26: The other short desc use "cluster manager" without dash.

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,350 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = apiClient.QueryStruct(context.Background(), "POST", "1.0", api.NewURL().Path("cluster-managers"), payload, nil)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Successfully joined cluster manager.")
```
roosterfish, line 101: Please omit `.`

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,350 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = apiClient.QueryStruct(context.Background(), "DELETE", "1.0", api.NewURL().Path("cluster-managers", "default"), nil, nil)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Cluster manager configuration cleared.")
```
roosterfish, line 173: I think we don't have to print a string here. Its fine if the command simply returns with exit 0.

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,350 @@
...
+microcloud cluster-manager get status-last-error-response`)
+
+	cmd.RunE = c.run
+
+	return cmd
+}
+
+func (c *cmdClusterManagerGet) run(_ *cobra.Command, args []string) error {
+	if len(args) != 1 {
+		return fmt.Errorf("Expected exactly one argument with the key.")
```
roosterfish, line 202: Please omit `.`

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	var clusterManager *types.ClusterManagerGet
+	err = apiClient.QueryStruct(context.Background(), "GET", "1.0", api.NewURL().Path("cluster-managers", "default"), nil, &clusterManager)
+	if err != nil {
+		return err
+	}
+
+	switch key {
+	case "addresses":
```
roosterfish, line 219: Lets use a `?format=(json|yaml)` query param so you can print the response from the API directly.
When adding new fields/config keys this will work out of the box with this command without requiring any modification.

cmd/microcloud/cluster_manager.go
roosterfish: As discussed we don't required this right from the beginning and can always add it later.

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,350 @@
...
+microcloud cluster-manager set update-interval-seconds 50`)
+
+	cmd.RunE = c.run
+
+	return cmd
+}
+
+func (c *cmdClusterManagerSet) run(_ *cobra.Command, args []string) error {
+	if len(args) != 2 {
+		return fmt.Errorf("Expected exactly two arguments: key and value.")
```
roosterfish, line 259: Please omit the `.`

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	payload := types.ClusterManagerPut{}
+
+	switch key {
+	case "update-interval-seconds":
+		emptyString := ""
+		payload.UpdateInterval = &emptyString
+	default:
+		return fmt.Errorf("Invalid key.")
```
roosterfish, line 327: `.`

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/canonical/microcloud/microcloud/api/types"
+	"github.com/canonical/microcloud/microcloud/client"
+	"github.com/canonical/microcloud/microcloud/database"
+	"github.com/canonical/microcloud/microcloud/service"
+)
+
+// SendClusterManagerStatusMessageTask starts a go routine, that sends periodic status messages to cluster manager.
+func SendClusterManagerStatusMessageTask(ctx context.Context, sh *service.Handler, s state.State) {
+	go func(ctx context.Context, sh *service.Handler, s state.State) {
+		ticker := time.NewTicker(60 * time.Second)
```
roosterfish, line 22: The same is already defined as a constant in the API handler. Let's reuse it here too and maybe move it to another place so its exported.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -0,0 +1,231 @@
...
+	err = clusterManagerClient.PostStatus(clusterCert, payload)
+	if err != nil {
+		logger.Error("Failed to send status message to cluster manager", logger.Ctx{"err": err})
+		_ = database.SetClusterManagerStatusLastError(s, ctx, time.Now(), err.Error())
+		return nextUpdate
+	}
+
+	_ = database.SetClusterManagerStatusLastSuccess(s, ctx, time.Now())
+
+	logger.Debug("Finished sendClusterManagerStatusMessage.")
```
roosterfish, line 143: `.`

database/cluster_manager_config.go
```diff
@@ -0,0 +1,117 @@
...
+	"context"
+	"database/sql"
+	"errors"
+	"time"
+
+	"github.com/canonical/microcluster/v2/state"
+)
+
+// ClusterManagerDefaultName is the default name for the cluster manager.
+const ClusterManagerDefaultName = "default"
```
roosterfish, line 13: Ah I see, you are falling back to this name by default now in `ClusterManagerIsConfigured`. Think that is fine for the beginning but please lets use the same interface and always use the name and not have some funcs that require the ID of the cluster manager in the DB.

database/cluster_manager_config.go
```diff
@@ -0,0 +1,117 @@
...
+			Name: &name,
+		}
+
+		clusterManagers, err := GetClusterManagers(ctx, tx, filter)
+		if err != nil {
+			return err
+		}
+
+		if len(clusterManagers) < 1 {
+			return errors.New("Cluster manager not found")
```
roosterfish, line 105: This should return a `api.StatusErrorf(http.StatusNotFound, "Cluster manager %q not found", "default")` because when calling `loadDefaultClusterManager` you can then check for the 404 or simply return the HTTP error to the actual caller if its from a HTTP request.

database/cluster_manager_config.go
```diff
@@ -0,0 +1,117 @@
...
+const ClusterManagerDefaultName = "default"
+
+// UpdateIntervalKey is the key for the update interval configuration.
+const UpdateIntervalKey = "UpdateInterval"
+
+// ClusterManagerIsConfigured finds if cluster manager is setup in the database.
+func ClusterManagerIsConfigured(state state.State, ctx context.Context) (bool, error) {
+	_, err := loadDefaultClusterManager(ctx, state)
+	if err != nil {
+		if err.Error() == "Cluster manager not found" {
```
roosterfish, line 22: See the other comment about returning a 404 from `loadDefaultClusterManager`. You can then use `api.StatusErrorMatch()` to check if it's a 404.

database/cluster_manager_schema.go
```diff
@@ -0,0 +1,50 @@
...
+	stmt := `
+CREATE TABLE cluster_manager (
+    id                         INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+    addresses                  TEXT NOT NULL,
+    certificate_fingerprint    TEXT NOT NULL,
+    name                       TEXT NOT NULL,
+    status_last_success_time   DATETIME,
+    status_last_error_time     DATETIME,
+    status_last_error_response TEXT,
+    UNIQUE (name)
```
roosterfish, line 27: Good! As the name is marked as UNIQUE we can use it in all the DB funcs when loading the cluster manager (instead of using the ID which is usually unknown to the caller).

roosterfish (changes requested):
Thanks for moving the API bits into their own funcs. This looks good now.

My main concerns are around the implicit handling of the `default` cluster manager. I think its fine for now but the DB funcs to get/set/update the cluster manager should either always implicitly use `default` or the ID but not both to stay consistent.

api/types/cluster_manager.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+// ClusterManagersPost represents the cluster manager configuration when receiving a POST request in MicroCloud.
+type ClusterManagersPost struct {
+	Name  string `json:"name" yaml:"name"`
+	Token string `json:"token" yaml:"token"`
+}
+
+// ClusterManagerGet represents the configuration in MicroCloud of cluster manager.
+type ClusterManagerGet struct {
```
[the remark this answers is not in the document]
edlerd: The name would clash with the database type. I think it is better to distinguish between them.

api/cluster_manager.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// clusterManagerGet returns the cluster manager configuration.
+func clusterManagerGet(state state.State, r *http.Request) response.Response {
+	clusterManager, updateIntervalConfig, err := database.LoadClusterManagerConfig(state, r.Context())
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if clusterManager.Addresses == "" {
```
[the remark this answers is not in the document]
edlerd: yes, all good ideas :)

api/types/cluster_manager.go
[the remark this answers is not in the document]
roosterfish: Should be fine as it is in another package.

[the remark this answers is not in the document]
tomponline: > Should be fine as it is in another package.

Agree lets stick with the conventions.

cmd/microcloud/cluster_manager.go
edlerd: Can you clarify what you meant?

api/types/cluster_manager.go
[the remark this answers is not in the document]
edlerd: ok, will adjust it.

cmd/microcloud/cluster_manager.go
roosterfish: The API endpoint to view the cluster manager (including config) can grow a `format` query param so that instead of JSON it returns a yaml payload which you can print right away so you don't have to perform this mapping.

roosterfish: Ah my bad, I thought it queries the entirety of the cluster manager.

This is now the downside of merging both config table and the cluster manager table together. What if the API returns the config keys as `map[string]string` but only in the frontend we merge everything together? This way you could easily filter in this func for the requested config key.

edlerd: That wouldn't help either, we are requesting single config fields here, also splitting each column in the table.

edlerd: like this is for `microcloud cluster-manager get fingerprint` or `microcloud cluster-manager get addresses` any api response that has more than one field would have to be filtered in some way to pick single fields out of the repsonse struct.

I think the current approach to mangle the response into a struct and then printing the matching field is superior.

roosterfish: > That wouldn't help either, we are requesting single config fields here

Exactly. If the API returns a `map[string]string` you can just filter for the key and don't require another `case` in the `switch` for every new config key.

edlerd: Ah now I understand what you mean.

edlerd: Thanks for your patience :sweat_smile:

edlerd:
Some [test is failing](https://github.com/canonical/microcloud/actions/runs/14338918690/job/40193220902?pr=630), but unrelated to the changes:
```
+ sudo snap install lxd --channel 5.21/edge
error: cannot perform the following tasks:
- Download snap "core22" (1802) from channel "stable" (unexpected EOF)
Error: Process completed with exit code 1.
```

api/services_proxy.go
edlerd: Moving this up leads to blocking of the service proxy. Because if the body is not read, then the defer blocks any communication with other services.

cmd/microcloud/cluster_manager.go
edlerd: Adjusted the api layer. I think the case statement here should still be there as a guard and translator between the keys.

api/cluster_manager.go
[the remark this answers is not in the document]
roosterfish: Can you please also use the package name as prefix of the commit message? See the other commits as reference.

roosterfish:
I was able to reproduce the error we talked about. When MicroCloud tries to reach the cluster manager, it gets this response:

```
status_last_error_response: |+
  Failed to send request to cluster manager: 500 Internal Server Error, body: {"type":"error","status":"","status_code":0,"operation":"","error_code":500,"error":"exec tran: failed to forward metrics to Prometheus: failed to send metrics to Prometheus: Post \"http://192.168.1.133/cos-prometheus-0/api/v1/write\": dial tcp 192.168.1.133:80: connect: network is unreachable","metadata":null}
```

Is 192.168.1.133 hardcoded somewhere? It's not any of my local IPs from the test setup.

roosterfish (changes requested):
When I try to delete a cluster manager which is already doesn't exist anymore, I am getting this error. Why does MicroCloud need to reach out to the cluster manager when getting rid of it? A retry fixes it but still I think we shouldn't try to contact cluster manager:

```bash
root@micro01:~# microcloud cluster-manager delete
Error: Get "https://cc.lxd-cm.local:30000": Unable to connect to: cc.lxd-cm.local:30000 ([dial tcp 10.87.144.34:30000: connect: connection refused])
root@micro01:~# microcloud cluster-manager delete
Error: Cluster manager not found

```

edlerd:
> > When I try to delete a cluster manager which is already doesn't exist anymore, I am getting this error. Why does MicroCloud need to reach out to the cluster manager when getting rid of it? A retry fixes it but still I think we shouldn't try to contact cluster manager:
> 
> This is for microcloud to remove itself from the cluster manager registry. To, not leave a dead cluster entry behind. We can make this request optional and not show any errors. Microcloud still unconfigures the cluster manager locally as [detailed in the spec](https://docs.google.com/document/d/16wdV9l-iHz68qhn4M9eWaYOaBOPnO3Y8hoJDbfTB9wk/edit?tab=t.0#heading=h.6lyix28qmcvg).

@roosterfish Shall 1. we keep this as is 2. hide the error or 3. change the spec, remove the delete request and leave behind a dead cluster entry in cluster manager?

edlerd:
> When I try to delete a cluster manager which is already doesn't exist anymore, I am getting this error. Why does MicroCloud need to reach out to the cluster manager when getting rid of it? A retry fixes it but still I think we shouldn't try to contact cluster manager:

I created an [open issue in cluster manager](https://github.com/canonical/microcloud-cluster-manager/issues/183) for this. It shouldn't be a concern to this PR.

edlerd:
> Can we leave it up to the user to decide using a `--force` flag? Meaning if the connection doesn't work, it errors out but the cluster manager won't be removed. Only if you run with --force we don't reach out to the cluster manager but remove it from MicroCloud which will then result in an entry on cluster manager.

Ah right, doing the request first and offering a force flag sounds like the best approach here. Will add this to the PR shortly :-1:

api/cluster_manager.go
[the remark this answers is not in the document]
edlerd: Adjusted the commit messages.

edlerd:
@roosterfish please have another look at this one.

As discussed, I revised deletion to fail if the unregister request fails. And added a force flag to avoid the request on deletion. Also adjusted on the commit messages. The test failure seems to be unrelated to the changes.

roosterfish:
> The test failure seems to be unrelated to the changes.

Thanks. Speaking about tests, can we have some basic tests to not cause too much degradation in coverage?

edlerd:
> Thanks. Speaking about tests, can we have some basic tests to not cause too much degradation in coverage?

Thanks for you help on addidng the cluster manager test suite. Now ready for review :)

test/suites/cluster-manager.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+test_cluster_manager() {
+  # Reset the systems to a single node
+  reset_systems 1 0 1
```
roosterfish, line 5: Should work with `1 0 0`. I think there was another hiccup earlier on when we were testing this.

roosterfish (changes requested):
Thanks for adding the tests, think this is great so we don't reduce the coverage percentage too much. Just added a few smaller comments.

test/suites/cluster-manager.sh
```diff
@@ -0,0 +1,54 @@
...
+  export SKIP_LOOKUP=1
+  export SETUP_CEPH="no"
+  export SETUP_OVN="no"
+  export SETUP_ZFS="no"
+  join_session init micro01
+
+  lxc exec micro01 -- tail -1 out | grep "MicroCloud is ready" -q
+
+  echo "==> Test cluster manager without previous setup"
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager show | grep "Error: Cluster manager not found" -q || true
```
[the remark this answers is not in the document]
edlerd: This is necessary or the test fails. I suspect this is because the return code is still non zero:

```
edlerd@EDLERD-T14:~$ lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager show | grep "Error: Cluster manager not found" -q
Error: Failed to fetch instance "micro01" in project "default": Instance not found
edlerd@EDLERD-T14:~$ echo $?
1
```

test/suites/cluster-manager.sh
```diff
@@ -0,0 +1,54 @@
...
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager delete | grep "Error: Cluster manager not found" -q || true
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager get update-interval-seconds | grep "Error: Cluster manager not found" -q || true
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager set update-interval-seconds 10 | grep "Error: Cluster manager not found" -q || true
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager unset update-interval-seconds | grep "Error: Cluster manager not found" -q || true
+
+  echo "==> Create a cert for dummy cluster manager"
+  lxc exec micro01 -- openssl req -x509 -newkey rsa:2048 -nodes -keyout key.pem -out cert.pem -days 1 -subj "/CN=localhost" -addext "subjectAltName = DNS:localhost"
+  echo "==> start dummy cluster manager server"
```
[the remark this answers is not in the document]
edlerd: Ah yes, sure.

test/suites/cluster-manager.sh
```diff
@@ -0,0 +1,54 @@
...
+  echo "==> run cluster manager commands"
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager get update-interval-seconds | grep "60" -q
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager set update-interval-seconds 15
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager get update-interval-seconds | grep "15" -q
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager unset update-interval-seconds
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager set update-interval-seconds 60
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager show | grep "certificate_fingerprint:" -q
+
+  echo "==> delete cluster manager"
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager delete
```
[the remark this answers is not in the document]
edlerd: Yes, good idea. Will add this shortly.

test/suites/cluster-manager.sh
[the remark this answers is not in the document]
edlerd: Ah but we can redirect stderr to stdout and then grep should catch it and we can drop the true.

test/suites/cluster-manager.sh
edlerd: The `1 0 0` still fails locally (see below). I'd keep it as `1 0 1`.

```
edlerd@EDLERD-T14:~$ lxc shell micro01 --project=microcloud-test
root@micro01:~# cat out

  Detected 'TEST_CONSOLE=1', MicroCloud CLI is in testing mode. Terminal interactivity is disabled.

  Interactive microcloud commands will read text instructions by line:

cat << EOF | microcloud init
table:select                # selects an element in the table
table:select-all            # selects all elements in the table
table:select-none           # de-selects all elements in the table
table:up                    # move up in the table
table:down                  # move down in the table
table:wait <time.Duration>  # waits before the next instruction
table:expect <count>        # waits until exactly <count> peers are available, and errors out if more are found
table:filter <text>         # applies filtering text to the table output
table:done                  # confirms the table selection and exits the table

# anything not prefixed with 'table:' will be treated as a raw string. This is used for text entry for individual questions
EOF

Do you want to set up more than one cluster member? (yes/no) [default=yes]: 
 Using address 10.195.89.4 for MicroCloud

Gathering system information ...
Warning: No disks available for distributed storage. Skipping configuration
! Some systems are ineligible for distributed networking. At least one interface in state UP with no IPs assigned or a bridge is required
Continue anyway? (yes/no) [default=yes]: Error: User aborted
```

edlerd:
New [cluster manager](https://github.com/canonical/microcloud/actions/runs/14709390515/job/41277838254?pr=630) test suite is passing. The failing test below seems unrelated.

I addressed the issues flagged above. Please have another review @roosterfish

api/cluster_manager.go
```diff
@@ -0,0 +1,260 @@
...
+		if err != nil {
+			return response.BadRequest(err)
+		}
+
+		if args.Token == "" {
+			return response.BadRequest(errors.New("No token provided"))
+		}
+
+		if args.Name != database.ClusterManagerDefaultName {
+			return response.BadRequest(errors.New("Invalid cluster manager name, only " + database.ClusterManagerDefaultName + " is allowed"))
```
roosterfish, line 90: We usually do `fmt.Sprintf("... %q ...")` across the code base.

api/cluster_manager.go
```diff
@@ -0,0 +1,260 @@
...
+
+		joinToken, err := shared.JoinTokenDecode(args.Token)
+		if err != nil {
+			return response.BadRequest(err)
+		}
+
+		// ensure cluster manager is not already configured
+		existingClusterManager, _, err := database.LoadClusterManager(state, r.Context(), args.Name)
+		if err != nil {
+			if err.Error() == "Cluster manager not found" {
```
roosterfish, line 101: `LoadClusterManager` already returns a 404 status error so you can use the `api.StatusErrorCheck(err, http.StatusNotFound)` func instead which doesn't require comparing the error string.

api/cluster_manager.go
```diff
@@ -0,0 +1,260 @@
...
+				return response.SmartError(err)
+			}
+		}
+
+		err = database.RemoveClusterManager(state, r.Context(), *clusterManager)
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		if err != nil {
```
roosterfish, line 241: Redundant `if`.

client/cluster_manager_client.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		return err
+	}
+
+	req, err := c.craftRequest("POST", RemoteClusterPath, bytes.NewReader(reqBody))
+	if err != nil {
+		return err
+	}
+
+	err = c.sendRequest(clusterCert, req)
```
roosterfish, line 54: Not a blocker but you can do `return c.sendRequest(clusterCert, req)`

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -0,0 +1,228 @@
...
+	if err != nil {
+		logger.Error("Failed to get cluster certificate", logger.Ctx{"err": err})
+		return nextUpdate
+	}
+
+	clusterManagerClient := client.NewClusterManagerClient(clusterManager)
+	err = clusterManagerClient.PostStatus(clusterCert, payload)
+	if err != nil {
+		logger.Error("Failed to send status message to cluster manager", logger.Ctx{"err": err})
+		_ = database.SetClusterManagerStatusLastError(s, ctx, database.ClusterManagerDefaultName, time.Now(), err.Error())
```
roosterfish, line 134: Let's log those errors too, also the line below. 
Or is there a reason for skipping them?

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	token := args[0]
+
+	if token == "" {
+		return fmt.Errorf("Token cannot be empty")
+	}
+
+	payload := types.ClusterManagersPost{
+		Name:  "default",
```
roosterfish, line 88: You can use the `ClusterManagerDefaultName` constant.

cmd/microcloud/cluster_manager.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func (c *cmdClusterManagerShow) run(_ *cobra.Command, _ []string) error {
+	apiClient, err := getApiClient(c.common)
+	if err != nil {
+		return err
+	}
+
+	var clusterManager *types.ClusterManager
+	err = apiClient.QueryStruct(context.Background(), "GET", "1.0", api.NewURL().Path("cluster-managers", "default"), nil, &clusterManager)
```
roosterfish, line 131: You can use the `ClusterManagerDefaultName` const.

roosterfish: There are a few more of those.

roosterfish (changes requested):

roosterfish:
@edlerd can you please rebase `main`?

edlerd:
Thanks for the review @roosterfish 

I rebased on main and addressed all the recent comments.

roosterfish:
> I rebased on main and addressed all the recent comments.

Thanks, a few more static analysis errors now, probably from the rebase https://github.com/canonical/microcloud/actions/runs/15161316988/job/42627897986?pr=630.
We recently landed https://github.com/canonical/microcloud/pull/777.

test/suites/cluster-manager.sh
roosterfish: Checked again, I guess you have to set `PROCEED_WITH_NO_OVERLAY_NETWORKING=no` together with `SETUP_OVN="no"` as this will answer the failing question. Then you should be able to use `reset_systems 1 0 0`.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -0,0 +1,235 @@
...
+	leaderInfo, err := leaderClient.Leader(ctx)
+	if err != nil {
+		logger.Error("Failed to get database leader info", logger.Ctx{"err": err})
+		return nextUpdate
+	}
+
+	if leaderInfo.Address != s.Address().URL.Host {
+		logger.Debug("Not the leader, skipping status message")
+		return nextUpdate
+	}
```
roosterfish, line 71: I was just wondering if this should come after fetching the config? Because otherwise `sendClusterManagerStatusMessage` always returns `0` on non leaders which causes only the loop on the leader to pick up the latest interval but the followers will always use the old interval.
Which means as soon as the leader changes, they are pushing info using the old interval until the next iteration.

roosterfish (changes requested):

cmd/microcloudd/cluster_manager_task.go
edlerd: I think this is better as is. The default time is applied on a non-leader when they get elected, and the default should be reasonable as a start on leader change.

The downside of moving this after the config fetch is that every cluster member will do a db read to fetch the correct interval. It seems cheaper to keep it only on the leader and the other members will not stress the db.

roosterfish: I think the "number of cluster members" DB reads per minute (default) won't put too much load on the DB :) But maybe if the leader changes and you have an interval of let's say 10m it might be desired to send a new ping rather sooner than later to surface the latest details also in cluster manager. Not a blocker from my side. Happy to leave it as is for now.

edlerd: If reads are cheap, happy to adjust it. Switched the order to always read the config first and then do the leader check to have an updated interval on all members.

roosterfish (changes requested):
Gave it another go, two more small things and then I have the feeling the PR can be merged.

(1) The `status_last_error_response` field when running `microcloud cluster-manager show` doesn't get cleaned if the most recent ping was successful. Whilst I understand it might be interesting to know that there was once an error, it gives you the feeling that something isn't right even though the most recent ping was successful. 
Can we please cleanup this field if there was no error on the most recent ping?

(2) I thought we wanted to merge both the contents from the cluster manager table and config table when running `microcloud cluster-manager show` to give the user a seamless experience when running `get/set` commands. Because right now the same command updates the `address` at the root but also the `extra_configuration.*` keys. 
Let's get rid of the `extra_configuration:` and merge them all on the root level. This way we still have to check if its a setting from the core or config table but the management of the settings follows the same approach from user perspective.

The rest looks great!

api/types/cluster_manager.go
```diff
@@ -0,0 +1,87 @@
...
+	// Example: 2025-03-26T11:37:17.83536772Z
+	StatusLastErrorTime time.Time `json:"status_last_error_time" yaml:"status_last_error_time"`
+
+	// The last error response received from the cluster manager
+	// Example: Failed to send request to cluster manager: 403 Forbidden, body: {"type":"error","status":"","status_code":0,"operation":"","error_code":403,"error":"not authorized","metadata":null}
+	StatusLastErrorResponse string `json:"status_last_error_response" yaml:"status_last_error_response"`
+
+	// Additional configuration for the cluster manager
+	// Example: update_interval: 60
+	ExtraConfiguration map[string]string `json:"extra_configuration" yaml:"extra_configuration"`
```
roosterfish, line 42: Let's do the same as in LXD and just name this `config`.

edlerd:
> (1) The `status_last_error_response` field when running `microcloud cluster-manager show` doesn't get cleaned if the most recent ping was successful. Whilst I understand it might be interesting to know that there was once an error, it gives you the feeling that something isn't right even though the most recent ping was successful. Can we please cleanup this field if there was no error on the most recent ping?
> 
> (2) I thought we wanted to merge both the contents from the cluster manager table and config table when running `microcloud cluster-manager show` to give the user a seamless experience when running `get/set` commands. Because right now the same command updates the `address` at the root but also the `extra_configuration.*` keys. Let's get rid of the `extra_configuration:` and merge them all on the root level. This way we still have to check if its a setting from the core or config table but the management of the settings follows the same approach from user perspective.

Good ideas, I addressed both (1), (2) and the comment above.

roosterfish (approved):
LGTM!

PR #633, opened 2025-02-13
doc: improve explanation of networking approach

minaelee:
This PR adds an explanation/networking page to explain MicroCloud's networking approach, including its use of OVN and a comparison to VPCs (Virtual Private Clouds).

This reduces the need for a Networking section in the About MicroCloud page. A separate commit removes most of that section, and also re-allocates most of the remaining information to a more suitable location: the Reference page on networking requirements. It also updates relevant links.


doc/explanation/networking.md
```diff
@@ -0,0 +1,105 @@
...
+Along with enabling communication between MicroCloud cluster members, OVN also manages key network components such as overlays, virtual routers and switches, NAT, and more. 
+
+(exp-networking-ovn-features)=
+### Key features and benefits
+
+- **Logical networks**: OVN networks are logical networks implemented in software, which means they operate independently of physical network infrastructure and allow for segmentation within the same cluster. For example, you can use separate networks for frontend and backend services.
+
+- **Distributed virtual routers and logical switches**: Data plane operations are distributed across cluster members rather than run by a centralised controller, allowing network traffic to be processed at the edge, closer to where workloads run.
+
+- **Consistent networking across locations**: Since OVN logical networks are software-defined, they can span multiple physical locations or clusters while maintaining consistent configurations.
```
tomponline, line 26: @minaelee this suggests that microcloud supports this, which it does not at this time (no cross-cluster interconnect at ovn level). Or does this line refer to something else?

doc/explanation/networking.md
```diff
@@ -0,0 +1,105 @@
+---
+relatedlinks: https://docs.ovn.org,https://ubuntu.com/blog/data-centre-networking-what-is-ovn,https://en.wikipedia.org/wiki/Virtual_private_cloud
+---
+
+(exp-networking)=
+# MicroCloud's networking approach
+
+This page describes a high-level overview of MicroCloud's networking approach, particularly its use of [Open Virtual Network (OVN)](https://www.ovn.org/). It also compares and contrasts MicroCloud networking to [Virtual Private Clouds (VPCs)](https://en.wikipedia.org/wiki/Virtual_private_cloud).
```
tomponline, line 8: @escabo @mionaalex  is it ok to use the VPC reference here?

doc/explanation/networking.md
```diff
@@ -0,0 +1,105 @@
...
+
+The following figure shows the OVN network traffic flow in a MicroCloud/LXD cluster:
+
+```{figure} /images/ovn_networking_1.svg
+:width: 100%
+
+OVN networking (single network)
+```
+
+In MicroCloud, the OVN network connects all cluster members. Intra-cluster traffic (communication between members of the same cluster, also known as OVN east/west traffic) travels through an OVN tunnel over a designated {abbr}`NIC (Network Interface Card)` (`eth1` in the figure above).
```
tomponline, line 47: "the OVN network" sounds a bit misleading, true, microcloud setups an OVN network by default, but its also possible and useful for the user of microcloud to setup multiple OVN networks, as you mentioned earlier:

> For example, you can use separate networks for frontend and backend services.

Perhaps say "In MicroCloud, each OVN network spans across all cluster members" (chose "span" here on purpose).

> Intra-cluster traffic (communication between members of the same cluster, also known as OVN east/west traffic) travels through an OVN tunnel ...

This line didn't quite read right to me technically, the context is about OVN east/west traffic, yet it means "communication between members of the same cluster", which to me (and I am biased ofc) suggests communication between the `micro*` or `lxd` services on those members, which OVN traffic is not part of that.

How about this instead to avoid the ambiguity?

> Intra-cluster traffic (communication between instances on members within the same cluster, also known as OVN east/west traffic) travels through an OVN tunnel

doc/explanation/networking.md
```diff
@@ -0,0 +1,105 @@
...
+
+For external connectivity, the OVN network connects to an uplink network via a virtual router on a designated NIC (`eth0` in the figure above). The virtual router is active on only one cluster member at a time but can migrate between cluster members as needed, ensuring all members maintain uplink access.
```
tomponline, line 49: > For external connectivity

Suggest

> For external connectivity (OVN north/south traffic)

tomponline, line 49: > ensuring all members maintain uplink access

suggest

> ensuring uplink access remains operational should the active member because unavailable

doc/explanation/networking.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+Both OVN networks depicted are completely independent. Both networks are available on all cluster members (with each virtual router active on one random cluster member). Any instance can use either of the networks, and the traffic on one network is completely isolated from the other network.
+
+See the {ref}`LXD documentation on OVN networks <lxd:network-ovn>` for more information.
+
+(exp-networking-ovn-underlay)=
+### Dedicated underlay network
+
+During {ref}`MicroCloud initialisation <howto-initialise>`, you can choose to {ref}`configure a dedicated underlay network for OVN traffic <howto-ovn-underlay>`. This requires an additional network interface on each cluster member.
```
tomponline, line 68: Optional, but somewhere it'd be nice to indicate that for all of the optional dedicated interfaces (ceph, ovn) and the microcloud cluster interface itself, that we dont literally mean a physical NIC, but it can be a VLAN interface ontop of a single physical or multiple physical bonded NICs. cc @MggMuggins

doc/explanation/networking.md
```diff
@@ -0,0 +1,105 @@
...
+
+A dedicated underlay network serves as the physical infrastructure over which the virtual (overlay) network is constructed. While optional, it offers several benefits:
```
tomponline, line 70: Given my comment here https://github.com/canonical/microcloud/pull/633/commits/ecf96744016ca75a15c98162ef41c43a3a267fc3#r1954242012 suggest we add the word "potential":

> it offers several potential benefits:

As "Reduced congestion" and "Optimised performance" wont necessarily be true if using a shared physical link with VLANs on for "Traffic isolation".

tomponline (changes requested):
Overall great thanks, ive left a few suggestions though.

doc/explanation/networking.md
mionaalex: @tomponline Yes, I think this is fine. It is a common concept, and I think it will help people understand it better.

doc/explanation/networking.md
minaelee: It is already mentioned for the intra-cluster interface (it says: "This network interface can be, for example, a dedicated physical network interface, a VLAN, or a virtual function on an {abbr}`SR-IOV (Single root I/O virtualisation)`-capable network interface.") and I added a commit that mentions that a VLAN interface or physical NIC can be used in the reference/index.md section on optional interfaces.

doc/explanation/networking.md
minaelee: Fixed to: 
- **Consistent networking across locations**: Since OVN logical networks are software-defined, the same configurations can be consistently deployed across multiple clusters.

doc/explanation/networking.md
minaelee: Updated per your suggestions

tomponline (dismissed):

mionaalex (approved):
Only now got a chance to thoroughly review. Approved from my side when it comes to structure, readability and context. I think it provides a good overview and clarity.  For technical correctness I'll rely on the engineers to double-check.

doc/how-to/ovn_underlay.md
```diff
@@ -1,14 +1,15 @@
 (howto-ovn-underlay)=
 # How to configure an OVN underlay network
 
-When running {command}`microcloud init`, if you decided to setup distributed networking and if you have enough at least one unused network interface per cluster member with an IP address, you are asked if you want to configure an underlay network for OVN.
-Here is the question you will be asked:
+For an explanation of the benefits of using an OVN underlay network, see {ref}`exp-networking-ovn-underlay`. 
+
+When running {command}`microcloud init`, if you chose to set up distributed networking and if you have enough at least one unused network interface per cluster member with an IP address, MicroCloud asks if you want to configure an underlay network for OVN:
```
roosterfish, line 6: I think the "unused network" in the original version isn't correct either.
Any interface with an address can be configured regardless whether or not the interface got already configured for any of the other networks.

doc/reference/index.md
```diff
@@ -54,27 +52,57 @@ For production environments, we recommend at least 3 NVMe disks per cluster memb
...
 In production environments, we recommend dual-port network cards with a minimum 10 GiB capacity, or higher if low latency is essential.
 
-If you want to partially or fully disaggregate the Ceph networks and the OVN underlay network, you need more dedicated interfaces. For details, see: {ref}`howto-ceph-networking`.
+(microcloud-networking-intracluster)=
+Network interface for intra-cluster traffic
+:  MicroCloud requires one network interface that is pre-configured with an IP address that is within the same subnet as the IPs of the other cluster members.
+   The network that it is connected to must support multicast.
+
+   This network interface can be, for example, a dedicated physical network interface, a VLAN, or a virtual function on an {abbr}`SR-IOV (Single root I/O virtualisation)`-capable network interface.
+   It serves as the dedicated network interface for MicroOVN and is used for multicast discovery (during setup) and all internal traffic between the MicroCloud, OVN, and Ceph members.
```
roosterfish, line 72:
```suggestion
   It serves as the dedicated network interface for MicroCloud and is used for multicast discovery (during setup) and all internal traffic between the MicroCloud, OVN, and Ceph members.
```

roosterfish (changes requested):
Looks good, just two small comments.

minaelee:
> Looks good, just two small comments.

Addressed, thanks!

roosterfish (approved):

PR #634, opened 2025-02-13
Network: Populate the `NetworkInfo` with the right subnet address

roosterfish:
Don't falsely populate the subnet by using the regular address. Use the actual network address.

Before this was setting `10.1.2.3/24` as the `Subnet` but it should actually set `10.1.2.0/24`.


gabrielmougard (approved):
LGTM

PR #651, opened 2025-02-20
Rename all occurances of service to component

edlerd:
# Done
* refactored all occurances of service to component
* to free up the service name


minaelee (changes requested):
Reviewed docs changes; see comments, thanks!

Issue #521, opened 2024-12-02
Clarify uplink interface error message


roosterfish:
In case eligible uplink interfaces aren't `UP`, MicroCloud prevents the setup of distributed networking but doesn't indicate that the cause is the interface being `DOWN`:

`Some systems are ineligible for distributed networking, which requires either an interface with no IPs assigned or a bridge. Continue anyway?`.

In the error message indicate that the interface has to be `UP`.

PR #666, opened 2025-02-25
Questionnaire: Clarify uplink interface error message

roosterfish:
Fixes https://github.com/canonical/microcloud/issues/521

Add a note that the uplink interface has to be in state UP.
Also use the same format for printing warnings as we do for printing errors.

### DIFF
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -1462,21 +1462,13 @@ func (c *initConfig) askClustered(s *service.Handler, expectedServices map[types
 
 			if info.ServiceClustered(serviceType) {
 				warning := fmt.Sprintf("%q is already part of a %s cluster", info.ClusterName, serviceType)
-				question := "Do you want to add this cluster to Microcloud? (add/skip) [default=add]"
-				validator := func(s string) error {
-					if !shared.ValueInSlice(s, []string{"add", "skip"}) {
-						return fmt.Errorf("Invalid input, expected one of (add,skip) but got %q", s)
-					}
-
-					return nil
-				}
-
-				addOrSkip, err := c.asker.AskStringWarn(warning, question, "add", validator)
+				question := "Do you want to add this cluster to MicroCloud?"
+				addOrSkip, err := c.asker.AskBoolWarn(warning, question, true)
 				if err != nil {
 					return err
 				}
 
-				if addOrSkip != "add" {
+				if !addOrSkip {
 					delete(s.Services, serviceType)
 				}
 
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -182,7 +182,9 @@ func (c *initConfig) askMissingServices(services []types.ServiceType, stateDirs 
 
 		// Ignore missing services in case of preseed.
 		if !c.autoSetup {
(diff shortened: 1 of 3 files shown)


gabrielmougard (approved):
LGTM

cmd/microcloud/ask.go
```diff
@@ -1453,7 +1454,8 @@ func (c *initConfig) askClustered(s *service.Handler, expectedServices map[types
 			}
 
 			if info.ServiceClustered(serviceType) {
-				question := fmt.Sprintf("%q is already part of a %s cluster. Do you want to add this cluster to Microcloud? (add/skip) [default=add]", info.ClusterName, serviceType)
+				question := "Do you want to add this cluster to Microcloud? (add/skip) [default=add]"
+				warning := fmt.Sprintf("%q is already part of a %s cluster", info.ClusterName, serviceType)
```
MggMuggins, line 1458: Just a nit but I think this reads easier if the question comes after the warning (as that is the order they will be displayed).

MggMuggins (commented):
Good stuff!

cmd/microcloud/ask.go
```diff
@@ -1453,7 +1454,8 @@ func (c *initConfig) askClustered(s *service.Handler, expectedServices map[types
 			}
 
 			if info.ServiceClustered(serviceType) {
-				question := fmt.Sprintf("%q is already part of a %s cluster. Do you want to add this cluster to Microcloud? (add/skip) [default=add]", info.ClusterName, serviceType)
+				question := "Do you want to add this cluster to Microcloud? (add/skip) [default=add]"
```
minaelee, line 1457:
```suggestion
				question := "Do you want to add this cluster to MicroCloud? (add/skip) [default=add]"
```

cmd/microcloud/ask.go
minaelee: I agree about the order. Also, why not y/n instead of add/skip?

minaelee (changes requested):

cmd/microcloud/ask.go
roosterfish: Good points, fixed and moved to `yes/no` to be consistent with the other questions.

cmd/microcloud/ask.go

MggMuggins (dismissed):

cmd/microcloud/ask.go
```diff
@@ -1274,7 +1280,9 @@ func (c *initConfig) askNetwork(sh *service.Handler) error {
 		}
 
 		if !supportsFAN {
-			proceedWithNoOverlayNetworking, err := c.asker.AskBoolWarn("FAN networking is not usable. Do you want to proceed with setting up an inoperable cluster?", false)
+			warning := "FAN networking is not usable"
```
minaelee, line 1283: Small nit - doesn't seem like FAN is meant to be capitalized as it's not an acronym. Per: https://canonical.com/blog/introducing-the-fan-simpler-container-networking & https://wiki.ubuntu.com/FanNetworking

minaelee (changes requested):

cmd/microcloud/ask.go
roosterfish: Good catch, fixed.

minaelee (approved):
lgtm, thanks!

PR #681, opened 2025-03-06
Backports (v2-edge)

roosterfish:
Backports until ea1c2dbdfc29d2556e5798bba3a31ad32c80f55d (March 5, 2025)


roosterfish:
Doc build currently failing due to https://github.com/canonical/lxd/pull/15120.

test/main.sh
```diff
@@ -71,30 +71,30 @@ cleanup() {
     set +x
   fi
 
-	for name in $(lxc list -c n -f csv micro); do
```
markylaing, line 74: General: Should we use `shfmt` (https://github.com/mvdan/sh) and have a linter for it?

test/suites/instances.sh
```diff
@@ -689,7 +695,7 @@ EOF
   check_instance_connectivity "c1" "c3" "0"
   check_instance_connectivity "c1" "c4" "0"
 
-  lxc exec micro01 -- lxc delete -f c1 c2 c3 c4
+  lxc exec micro01 -- lxc delete -f c1 c3 c4
```
markylaing, line 698: Also general: Something I've been thinking for a while in LXD tests is that it might be nice to separate clean up from the tests themselves. 

LXD test suite functions must have the format `test_{name}`, I'm not sure if that's the same here, but we could have `setup_{name}` and `teardown_{name}` as well.

cmd/microcloud/add.go
```diff
@@ -24,6 +24,7 @@ type cmdAdd struct {
 	flagSessionTimeout int64
 }
 
+// Command returns the subcommand to add new systems to MicroCloud.
```
markylaing, line 27: Not for now but I generally fix these lint errors by unexporting the functions, since they shouldn't need to be exported.

markylaing (approved):
LGTM just a couple of general questions :)

test/main.sh
roosterfish: Absolutely. We could definitely have way more tests as part of the linting stage. Same applies for Microcluster :)

test/suites/instances.sh
roosterfish: In case of the MicroCloud, the tests run in VMs anyway. 
You often see the `reset_systems x y z` line which restores the VMs so you always start with a fresh environment and don't leave traces of the previous test behind.

But there are some tests of course which create/remove some workload as part of the test execution.
In this specific case I assume its to clean up the deployed resources so that when adding another test you don't have to deal with left overs.

cmd/microcloud/add.go
roosterfish: Fair point, wasn't thinking about it when adding those. I have logged https://github.com/canonical/microcloud/issues/695.

PR #696, opened 2025-03-12
Update Go modules and bump Go min (for LXD)

simondeziel:


Makefile
```diff
@@ -85,6 +85,9 @@ update-gomod:
 
 	go mod tidy -go=$(GOMIN)
 
+	# Use the bundled toolchain that meets the minimum go version
+	go get toolchain@none
```
roosterfish, line 89: Neat, wasn't aware of it.

roosterfish (approved):

simondeziel:
@roosterfish once done, I'll backport the needed changes to `v2-edge` to close the handful of security issues found by dependabot (along with the broken PRs it turned up).

roosterfish:
> @roosterfish once merged, I'll backport the needed changes to `v2-edge` to close the handful of security issues found by dependabot (along with the broken PRs it turned up).

Thank you, the dependabot messages are coming from the snap scan (2/stable). Last time I checked we weren't directly affected by any of those as we don't use the impacted funcs/code. They warnings should go away after the 2.1.1 release.

simondeziel:
> Thank you, the dependabot messages are coming from the snap scan (2/stable). Last time I checked we weren't directly affected by any of those as we don't use the impacted funcs/code. The warnings should go away after the 2.1.1 release.

The snap scan would have been Trivy which unfortunately is quite noisy due to the infrequent snap rebuilds/publications. That said, there was https://github.com/canonical/microcloud/security/code-scanning/223 that needed more than a snap rebuild.

I doubt it has any actual security implications for MC users as in order to feed JSON to MC, you need admin access AFAICT. I feel like most of the time, it's easier/quicker to just bundle the fixed version rather than do the impact analysis :/

roosterfish:
> I doubt it has any actual security implications for MC users

This dependency comes from the Microcluster->LXD->OpenFGA chain of dependencies so we shouldn't use those bits overall:

```
$ go mod why -m github.com/go-jose/go-jose/v4
# github.com/go-jose/go-jose/v4
github.com/canonical/microcloud/microcloud/api
github.com/canonical/lxd/client
github.com/zitadel/oidc/v3/pkg/client/rp
github.com/go-jose/go-jose/v4
```

Meaning we should probably first update the LXD dependency in Microcluster.

roosterfish:
I have created a backport PR of your recent changes on Microcluster `v3` to `v2` https://github.com/canonical/microcluster/pull/347. Once this is merged we have to create a new release, then we can properly import it here to get rid of the jose sec warning.

roosterfish:
Just a follow up on this because I was wondering why the security alerts disappeared even before we created a new Microcluster release.

As we pin the LXD dependency to `stable-5.21`, we don't just rely on it being pulled through Microcluster, but instead update the dependency directly. 
And as LXD already has the latest packages in `stable-5.21`, we don't even have to create a Microcluster release to get them updated also in MicroCloud.

PR #698, opened 2025-03-14
client/proxy: error out if UseAuthProxy() is provided no HMAC and TLS verification disabled

simondeziel:


roosterfish (approved):
Thanks, this should prevent accidental misuse if used wrongly in the future.

simondeziel:
> Thanks, this should prevent accidental misuse if used wrongly in the future.

Yeah, that's the goal indeed. FYI, I zoomed onto this bit of code while reviewing the CodeQL alerts that were closed as `won't fix`. I reopened all of them and CodeQL immediately closed them all as `fixed`. So something to keep in mind when silencing them.

This leaves only 1 CodeQL alert marked as `false positive`: https://github.com/canonical/microcloud/security/code-scanning?query=is%3Aclosed+branch%3Amain+tool%3ACodeQL+resolution%3Adismissed
I do agree with your conclusion on that one and I appreciate you putting out a descriptive note along with the ignore :)

PR #699, opened 2025-03-17
GitHub: Use our own files changed check

roosterfish:


.github/actions/check-changes/action.yml
```diff
@@ -0,0 +1,35 @@
...
+          changes="$(git diff 'origin/${{ github.base_ref }}..HEAD' --name-only '${{ inputs.glob }}')"
+        else
+          echo "Checking changes made by the last (merge) commit"
+
+          changes="$(git diff HEAD^ --name-only '${{ inputs.glob }}')"
+        fi
+
+        if [ "$changes" != "" ]; then
+          echo "Found changes matching the glob pattern:"
+          echo "$changes"
```
simondeziel, line 30: The `--name-only` makes it reasonable to output. If that becomes too verbose, we can use `git diff --quiet` which just gives a verdict as the exit code.

simondeziel (dismissed):
LGTM, thanks. I've never used negative path globbing pattern with `git` but it's cool if it works like that!

roosterfish:
> I've never used negative path globbing pattern with git but it's cool if it works like that!

Pushed another change to include a link to the Git glob reference, thanks for the suggestion.

.github/actions/check-changes/action.yml
simondeziel: > If have added it explicitly because I remembered back then when adding the `dorny/paths-filter` action I wondered which files caused the pipeline to either run or not and I found it very helpful to have the info.

For sure.

> Also to see if the glob pattern works as expected.

I couldn't quickly find any doc about these glob patterns but it seems that multiple patterns are supported like that:

```
git diff --name-only HEAD^ ':!po/*' ':!doc/*'
```

simondeziel: Ah, just saw your update with a link to the doc, thanks!

simondeziel (approved):

PR #712, opened 2025-03-24
github: Update `golangci-lint` and migrate config to new format

kadinsayani:


.golangci.yaml
```diff
@@ -15,6 +15,10 @@ linters:
       - third_party$
       - builtin$
       - examples$
+    rules:
+      - linters:
+          - staticcheck
+        text: "ST1005:" # ST1005: error strings should not be capitalized (5585 occurences as of 2023-10-20)
```
roosterfish, line 21: Is this "5585 occurrences" coming from https://github.com/canonical/lxd/pull/15249?

roosterfish (approved):
Thanks!

.golangci.yaml
roosterfish: I'll merge it now to unblock the pipelines but please can you fix this in a follow up PR?

.golangci.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - whitespace
+  exclusions:
+    generated: lax
+    presets:
+      - comments
+      - common-false-positives
+      - legacy
+      - std-error-handling
+    paths:
+      - third_party$
```
roosterfish, line 15: Are those the standard paths from the reference file? Think we can remove them too.

Issue #562, opened 2024-12-18
Better text wrap in prompts during setup


MggMuggins:
This warning is rather long:
```
! Warning: Some systems are ineligible for distributed networking, which requires either an interface with no IPs assigned or a bridge. Continue anyway? (yes/no) [default=yes]: no
```

Lipgloss doesn't wrap to the next line, so a standard 80-char terminal will cut it off. I also ended up with this when I resized the window after letting it sit for a while:
![image](https://github.com/user-attachments/assets/ff5df6aa-100e-4fb0-b2f6-c48006b553e2)

roosterfish:
Just observed the same whilst working on https://github.com/canonical/microcloud/pull/666. 
It's not a full fix but at least it cuts the message in half for smaller terminals.

PR #716, opened 2025-03-27
CLI: Better wrap text

roosterfish:
This PR addresses most of the pain points reported in https://github.com/canonical/microcloud/issues/562. However it's not closing it fully as the upstream bubbletea library has an open issue about the rendering https://github.com/charmbracelet/bubbletea/issues/573.

To overcome the rendering issue for simple questions (as it can be seen in https://github.com/canonical/microcloud/issues/562), we don't have to use a bubbletea model, but instead can fall back to using the standard asker but with the applied lipgloss coloring which was once added in https://github.com/canonical/microcloud/pull/505.
The questions now render correctly when resizing the terminal window. Only tables remain to be an issue in case the window gets resized to something smaller than required by the table. But this is depending on the upstream issue.

Another effect of switching to the standard asker is that the text typed in the questions isn't color coded but instead using the standard white. But that seems to be acceptable given that the rendering now works as intended.

Also all the long lines of text within the bubbletea table model have been moved to use line wrapping from the muesli library (also used in the bubbletea examples) to properly render it in the given terminal window size.


markylaing (approved):
LGTM! Always good to have a PR that reduces lines ;)

PR #718, opened 2025-03-28
test/e2e: add script/Terraform recipe to test MicroCloud deployments

simondeziel:
This PR lays the foundation to deploy instances on a MicroCloud, run tests with those and finally tear them down.

The tests themselves will be added in subsequent PRs. The initial plan is to add connectivity and live migration tests.

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -1,0 +1,49 @@
+# End to end testing of MicroCloud
+
+Once MicroCloud is deployed, end to end testing can verify that everything works as it should.
+
+## Testing
+
+The [`run`](run) script will setup a dedicated project using the [Terraform provider for LXD](https://github.com/terraform-lxd/terraform-provider-lxd) and deploy a number of instances on the MicroCloud. Those instances will be a mix of containers and virtual machines.
+
+It is possible to override the variables defined in [`variables.tf`](variables.tf) by creating a file named `terraform.tfvars`:
+
+```sh
+# terraform.tfvars
+containers_per_host = 25
+vms_per_host = 5
+```
+
+Then execute the script and provide it the name of the LXD remote corresponding to the MicroCloud deployment to test (`mc` here):
+
+```sh
+./run mc
+```
+
+If the LXD remote to test is not configured yet, it's possible to add it with an identity token issued from a MicroCloud cluster member (`micro1` here):
+
+```sh
+# Create the needed group
+root@micro1:~# lxc auth group create e2e-testing
+Group e2e-testing created
+
+# Grant admin permission on server
+root@micro1:~# lxc auth group permission add e2e-testing server admin
+
+# Issue the identity token
+root@micro1:~# lxc auth identity create tls/e2e-testing
+TLS identity "tls/e2e-testing" (d16614b8-f39e-4ef5-8e9b-90764b61b660) pending identity token:
+eyJjbGllbnRfbmFtZSI6ImUyZS10ZXN0aW5nIiwiZmluZ2VycHJpbnQiOiIyNzY1OGNjYmRlNmRkZDcwNDliMzliNmY5NzA1MWIzMWVkMWQ0NDM2OTFjZGFjMWIxNmI1ODBjNmI4NzFiNmM2IiwiYWRkcmVzc2VzIjpbIjE3Mi4yNC4yNi4xODo4NDQzIiwiWzIwMDE6NDcwOmIxYzM6Nzk0Njo4NWYzOmQzYTg6ZDcxZTo3YjA1XTo4NDQzIl0sInNlY3JldCI6ImJjNjlmOTM4YzM5NjRhNTI1ZDI3OTlmMTEyM2Q2NWE5ZGExZDE1ZGFmNjVjZjlmNTViYjQ1ZDdmZjBiMjg1YmYiLCJleHBpcmVzX2F0IjoiMDAwMS0wMS0wMVQwMDowMDowMFoiLCJ0eXBlIjoiQ2xpZW50IGNlcnRpZmljYXRlIn0=
+
+# Add the freshly created identity to the freshly created group
+root@micro1:~# lxc auth identity group add tls/e2e-testing e2e-testing
+```
+
+This token can be used on the machine where the `run` script will be used:
+
+```sh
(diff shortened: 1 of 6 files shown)


test/e2e/run
```diff
@@ -0,0 +1,74 @@
...
+    if ! command -v terraform >/dev/null; then
+        echo "Missing command: terraform" >&2
+        exit 1
+    fi
+}
+
+
+apply() {
+    echo "terraform apply"
+    for i in 1 2 3; do
```
roosterfish, line 38: Simple but effective back-off. I like the pattern as you can reuse the index for both the back-off and to indicate the number of attempts.

In this case why is it required here to be able to do retries? Is it in case the remote LXD hasn't yet settled?

test/e2e/run
```diff
@@ -0,0 +1,74 @@
...
+}
+
+destroy() {
+    read -rp "Are you sure you want to destroy the environment? [y/N] " confirm
+    if [ "${confirm:-N}" != "y" ]; then
+        echo "Not destroying the environment"
+        return
+    fi
+
+    for i in 1 2 3; do
```
roosterfish, line 53: Worth putting the back-off code into its own func to be reused by `apply` and `destroy` and potentially other tf commands in the future?

roosterfish (changes requested):
Looks good, cannot wait having this executed as part of the regular test suite!

test/e2e/main.tf
```diff
@@ -0,0 +1,121 @@
...
+resource "lxd_instance" "e2e-vm" {
+  for_each = {
+    for _, v in local.vms : v.instance => v.target
+  }
+
+  name             = each.key
+  target           = each.value
+  type             = "virtual-machine"
+  remote           = var.remote
+  project          = var.project
```
MusicDin, line 87: `var.project` -> `lxd_project.e2e.name` to ensure there is a dependency between the VM resource and project

test/e2e/variables.tf
```diff
@@ -0,0 +1,38 @@
+variable "remote" {
+  description = "LXD remote to use"
+  type        = string
+  validation {
+    condition     = !can(regex(":", var.remote))
```
MusicDin, line 5: Could we simply to `!contains(var.remote, ":")`?

MusicDin (commented):
Looking good, thanks!

test/e2e/main.tf
```diff
@@ -0,0 +1,121 @@
...
+
+# Project
+resource "lxd_project" "e2e" {
+  name   = var.project
+  remote = var.remote
+}
+
+# Profile
+resource "lxd_profile" "e2e" {
+  name    = "e2e"
```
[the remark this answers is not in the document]
MusicDin: This is on is Terraform specific. Terraform creates resource dependency tree based on the referenced values between resources.

The `lxd_profile` currently sets project to `lxd_project.e2e.name`. This is a reference to the name field of the project resource e2e. Terraform will ensure the project is created before any resource that references it. In this particular case, profile is created only once the project is successfully created. Same applies when destroying, just in the reverse order - all resources referencing the project will be removed before the project.

If the `var.project` is used, there would be no reference between `lxd_project` and `lxd_profile`, which means Terraform would try to provision them concurrently. This may result the project not being available when profile is created, causing an error because project does not exist yet.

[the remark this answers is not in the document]
roosterfish: Ok thx.

test/e2e/README.md
```diff
@@ -0,0 +1,38 @@
...
+Then execute the script and provide it the name of the LXD remote corresponding to the MicroCloud deployment to test (`mc` here):
+
+```sh
+./run mc
+```
+
+If the LXD remote to test is not configured yet, it's possible to add it with a remote add token issue from a MicroCloud cluster member (`micro1` here):
+
+```sh
+root@micro1:~# lxc config trust add --name e2e-testing
```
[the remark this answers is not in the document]
simondeziel: Nice, however this requires quite a bit more config on the MC side. I hope I got it right :)

test/e2e/run
simondeziel: It's because Terraform is more efficient at parallelizing the operations it seems so it hits the issue I complained about where the RBD snapshot of the freshly downloaded image is not yet available on all cluster members.

That said, even when that LXD issue will be addressed, I think it's best to have some basic retry logic in there just in case there is a transient error like a network hiccup and such.

test/e2e/variables.tf
simondeziel: Would have been nice but [`contains()`](https://developer.hashicorp.com/terraform/language/functions/contains) only seems to accept lists, not strings.

simondeziel:
> Looks good, cannot wait having this executed as part of the regular test suite!

It should indeed be possible in the end :)

test/e2e/run
MusicDin: I would personally try without retry. If the new issues surface due to parallelization, we can work around them in the provider until they get eventually resolved in LXD. My point being that other users will also run into them, but may not be fond of retrying. What are your thoughts on that?

test/e2e/variables.tf
MusicDin: Ahh yeah, [strcontains()](https://developer.hashicorp.com/terraform/language/functions/strcontains) should be the right one.

test/e2e/run
simondeziel: The problem is that it reliably fails the first time :/ So the 1st retry is needed to paper over the bug. Longer term though I agree with you that we should aim at not doing automatic retries. Maybe only consider a retry with a manual confirmation to push through the occasional network hiccup and such.

MusicDin: Ahh, didn't know it reliably fails. But this occurs only with images, right? Maybe we should detect that particular error, and retry within the provider. But yeah, this discussion should be moved to tf provider.

Just FYI, the parallelism can be adjusted (`apply --parallelism=1` (defaults to 10)). Could be worth disabling it just to confirm the parallelism is the root cause, and to see the impact on deployment time.

simondeziel: Indeed, adding `-parallelism=1` to the apply worked.

simondeziel: @MusicDin any preferred way forward (`retry` loop, or `-parallelism=1`)?

roosterfish: I would guess fetching the image upfront also wouldn't do the trick?

MusicDin: > I guess fetching the image upfront also wouldn't do the trick?

Good point. If only images are causing issues, using `lxd_cached_image` resource before spinning up the instances might do the trick.

roosterfish: Yeah that was the idea, but tbh I don't know how LXD behaves inside a cluster when you copy that image.

MusicDin (approved):

simondeziel:
@MusicDin I ended up hardcoding the project name to better align with the README instruction. This will also make it easier to deal with `lxc` in connectivity tests, etc.

roosterfish (approved):
LGTM!

PR #727, opened 2025-04-01
doc: update linkcheck_timeout from 30 to 45 seconds

minaelee:
To avoid read timeout issues (different from rate limit timeout), set `linkcheck_timeout` to `45` seconds. Default when unset is `30`.


roosterfish (approved):
Thanks!

PR #728, opened 2025-04-02
e2e-testing: connectivity and rolling reboot tests

simondeziel:
Here's an example (local) run:

```
$ DESTROY=no ./run mc
Using overrides from terraform.tfvars:
--------------------------------------
# terraform.tfvars
containers_per_host = 1
vms_per_host = 1
--------------------------------------

Setting up e2e environment on mc
terraform apply
╷
│ Warning: Provider development overrides are in effect
│ 
│ The following provider development overrides are set in the CLI configuration:
│  - terraform-lxd/lxd in /home/sdeziel/git/terraform-provider-lxd
│ 
│ The behavior may therefore not match any released version of the provider and applying changes may cause the state to become incompatible with published releases.
╵
data.lxd_info.cluster: Reading...
lxd_project.e2e: Refreshing state... [name=e2e-testing]
data.lxd_info.cluster: Read complete after 0s
lxd_cached_image.ctn[0]: Refreshing state...
lxd_cached_image.vm[0]: Refreshing state...
lxd_profile.e2e: Refreshing state... [name=e2e-testing]
lxd_instance.e2e-vm["v02"]: Refreshing state... [name=v02]
lxd_instance.e2e-vm["v01"]: Refreshing state... [name=v01]
lxd_instance.e2e-vm["v03"]: Refreshing state... [name=v03]
lxd_instance.e2e-ctn["c01"]: Refreshing state... [name=c01]
lxd_instance.e2e-ctn["c03"]: Refreshing state... [name=c03]
lxd_instance.e2e-ctn["c02"]: Refreshing state... [name=c02]

No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

Outputs:

instances = {
  "ctns" = [
    {
      "instance" = "c01"
      "target" = "micro1"
    },
    {
      "instance" = "c02"
      "target" = "micro2"
    },
    {
      "instance" = "c03"
      "target" = "micro3"
    },
  ]
  "vms" = [
    {
      "instance" = "v01"
      "target" = "micro1"
    },
    {
      "instance" = "v02"
      "target" = "micro2"
    },
    {
      "instance" = "v03"
      "target" = "micro3"
    },
  ]
}
Setup done
Connectivity check:
 ● v02 => c01 (10.170.218.2): ✅
 ● v02 => v03 (10.170.218.3): ✅
 ● c01 => v02 (10.170.218.4): ✅
 ● c01 => c02 (10.170.218.5): ✅
 ● c03 => v01 (10.170.218.7): ✅
 ● c03 => v02 (10.170.218.4): ✅
DONE
Evacuating micro1
Waiting for micro1 to be rebooted                                               
Rebooting micro1
Waiting for micro1 to boot ....... DONE
Evacuating micro2                                                                
Waiting for micro2 to be rebooted                                               
Rebooting micro2
Waiting for micro2 to boot ...... DONE
Evacuating micro3                                                                
Waiting for micro3 to be rebooted                                               
Rebooting micro3
Waiting for micro3 to boot ....... DONE
Allow time for stopped instances to start back again .................... DONE   
Connectivity check:
 ● c03 => v03 (10.170.218.3): ✅
 ● c03 => v02 (10.170.218.4): ✅
 ● c03 => c02 (10.170.218.5): ✅
 ● v01 => c02 (10.170.218.5): ✅
 ● v01 => c03 (10.170.218.6): ✅
 ● v01 => v02 (10.170.218.4): ✅
 ● v03 => v02 (10.170.218.4): ✅
 ● v03 => v01 (10.170.218.7): ✅
 ● v03 => c01 (10.170.218.2): ✅
DONE
terraform destroy
Skipping destroy
Done
```

### DIFF
--- a/test/e2e/main.tf
+++ b/test/e2e/main.tf
@@ -90,6 +90,8 @@ resource "lxd_instance" "e2e-vm" {
   profiles         = [lxd_profile.e2e.name]
   image            = lxd_cached_image.vm[0].fingerprint
   wait_for_network = true
+  allow_restart    = true
+
   config = {
     "migration.stateful" = "true"
   }
--- a/test/e2e/main.tf
+++ b/test/e2e/main.tf
@@ -90,6 +90,9 @@ resource "lxd_instance" "e2e-vm" {
   profiles         = [lxd_profile.e2e.name]
   image            = lxd_cached_image.vm[0].fingerprint
   wait_for_network = true
+  config = {
+    "migration.stateful" = "true"
+  }
 }
 
 # Locals
--- a/test/e2e/main.tf
+++ b/test/e2e/main.tf
@@ -112,7 +112,7 @@ locals {
   vms = flatten([
     for index, cluster_member_name in local.cluster_member_names : [
       for i in range(var.vms_per_host) : {
-        instance = "vm${format("%02d", var.vms_per_host * index + i + 1)}"
+        instance = "v${format("%02d", var.vms_per_host * index + i + 1)}"
         target   = cluster_member_name
       }
     ]
--- a/test/e2e/main.tf
+++ b/test/e2e/main.tf
@@ -18,7 +18,7 @@ resource "lxd_profile" "e2e" {
   # Configuration
   config = {
     "limits.cpu"    = 1
-    "limits.memory" = "1GiB"
+    "limits.memory" = "384MiB"
   }
 
   # Devices
--- a/test/e2e/run
+++ b/test/e2e/run
@@ -27,6 +27,11 @@ init() {
         exit 1
     fi
 
+    if [ "$(clusterMembers ONLINE | wc -l)" != "$(clusterMembers ALL | wc -l)" ]; then
+        echo "Not all cluster members are online, aborting" >&2
+        exit 1
+    fi
+
     if ! command -v terraform >/dev/null; then
         echo "Missing command: terraform" >&2
         exit 1
@@ -99,6 +104,14 @@ instanceList() {
     lxc list --project e2e-testing -f csv -c n "${REMOTE}:"
 }
 
+clusterMembers() {
+    local state="${1}"
+    [ "${state}" = "ALL" ] && state=".*"
+    # XXX: cannot use the column number for the member state as if multiple
+    # roles are provided, the column number will change
+    lxc cluster list --project e2e-testing -f csv "${REMOTE}:" | awk -F, "/,${state},/ {print \$1}"
+}
+
 randomPick() {
     local count="${1}"
     shift
@@ -143,9 +156,36 @@ connectivity() {
     echo "DONE"
 }
 
+evacuation() {
+    local member
+
+    for member in $(clusterMembers ONLINE); do
+        echo "Evacuating ${member}"
+        lxc cluster evacuate --project e2e-testing --force "${REMOTE}:${member}"
+
+        echo "Waiting for ${member} to be rebooted"
+        if [ -x "./reboot" ]; then
+            ./reboot "${REMOTE}" "e2e-testing" "${member}"
+        else
+            read -rp "Please reboot ${member} and press enter to continue once it is back online"
+        fi
+
+        lxc cluster restore --project e2e-testing --force "${REMOTE}:${member}"
+    done
+
+    echo -n "Allow time for stopped instances to start back again "
+    for _ in $(seq 20); do
+        echo -n "."
+        sleep 1
+    done
+    echo " DONE"
+}
+
 init
 setup
 
+connectivity
+evacuation
 connectivity
 
 destroy
--- a/test/e2e/run
+++ b/test/e2e/run
@@ -30,6 +30,15 @@ init() {
         echo "Missing command: terraform" >&2
         exit 1
     fi
+
+    # Show overrides
+    if [ -s ./terraform.tfvars ]; then
+        echo "Using overrides from terraform.tfvars:"
+        echo "--------------------------------------"
+        cat ./terraform.tfvars
+        echo "--------------------------------------"
+        echo
+    fi
 }
 
 retry() {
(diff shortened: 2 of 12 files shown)


simondeziel:
For reviewers, the CI bits related to this PR start here: https://github.com/canonical/microcloud/actions/runs/14253651205/job/39952160052?pr=728#step:3:5531

test/e2e/main.tf
```diff
@@ -90,6 +90,11 @@ resource "lxd_instance" "e2e-vm" {
   profiles         = [lxd_profile.e2e.name]
   image            = lxd_cached_image.vm[0].fingerprint
   wait_for_network = true
+  allow_restart    = true
+
+  config = {
+    "migration.stateful" = "true"
```
MusicDin, line 96: I would just add the comment next to `allow_restart` why is it required. 

We added `allow_restart` to Terraform to support non-stateful migrations and instance renaming, in which case the provider is allowed temporary stop the instance to complete those operations. Therefore, having a `migration.stateful` flag next to `allow_restart` seems a bit odd to me. Maybe `allow_restart` is not needed at all?

MusicDin:
To me it already looks good. I also like that despite having the retries the remote storage issue is clearly visible in the logs 🚀

test/e2e/main.tf
simondeziel: I added the `allow_restart` after aborting a test run in the middle of a rolling reboot. When I re-ran from there, Terraform wanted to put the evacuated containers on their respective `target` but couldn't as that implied downtime.

So this was an edge case because an aborted rolling reboot will still require the user to `lxc cluster restore` the evacuated member (thus fixing the container placement) before re-running the script. The check to require all cluster members to be `ONLINE` came after.

With all that said, I am no longer sure `allow_restart` make much sense but then if Terraform wants to relocate a container, it shouldn't matter for it to go offline briefly as it's a test instance anyway.

However, maybe you are saying that we should consider this a bug and instead have Terraform error out if it meets with a misplaced instance?

test/e2e/run
```diff
@@ -80,10 +80,58 @@ setup() {
...
+}
+
+checkConnectivity() {
+    local src="${1}"
+    local dst="${2}"
+    local dstIPv4
+    dstIPv4="$(getIPv4 "${dst}")"
+
+    echo -n " ● ${src} => ${dst} (${dstIPv4}):"
+    # Using bash /dev/tcp to check for connectivity as it is available in every Ubuntu images event the minimal ones.
```
roosterfish, line 137: `even`.

test/e2e/main.tf
```diff
@@ -18,7 +18,7 @@ resource "lxd_profile" "e2e" {
   # Configuration
   config = {
     "limits.cpu"    = 1
-    "limits.memory" = "1GiB"
+    "limits.memory" = "384MiB"
```
roosterfish, line 21: Just out of interest, what is this number based on?

test/suites/e2e.sh
```diff
@@ -0,0 +1,62 @@
...
+    validate_system_microceph "${m}" 1 disk2
+    validate_system_microovn "${m}"
+  done
+
+  # e2e test with Terraform
+  cd e2e || exit 1
+  TOKEN="$(lxc exec micro01 -- lxc config trust add --quiet --name e2e)"
+  MICRO01_IP="$(lxc list -f csv -c 4 micro01 | awk '/enp5s0/ {print $1}')"
+  REMOTE_ADDRESS="https://${MICRO01_IP}:8443"
+  lxc remote add mc "${REMOTE_ADDRESS}" --token "${TOKEN}"
```
roosterfish, line 46: Worth to cleanup the remote (and trust) at the end of the test from the runners LXD?

roosterfish (changes requested):
Looks great, just a few comments.

As we now launch instances in both the `instances` and `e2e` suite, do you think it would make sense to deprecate the corresponding tests (including the network pings) in the `instances` suite in favor of `e2e`?

test/e2e/main.tf
simondeziel: It makes VM live-migration faster.

test/suites/e2e.sh
simondeziel: I thought of that too but it makes repeated testing quite a bit more annoying. How about I add a prompt to do this in the `destroy()` function when it's configured to prompt or destroy automatically?

test/e2e/main.tf
roosterfish: I mean can it also be 383 MiB? Or is there something more to it? :)

simondeziel:
> As we now launch instances in both the `instances` and `e2e` suite, do you think it would make sense to deprecate the corresponding tests (including the network pings) in the `instances` suite in favor of `e2e`?

Connectivity checks in the `e2e` variant are super cheap as nothing needs to be installed. How about I rework (in a subsequent PR) the `instances` ones to also use `bash`'s TCP support?

This would avoid `cloud-init` having to update the package definitions and install `iputils-ping`.

test/suites/e2e.sh
simondeziel: After thinking some more, the `run` script can be run on a local system so using the Unix socket. It is also possible for it to use an existing remote that didn't need any token to begin with.

This makes cleanup a bit hard. I'd want a `lxc config trust whoami` kind  of command to know what to cleanup if anything.

roosterfish:
> How about I rework (in a subsequent PR) the `instances` ones to also use `bash`'s TCP support?

Yeah that sounds like a good plan, thank you.

test/suites/e2e.sh
roosterfish: I see. It just caught my eye as I remember this cleanup (in the link I posted) caused trouble in the past because it wasn't there. But in this case it was part of the `validate_system_lxd` func which gets executed lots of times throughout the test suite execution without resetting the outer runner.

Let's leave it for now.

roosterfish (approved):

PR #744, opened 2025-04-09
Faster connectivity tests

simondeziel:


test/suites/instances.sh
```diff
@@ -33,19 +33,18 @@ check_instance_connectivity() {
...
   done
 
-  echo "Test connectivity to lxdbr0"
+  echo "Test connectivity to lxdbr0 via DNS, IPv4 and IPv6"
   IPV4_GW="$(lxc network get lxdbr0 ipv4.address | cut -d/ -f1)"
   IPV6_GW="$(lxc network get lxdbr0 ipv6.address | cut -d/ -f1)"
   for m in "${instance_1}" "${instance_2}" ; do
-    lxc exec micro01 -- lxc exec "${m}" -- ping -nc1 -w5 -4 "${IPV4_GW}"
-    lxc exec micro01 -- lxc exec "${m}" -- ping -nc1 -w5 -6 "${IPV6_GW}"
+    lxc exec micro01 -- lxc exec "${m}" -- bash -cex "for dst in _gateway ${IPV4_GW} ${IPV6_GW}; do grep -qm1 ^SSH- < /dev/tcp/\$dst/22; done"
```
tomponline, line 40: do you need to wrap this with the `time` command so it doesnt hang if the the dst is non responsive?

simondeziel: Fortunately, it doesn't hang indefinitely:

```
$ time grep -m1 ^SSH- < /dev/tcp/1.1.1.1/22
bash: connect: Connection timed out
bash: /dev/tcp/1.1.1.1/22: Connection timed out

real	2m16.171s
user	0m0.001s
sys	0m0.001s
```

but to fail quickly, I am now using `timeout 5 bash ...`. I tried using `timeout 5 grep` but that wouldn't work as it's `bash` that's hanging while trying to open the TCP connection :)

roosterfish (approved):
LGTM

Issue #670, opened 2025-03-03
Enable usage of LXD 6 track


roosterfish:
Right now MicroCloud requires uses to run the LXD 5.21 LTS track. However it might be desired to still use MicroCloud using the latest version of LXD (non LTS).

roosterfish:
I'll now mark this as a bug as you might end up in a lock when upgrading LXD to 6.x after forming the MicroCloud as this will hinder MicroCloud from doing any further operation as the version of LXD doesn't anymore matches 5.21.x.

PR #752, opened 2025-04-16
Service: Allow MicroCloud to be used with LXD 5.21 or newer

roosterfish:
Fixes https://github.com/canonical/microcloud/issues/670

Furthermore this extends the test matrix to also include LXD version 6.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -166,6 +166,9 @@ jobs:
           - "5.21/edge"
           - "6/edge"
         microcloud: ["latest/edge"]
+        exclude:
+          # Exclude LXD 6 for regular development pipelines to speed up the result cycle.
+          - lxd: ${{ github.event_name != 'schedule' && '6/edge' }}
     steps:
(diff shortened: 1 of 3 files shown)


roosterfish:
This doubles the chance for the e2e tests to fail as we this PR adds two more runs using LXD `latest/edge`.

simondeziel (approved):
LGTM but instead of always doubling the testing matrix, maybe we could have the `6/edge` only only trigger during the daily tests rather than against every PRs?

roosterfish:
Add a warning message if a non-LTS LXD version is used.

roosterfish:
> LGTM but instead of always doubling the testing matrix, maybe we could have the `6/edge` only only trigger during the daily tests rather than against every PRs?

Thanks for the suggestion. I found a neat way to skip the tests using the `matrix.exclude` keyword to speed up the result cycle and skip LXD 6 for regular pipelines.

simondeziel (approved):

PR #754, opened 2025-04-16
github: Remove unsed step that installs dqlite dev

roosterfish:
We already build the MicroCloud binary with our own version of dqlite. This step is not required anymore and was missed when swichting the build process.


simondeziel (approved):
Great, thanks!

tomponline (approved):

PR #756, opened 2025-04-24
doc: update config for move to documentation.ubuntu.com

minaelee:
The move to documentation.ubuntu.com/microcloud is approved and is done through the ReadTheDocs dashboard (by adding it as a subproject of the Ubuntu documentation library project). This PR updates the Sphinx custom_conf.py to reflect the change in URL and needs to be done along with the move. Otherwise, static resources won't fetch correctly.


roosterfish (approved):
Great news, thanks for this!

PR #777, opened 2025-05-06
golangci-lint: enable perfsprint, prealloc and testifylint linters

simondeziel:
Only `revive` is missing from the linters list use by LXD. I kept it for another day ;)

### DIFF
--- a/cmd/microcloud/preseed.go
+++ b/cmd/microcloud/preseed.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -241,11 +242,11 @@ func (c *initConfig) RunPreseed(cmd *cobra.Command) error {
 	}
 
 	if !status.Ready && !c.bootstrap && initiator {
-		return fmt.Errorf("MicroCloud isn't yet initialized and cannot be the initiator")
+		return errors.New("MicroCloud isn't yet initialized and cannot be the initiator")
 	}
 
 	if status.Ready && !initiator {
-		return fmt.Errorf("MicroCloud is already initialized and can only be the initiator")
+		return errors.New("MicroCloud is already initialized and can only be the initiator")
 	}
 
 	systems, err := config.Parse(s, c, services)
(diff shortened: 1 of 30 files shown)


cmd/microcloud/preseed_test.go
```diff
@@ -333,7 +333,7 @@ func (s *preseedSuite) Test_preseedMatchDisksMemory() {
 	results, err := filter.Match(disks)
 	s.NoError(err)
 
-	s.Equal(len(results), 1)
+	s.Equal(1, len(results))
```
roosterfish: This commit doesn't seem to be required as the line gets rewritten anyway to `s.Len` in the next commit.

.github/actions/system-test/action.yml
```diff
@@ -20,7 +20,9 @@ runs:
       run: |
         # When br_netfilter is enabled, the multicast traffic that passes the native LXD bridge
         # will get masqueraded too which breaks the multicast discovery.
-        sudo rmmod br_netfilter
+        if lsmod | grep -qw ^br_netfilter; then
```
roosterfish, line 23: Interesting why it appears again in this PR. The lines actually have been changed on `main` already.

cmd/microcloud/preseed.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
```
roosterfish, line 5: Can you please squash this into the other commit where you replace the actual funcs? There are a few more of these commits after this one.

roosterfish (changes requested):
Thanks for this, only a few small comments.

cmd/microcloud/preseed_test.go
simondeziel: That's a side effect of running tools to make those changes (`golangci-lint fix ./...`) but I'm happy to squash them.

.github/actions/system-test/action.yml
simondeziel: It's gone now that I've rebased.

cmd/microcloud/preseed.go
simondeziel: I forgot to squash quite a few commits from that PR, sorry about that. Should be better sliced now.

roosterfish: Thanks!

roosterfish (approved):

Issue #653, opened 2025-02-20
Reoccuring pipeline failures


roosterfish:
This issue should track reoccurring pipeline errors which are observed from time to time:

- [x] `Error: System "micro02" failed to join the cluster: Failed to update cluster status of services: Put "https://10.69.231.229:9443/1.0/services": EOF`
- [ ] `lxc exec micro01 -- lxc launch ubuntu-minimal-daily:22.04 c1 -c limits.memory=512MiB -d root,size=2GiB -s remote -n default
Launching c1
Error: Failed instance creation: Fetch project database object: Failed to fetch from "projects" table: Failed to fetch from "projects" table: Failed to fetch from "projects" table: sql: transaction has already been committed or rolled back`

PR #799, opened 2025-05-22
cmd/tui/handler: avoid panic at runtime due to trying to copy into an inner nil slice

edlerd:
# Done
- cmd/tui/handler: avoid panic at runtime due to trying to copy into an inner nil slice

see https://github.com/canonical/microcloud/actions/runs/15183238063/job/42697492228 for a failing test run due to panick at runtime

image

### DIFF
--- a/cmd/tui/console.go
+++ b/cmd/tui/console.go
@@ -144,12 +144,12 @@ func (c *testConsole) parseInput(handler *InputHandler) error {
 			}
 
 			// Sanity check that the table didn't start with more rows than we are going to expect.
-			if len(handler.getAllRows()) > count {
-				return fmt.Errorf("Table has more rows (%d) than expected (%d)", len(handler.getAllRows()), count)
+			if handler.countAllRows() > count {
+				return fmt.Errorf("Table has more rows (%d) than expected (%d)", handler.countAllRows(), count)
 			}
 
 			// Wait until we receive the expected row count.
-			for len(handler.getAllRows()) < count {
+			for handler.countAllRows() < count {
 				time.Sleep(300 * time.Millisecond)
 			}
 		} else if filter != "" {
--- a/cmd/tui/handler.go
+++ b/cmd/tui/handler.go
@@ -22,8 +22,7 @@ type InputHandler struct {
 	// testMode is set to true if the handler is initialized in test mode with PrepareTestAsker.
 	testMode bool
 
-	table   *selectableTable
-	tableMu sync.Mutex
+	table *selectableTable
 
 	activeMu sync.RWMutex
 	active   bool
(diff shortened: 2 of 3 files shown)


roosterfish:
Thanks for tackling this. I have recorded a few of those errors also in https://github.com/canonical/microcloud/issues/653. 
There already is a section in the code that initializes the slice and I think it's a race on the `table.rawRows` field. 
Sometimes it is using locking but sometimes it doesn't.

roosterfish:
@edlerd I have created https://github.com/canonical/microcloud/issues/807 to track the progress. Do you want to look into it? Please also check my comment above.

edlerd:
> @edlerd I have created https://github.com/canonical/microcloud/issues/807 to track the progress. Do you want to look into it? Please also check my comment above.

Sure I opened this to fix it :)

> There already is a section in the code that initializes the slice and I think it's a race on the table.rawRows field.

Looking at the context of the change, I can't see how this would get initialized in another code path. The `allRows` object is defined within the function, and we need to allocate memory for the items that are copied in. I think this only ever passes without panic in case the passed in `i.table.rawRows` has no nested arrays. See below for full context of the function.

image

roosterfish:
> Looking at the context of the change, I can't see how this would get initialized

I am talking about the struct's `rawRows`, not the function local `allRows`. Almost the same name though :)

There are the the two `handleRemoveEvent` and `handleInsertEvent` which both update the struct's `rawRows` without obtaining the `tableMu` lock. So I have the suspicion that one of those two funcs is called whilst the `getAllRows` is called which is causing the panic.

edlerd:
@roosterfish unfortunately we don't have access to the table mutext in the event handlers. But I think there is an easy way out. We only ever interest the number of the table rows, so we can avoid the copy all together and just return the count in our helper function.

roosterfish:
Good catch that we don't even require the data from `getAllRows` but instead we just need the number. 

Now I think we don't even need to obtain the `tableMu` lock. Also the only other user is the `Render` func which isn't called concurrently as we only ever render one table at a time. 
So we can get rid of `tableMu sync.Mutex` completely on the handler.

Instead we should have a mutex on the `selectableTable` whenever we add/delete rows to the table. And then we can have another func on `selectableTable` which obtains the lock and gets the `len` of the rows. This func can then be called by `countAllRows` of the handler. 
This way we always have the right information about `len`.

edlerd:
> Now I think we don't even need to obtain the `tableMu` lock. Also the only other user is the `Render` func which isn't called concurrently as we only ever render one table at a time. So we can get rid of `tableMu sync.Mutex` completely on the handler.
> 
> Instead we should have a mutex on the `selectableTable` whenever we add/delete rows to the table. And then we can have another func on `selectableTable` which obtains the lock and gets the `len` of the rows. This func can then be called by `countAllRows` of the handler. This way we always have the right information about `len`.

Right, adjusted the PR to the new mutex inside of the selectable table.

roosterfish (approved):
Thanks!

PR #801, opened 2025-05-26
Cluster manager reverse tunnel

edlerd:
# Done
- added reverse tunnel to cluster manager
- proxy endpoint to expose the LXD API through the tunnel
- expect an authorization / bearer header for auth, reply with the session that the client can cache
- changes the CLI/config key naming from kebab-case to snake_case

Ready for review.

### DIFF
--- a/api/cluster_manager.go
+++ b/api/cluster_manager.go
@@ -49,7 +49,12 @@ func clusterManagerGet(state types.State, r *http.Request) types.Response {
 		return types.SmartError(err)
 	}
 
-	clusterManager, clusterManagerConfig, err := database.LoadClusterManager(state, r.Context(), name)
+	clusterManager, err := database.LoadClusterManager(state, r.Context(), name)
+	if err != nil {
+		return types.SmartError(err)
+	}
+
+	clusterManagerConfig, err := database.LoadClusterManagerConfigs(state, r.Context(), clusterManager.ID)
 	if err != nil {
 		return types.SmartError(err)
 	}
@@ -97,7 +102,7 @@ func clusterManagerPost(sh *service.Handler) func(state types.State, r *http.Req
 		}
 
 		// ensure cluster manager is not already configured
-		existingClusterManager, _, err := database.LoadClusterManager(state, r.Context(), args.Name)
+		existingClusterManager, err := database.LoadClusterManager(state, r.Context(), args.Name)
 		if err != nil {
 			if api.StatusErrorCheck(err, http.StatusNotFound) {
 				// ignore, this is the expected path
@@ -163,7 +168,7 @@ func clusterManagerPut(state types.State, r *http.Request) types.Response {
 		return types.SmartError(err)
 	}
 
-	clusterManager, _, err := database.LoadClusterManager(state, r.Context(), name)
+	clusterManager, err := database.LoadClusterManager(state, r.Context(), name)
 	if err != nil {
 		return types.SmartError(err)
 	}
@@ -191,8 +196,16 @@ func clusterManagerPut(state types.State, r *http.Request) types.Response {
 		}
 	}
 
-	if args.UpdateInterval != nil {
-		err = database.StoreClusterManagerConfig(state, r.Context(), name, database.UpdateIntervalSecondsKey, *args.UpdateInterval)
+	if args.UpdateIntervalSeconds != nil {
+		err = database.StoreClusterManagerConfig(state, r.Context(), name, database.UpdateIntervalSecondsKey, *args.UpdateIntervalSeconds)
+		if err != nil {
+			return types.SmartError(err)
+		}
+	}
+
+	if args.ReverseTunnel != nil {
+		reverseTunnelValue := strconv.FormatBool(*args.ReverseTunnel)
+		err = database.StoreClusterManagerConfig(state, r.Context(), name, database.ReverseTunnelKey, reverseTunnelValue)
 		if err != nil {
 			return types.SmartError(err)
 		}
@@ -215,7 +228,7 @@ func clusterManagerDelete(sh *service.Handler) func(state types.State, r *http.R
 			return types.SmartError(err)
 		}
 
-		clusterManager, _, err := database.LoadClusterManager(state, r.Context(), name)
+		clusterManager, err := database.LoadClusterManager(state, r.Context(), name)
 		if err != nil {
 			return types.SmartError(err)
 		}
--- a/api/types/cluster_manager.go
+++ b/api/types/cluster_manager.go
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"net/http"
 	"time"
 )
 
@@ -54,7 +55,11 @@ type ClusterManagerPut struct {
 
 	// Interval in seconds to send status messages to the cluster manager
 	// Example: 60
-	UpdateInterval *string `json:"update_interval" yaml:"update_interval"`
+	UpdateIntervalSeconds *string `json:"update_interval_seconds" yaml:"update_interval_seconds"`
+
+	// Enables or disables the reverse tunnel to the cluster manager
+	// Example: true, false
+	ReverseTunnel *bool `json:"reverse_tunnel" yaml:"reverse_tunnel"`
 }
 
 // StatusDistribution represents the distribution of items.
@@ -99,4 +104,22 @@ type ClusterManagerJoin struct {
 	ClusterName        string `json:"cluster_name" yaml:"cluster_name"`
 	ClusterCertificate string `json:"cluster_certificate" yaml:"cluster_certificate"`
 	Token              string `json:"token" yaml:"token"`
+}
+
+// ClusterManagerTunnelRequest represents the request received through the tunnel.
+type ClusterManagerTunnelRequest struct {
+	UUID    string      `json:"uuid"`
+	Method  string      `json:"method"`
+	Path    string      `json:"path"`
+	Headers http.Header `json:"headers"`
+	Body    []byte      `json:"body"`
+}
+
+// ClusterManagerTunnelResponse represents the response sent through the tunnel.
+type ClusterManagerTunnelResponse struct {
+	UUID    string         `json:"uuid"`
+	Status  int            `json:"status"`
+	Headers http.Header    `json:"headers"`
+	Cookies []*http.Cookie `json:"cookies"`
+	Body    []byte         `json:"body"`
 }
--- a/client/cluster_manager_client.go
+++ b/client/cluster_manager_client.go
@@ -10,10 +10,12 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/version"
+	"github.com/gorilla/websocket"
 
 	"github.com/canonical/microcloud/microcloud/api/types"
 	"github.com/canonical/microcloud/microcloud/database"
@@ -84,6 +86,26 @@ func (c *ClusterManagerClient) Delete(clusterCert *shared.CertInfo) error {
 	return err
 }
 
+// ConnectTunnel establishes a WebSocket connection to the cluster manager for reverse tunneling.
+func (c *ClusterManagerClient) ConnectTunnel(ctx context.Context, clusterCert *shared.CertInfo) (*websocket.Conn, error) {
+	tlsConfig, address, err := c.getTlsConfig(clusterCert)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get TLS config: %w", err)
+	}
+
+	dialer := websocket.Dialer{
+		TLSClientConfig: tlsConfig,
+	}
+
+	u := url.URL{Scheme: "wss", Host: address, Path: "/1.0/remote-cluster/ws"}
+	conn, _, err := dialer.DialContext(ctx, u.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return conn, nil
+}
+
 func (c *ClusterManagerClient) craftRequest(method string, path string, reqBody io.Reader) (*http.Request, error) {
 	url := "https://remote" + path // remote is a placeholder, real address will be set in sendRequest
 	req, err := http.NewRequest(method, url, reqBody)
@@ -122,7 +144,19 @@ func (c *ClusterManagerClient) sendRequest(clusterCert *shared.CertInfo, req *ht
 
 func (c *ClusterManagerClient) getHTTPClient(clusterCert *shared.CertInfo) (*http.Client, string, error) {
 	client := &http.Client{}
+	tlsConfig, address, err := c.getTlsConfig(clusterCert)
+	if err != nil {
+		return nil, "", fmt.Errorf("Failed to get TLS config: %w", err)
+	}
 
+	client.Transport = &http.Transport{
+		TLSClientConfig: tlsConfig,
+	}
+
+	return client, address, nil
+}
+
+func (c *ClusterManagerClient) getTlsConfig(clusterCert *shared.CertInfo) (*tls.Config, string, error) {
 	var address string
 	var remoteCert *x509.Certificate
 	var err error
@@ -171,9 +205,5 @@ func (c *ClusterManagerClient) getHTTPClient(clusterCert *shared.CertInfo) (*htt
 		return &cert, nil
 	}
 
-	client.Transport = &http.Transport{
-		TLSClientConfig: tlsConfig,
-	}
-
-	return client, address, nil
+	return tlsConfig, address, nil
 }
--- a/cmd/microcloud/cluster_manager.go
+++ b/cmd/microcloud/cluster_manager.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/canonical/lxd/shared/api"
 	cli "github.com/canonical/lxd/shared/cmd"
+	"github.com/canonical/lxd/shared/validate"
 	"github.com/canonical/microcluster/v3/microcluster"
 	microTypes "github.com/canonical/microcluster/v3/microcluster/types"
 	"github.com/spf13/cobra"
@@ -201,11 +202,11 @@ func (c *cmdClusterManagerGet) command() *cobra.Command {
 	cmd.Use = "get"
 	cmd.Short = "Get specific cluster manager configuration by key."
 	cmd.Example = cli.FormatSection("", `microcloud cluster-manager get addresses
-microcloud cluster-manager get certificate-fingerprint
-microcloud cluster-manager get update-interval-seconds
-microcloud cluster-manager get status-last-success-time
-microcloud cluster-manager get status-last-error-time
-microcloud cluster-manager get status-last-error-response`)
+microcloud cluster-manager get certificate_fingerprint
+microcloud cluster-manager get update_interval_seconds
+microcloud cluster-manager get status_last_success_time
+microcloud cluster-manager get status_last_error_time
+microcloud cluster-manager get status_last_error_response`)
 
 	cmd.RunE = c.run
 
@@ -233,20 +234,26 @@ func (c *cmdClusterManagerGet) run(_ *cobra.Command, args []string) error {
 	switch key {
 	case "addresses":
 		fmt.Printf("%s\n", strings.Join(clusterManager.Addresses, ", "))
-	case "certificate-fingerprint":
+	case "certificate_fingerprint":
 		fmt.Printf("%s\n", clusterManager.CertificateFingerprint)
-	case "update-interval-seconds":
+	case "update_interval_seconds":
 		value, ok := clusterManager.Config[database.UpdateIntervalSecondsKey]
 		if ok {
 			fmt.Printf("%s\n", value)
 		}
 
-	case "status-last-success-time":
+	case "status_last_success_time":
 		fmt.Printf("%s\n", clusterManager.StatusLastSuccessTime)
-	case "status-last-error-time":
+	case "status_last_error_time":
 		fmt.Printf("%s\n", clusterManager.StatusLastErrorTime)
-	case "status-last-error-response":
+	case "status_last_error_response":
 		fmt.Printf("%s\n", clusterManager.StatusLastErrorResponse)
+	case "reverse_tunnel":
+		value, ok := clusterManager.Config[database.ReverseTunnelKey]
+		if ok {
+			fmt.Printf("%s\n", value)
+		}
+
 	default:
 		return errors.New("Invalid key")
 	}
@@ -265,8 +272,9 @@ func (c *cmdClusterManagerSet) command() *cobra.Command {
 	cmd.Use = "set"
 	cmd.Short = "Set specific cluster manager configuration key."
 	cmd.Example = cli.FormatSection("", `microcloud cluster-manager set addresses example.com:8443
-microcloud cluster-manager set certificate-fingerprint abababababababababababababababababababababababababababababababab
-microcloud cluster-manager set update-interval-seconds 50`)
+microcloud cluster-manager set certificate_fingerprint abababababababababababababababababababababababababababababababab
+microcloud cluster-manager set update_interval_seconds 50
+microcloud cluster-manager set reverse_tunnel true`)
 
 	cmd.RunE = c.run
 
@@ -291,10 +299,18 @@ func (c *cmdClusterManagerSet) run(_ *cobra.Command, args []string) error {
 	switch key {
 	case "addresses":
 		payload.Addresses = []string{value}
-	case "certificate-fingerprint":
+	case "certificate_fingerprint":
 		payload.CertificateFingerprint = &value
-	case "update-interval-seconds":
-		payload.UpdateInterval = &value
+	case "update_interval_seconds":
+		payload.UpdateIntervalSeconds = &value
+	case "reverse_tunnel":
+		err := validate.IsBool(value)
+		if err != nil {
+			return errors.New("Invalid value for reverse_tunnel, expected 'true' or 'false'")
+		}
+
+		enabled := value == "true" || value == "yes" || value == "on" || value == "1"
+		payload.ReverseTunnel = &enabled
 	default:
 		return errors.New("Invalid key")
 	}
@@ -317,7 +333,7 @@ func (c *cmdClusterManagerUnset) command() *cobra.Command {
 	cmd := &cobra.Command{}
 	cmd.Use = "unset"
 	cmd.Short = "Unset specific cluster manager configuration key."
-	cmd.Example = cli.FormatSection("", `microcloud cluster-manager unset update-interval-seconds`)
+	cmd.Example = cli.FormatSection("", `microcloud cluster-manager unset update_interval_seconds`)
 
 	cmd.RunE = c.run
 
@@ -339,9 +355,13 @@ func (c *cmdClusterManagerUnset) run(_ *cobra.Command, args []string) error {
 	payload := types.ClusterManagerPut{}
 
 	switch key {
-	case "update-interval-seconds":
-		emptyString := ""
-		payload.UpdateInterval = &emptyString
+	case "update_interval_seconds":
+		payload.UpdateIntervalSeconds = new("")
+
+	case "reverse_tunnel":
+		disabled := false
+		payload.ReverseTunnel = &disabled
+
 	default:
 		return errors.New("Invalid key")
 	}
--- a/cmd/microcloudd/cluster_manager_task.go
+++ b/cmd/microcloudd/cluster_manager_task.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"net/url"
 	"slices"
 	"time"
@@ -11,6 +12,7 @@ import (
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
 	microTypes "github.com/canonical/microcluster/v3/microcluster/types"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/canonical/microcloud/microcloud/api/types"
 	"github.com/canonical/microcloud/microcloud/client"
@@ -19,8 +21,8 @@ import (
 )
 
 // SendClusterManagerStatusMessageTask starts a go routine, that sends periodic status messages to cluster manager.
-func SendClusterManagerStatusMessageTask(ctx context.Context, sh *service.Handler, s microTypes.State) {
-	go func(ctx context.Context, sh *service.Handler, s microTypes.State) {
+func SendClusterManagerStatusMessageTask(ctx context.Context, g *errgroup.Group, sh *service.Handler, s microTypes.State) {
+	g.Go(func() error {
 		ticker := time.NewTicker(database.UpdateIntervalDefaultSeconds * time.Second)
 		defer ticker.Stop()
 
@@ -33,15 +35,15 @@ func SendClusterManagerStatusMessageTask(ctx context.Context, sh *service.Handle
 				}
 
 			case <-ctx.Done():
-				return // exit the loop and close the go routine
+				return nil // exit the loop and close the go routine
 			}
 		}
-	}(ctx, sh, s)
+	})
 }
 
 func sendClusterManagerStatusMessage(ctx context.Context, sh *service.Handler, s microTypes.State) time.Duration {
 	logger.Debug("Starting sendClusterManagerStatusMessage")
-	var nextUpdate time.Duration = 0
+	var nextUpdate = time.Duration(database.UpdateIntervalDefaultSeconds) * time.Second
 
 	cloud := sh.Services[types.MicroCloud].(*service.CloudService)
 	isInitialized, err := cloud.IsInitialized(ctx)
@@ -55,9 +57,9 @@ func sendClusterManagerStatusMessage(ctx context.Context, sh *service.Handler, s
 		return nextUpdate
 	}
 
-	clusterManager, clusterManagerConfig, err := database.LoadClusterManager(s, ctx, database.ClusterManagerDefaultName)
+	clusterManager, err := database.LoadClusterManager(s, ctx, database.ClusterManagerDefaultName)
 	if err != nil {
-		if err.Error() == "Cluster manager not found" {
+		if api.StatusErrorCheck(err, http.StatusNotFound) {
 			logger.Debug("Cluster manager not configured, skipping status message")
 			return nextUpdate
 		}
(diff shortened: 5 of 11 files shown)


roosterfish:
@edlerd can this PR be closed?

edlerd:
> @edlerd can this PR be closed?

We might want to use this approach in at a later time, so I'd keep it open in draft state.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -363,3 +384,135 @@
...
+	if !ok || auth == "" {
+		logger.Error("Missing Authorization header in tunnel request", logger.Ctx{"reqID": req.ID})
+		return types.ClusterManagerTunnelResponse{ID: req.ID, Status: http.StatusUnauthorized}
+	}
+	newReq.Header.Set("Authorization", auth)
+
+	lxdHttpClient := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true, // todo: should verify certificate
```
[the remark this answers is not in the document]
edlerd: Resolved by fetching cert through the unix socket and verifying it here.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -363,3 +384,135 @@
...
+			},
+		},
+	}
+
+	lxdResponse, err := lxdHttpClient.Do(newReq)
```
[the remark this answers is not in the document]
edlerd: This is by design, not sure if there is a solution for this.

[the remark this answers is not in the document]
edlerd: Added an allowlist of methods and a prefix check for the path.

api/types/cluster_manager.go
```diff
@@ -55,6 +58,10 @@ type ClusterManagerPut struct {
 	// Interval in seconds to send status messages to the cluster manager
 	// Example: 60
 	UpdateInterval *string `json:"update_interval" yaml:"update_interval"`
+
+	// Enables or disables the reverse tunnel to the cluster manager
+	// Example: true, false
+	ReverseTunnel *string `json:"reverse_tunnel" yaml:"reverse_tunnel"`
```
roosterfish: Also left a comment on the spec, but I think we should make the struct keys and actual config keys used by the CLI (e.g. `ReverseTunnelKey`) similar.

api/types/cluster_manager.go
```diff
@@ -100,3 +107,27 @@ type ClusterManagerJoin struct {
...
+
+// ClusterManagerTunnel represents the tunnel connection the cluster manager.
+type ClusterManagerTunnel struct {
+	Mu     sync.RWMutex
+	WsConn *websocket.Conn
+}
+
+// ClusterManagerTunnelRequest represents the request received through the tunnel.
+type ClusterManagerTunnelRequest struct {
+	ID      string            `json:"id"`
```
roosterfish: In the spec it's `UUID` if I am not mistaken.

roosterfish: Also for `ClusterManagerTunnelResponse`.

api/types/cluster_manager.go
```diff
@@ -100,3 +107,27 @@ type ClusterManagerJoin struct {
...
+	Method  string            `json:"method"`
+	Path    string            `json:"path"`
+	Headers map[string]string `json:"headers"`
```
roosterfish: You may want to use https://pkg.go.dev/net/http#Header.

api/types/cluster_manager.go
roosterfish: Also it should be of type `bool`.

api/types/cluster_manager.go
```diff
@@ -100,3 +107,27 @@ type ClusterManagerJoin struct {
...
+	Body    []byte            `json:"body"`
+}
+
+// ClusterManagerTunnelResponse represents the response sent through the tunnel.
+type ClusterManagerTunnelResponse struct {
+	ID      string            `json:"id"`
+	Status  int               `json:"status"`
+	Headers map[string]string `json:"headers"`
+	Cookies map[string]string `json:"cookies"`
```
roosterfish: You may want to use https://pkg.go.dev/net/http#Cookie.

api/types/cluster_manager.go
roosterfish: Using this type and the one for cookies might save you a few lines later when passing the information on.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -21,25 +28,32 @@ import (
...
+		}
+
 		ticker := time.NewTicker(database.UpdateIntervalDefaultSeconds * time.Second)
 		defer ticker.Stop()
 
 		for {
 			select {
 			case <-ticker.C:
-				newUpdateTime := sendClusterManagerStatusMessage(ctx, sh, s)
+				newUpdateTime := sendClusterManagerStatusMessage(ctx, sh, s, tunnel)
```
roosterfish: The more I was looking at it, the more I think we shouldn't pass the `tunnel` into `sendClusterManagerStatusMessage` and leave it as its own action that regular pings the cluster manager with the status message.

Instead why not having another routine that is dedicated for the tunneling to separate it from the status ping?

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -363,3 +384,186 @@ func getLocalPools(lxdClient lxd.InstanceServer) ([]string, error) {
 
 	return localPools, nil
 }
+
+func ensureTunnel(ctx context.Context, sh *service.Handler, tunnel *types.ClusterManagerTunnel, needsTunnel bool, clusterManagerClient *client.ClusterManagerClient, clusterCert *shared.CertInfo) {
```
roosterfish: Then in this second routine, at the beginning you check if a tunnel should be established based on the config key. And if yes, ensure a tunnel is setup and continue on.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -363,3 +384,186 @@ func getLocalPools(lxdClient lxd.InstanceServer) ([]string, error) {
...
+	tunnel.Mu.Unlock()
+
+	if needsTunnel && hasConnection {
+		logger.Debug("Websocket already connected, skipping reconnection")
+		return
+	}
+
+	if needsTunnel && !hasConnection {
+		logger.Debug("Websocket not connected, establishing connection in new goroutine")
+		go openTunnel(ctx, sh, tunnel, clusterManagerClient, clusterCert)
```
roosterfish: I have the feeling we are spawning too many routines here. One to open and one for close. I was thinking more about something like the following:

```go
// Spawned when tunneling is enabled
go func() {
    for {
        if doTunnel {
            // ensure we have a tunnel open with cluster manager
        } else {
            // exit outer routine
        }

        // read from conn
        err = conn.ReadJSON(&req)
        // handle req and send response
        // or exit gracefully if conn gets closed in the meantime.
        // If the outer ctx gets cancelled, this should also close the ws connection here as it should be using the outer ctx.
    }
}
```

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -363,3 +384,186 @@ func getLocalPools(lxdClient lxd.InstanceServer) ([]string, error) {
...
+		logger.Warn("Received tunnel request with invalid path prefix", logger.Ctx{"path": req.Path})
+		return types.ClusterManagerTunnelResponse{ID: req.ID, Status: http.StatusBadRequest}
+	}
+
+	if req.Method != http.MethodGet && req.Method != http.MethodPost && req.Method != http.MethodPut && req.Method != http.MethodDelete {
+		logger.Warn("Received tunnel request with unsupported HTTP method", logger.Ctx{"method": req.Method})
+		return types.ClusterManagerTunnelResponse{ID: req.ID, Status: http.StatusMethodNotAllowed}
+	}
+
+	targetURL := "https://127.0.0.1:8443" + req.Path
```
roosterfish: I mentioned this also on the spec. Let's retrieve the `core.https_address` from LXD first and validate if its even reachable via `localhost`. 
This is currently the case for all MicroClouds but I think we can make it more bulletproof going forward.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -363,3 +384,186 @@ func getLocalPools(lxdClient lxd.InstanceServer) ([]string, error) {
...
+
+	cookies, hasCookies := req.Headers["Cookie"]
+	if hasCookies {
+		newReq.Header.Set("Cookie", cookies)
+	}
+
+	newReq.Header.Set("Authorization", auth)
+	newReq.Header.Set("User-Agent", "LXD MicroCloud (Linux; x86_64) (cookiejar)")
+
+	tlsConfig := shared.InitTLSConfig()
```
roosterfish: This TLS config setup looks similar to the one returned from `getTlsConfig`. Can we reuse it?

cmd/microcloudd/cluster_manager_task.go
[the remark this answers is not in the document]
roosterfish: I think it boils down to the `req.Path` coming from the cluster manager. As only trusted parties can even send a `req` (MicroCloud opened up a ws connection to this party beforehand), this should be acceptable.

roosterfish (changes requested):
Hi, thanks for this. I have left a few comments on the spec and here, mainly around how we spawn and manage the tunnel routine as well as some inconsistencies I have spotted between setting various cluster manager configs on the CLI compared to the underlying struct fields.

roosterfish:
Also can you please extend the cluster manager test suite to test the tunnel?

api/types/cluster_manager.go
edlerd: yes, good idea that is much simpler indeed!

cmd/microcloudd/cluster_manager_task.go
edlerd: Yes, that sounds like a good plan to untangle the two features! I'll adjust it accordingly.

edlerd: The tunnel management routine then can check every 10 seconds on the actual and desired tunnel state and act accordingly.

cmd/microcloudd/cluster_manager_task.go
edlerd: I think we do need a dedicated routine for opening the tunnel. Because the open function loops endlessly. And we need to periodically check if the tunnel needs to be closed due to config or db-leader change.

cmd/microcloudd/cluster_manager_task.go
edlerd: Yes, good idea :+1:

cmd/microcloudd/cluster_manager_task.go
edlerd: Do you refer to `getTlsConfig` in `cluster_manager_client`? That one handles the secure connection to cluster manager. We do have a fingerprint and validate the fetched cert against the fingerprint. Then we append our client certificate for mtls.

Here, on the other hand, we connect to LXD. We fetch the cert from LXD's `1.0/` endpoint via a Unix socket. Then we can open a secure the connection via HTTP using that server certificate. Auth is done with the `Authorization` header coming in through the tunnel (or a session cookie coming in).

So I think both are fairly different.

api/types/cluster_manager.go
edlerd: > ...make the struct keys and actual config keys used by the CLI (e.g. `ReverseTunnelKey`) similar.

Not 100% I understand you correctly here. Is this about underscore and minus? Minus is easier to type, so the right choice for the CLI. On the api I'd prefer to use underscore.

> ... should be of type `bool`.

That requires a bit of a dance with converting back and forth because in the db storage this is saved as a string.

edlerd:
> Also can you please extend the cluster manager test suite to test the tunnel?

Added some basic tests for setting / unsetting. To test the whole tunnel feature set, we'd need a full cluster manager in CI.

cmd/microcloudd/cluster_manager_task.go
roosterfish: Ok I see. Then likely one routine that does the `read()` and one that blocks on the `ctx.Done()` and closes the ws.

Let's also use an `errgroup` for this so we can wait for the routines to exit before starting all over or stop the daemon. 
In addition this helps catching errors (for example if cluster manager closes the ws) not caused on the MicroCloud side.

api/types/cluster_manager.go
edlerd: Adopted to use bool and unifying the cli key / api key names.

cmd/microcloudd/cluster_manager_task.go
edlerd: Adjusted to use an err group for the main routine and a wait group for the nested routine

cmd/microcloudd/cluster_manager_task.go
edlerd: Agreed to not introduce a helper, as the overlap is just two lines.

edlerd:
@roosterfish please have another look at this, all comments above should be resolved.

client/cluster_manager_client.go
```diff
@@ -84,6 +86,26 @@ func (c *ClusterManagerClient) Delete(clusterCert *shared.CertInfo) error {
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get TLS config: %w", err)
+	}
+
+	dialer := websocket.Dialer{
+		TLSClientConfig: tlsConfig,
+	}
+
+	u := url.URL{Scheme: "wss", Host: address, Path: "/1.0/remote-cluster/ws"}
+	conn, _, err := dialer.Dial(u.String(), nil)
```
roosterfish: Should we use `.DialContext` and pass the outer context into `ConnectTunnel`? It guards the time of the request until the dialer returns the weboscket conn.

cmd/microcloud/cluster_manager.go
```diff
@@ -291,10 +298,16 @@ func (c *cmdClusterManagerSet) run(_ *cobra.Command, args []string) error {
...
 		payload.Addresses = []string{value}
-	case "certificate-fingerprint":
+	case "certificate_fingerprint":
 		payload.CertificateFingerprint = &value
-	case "update-interval-seconds":
-		payload.UpdateInterval = &value
+	case "update_interval_seconds":
+		payload.UpdateIntervalSeconds = &value
+	case "reverse_tunnel":
+		if value != "true" && value != "false" {
```
roosterfish: Maybe LXD's `validate.IsBool`? Just a nit.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -20,7 +21,8 @@ import (
 
 // SendClusterManagerStatusMessageTask starts a go routine, that sends periodic status messages to cluster manager.
 func SendClusterManagerStatusMessageTask(ctx context.Context, sh *service.Handler, s microTypes.State) {
-	go func(ctx context.Context, sh *service.Handler, s microTypes.State) {
+	g, _ := errgroup.WithContext(ctx)
```
roosterfish: It should be `g, ctx :=` so that the `case <-ctx.Done():` below uses the derived context from the errgroup, not the outer one.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -33,15 +35,22 @@ func SendClusterManagerStatusMessageTask(ctx context.Context, sh *service.Handle
...
 
 			case <-ctx.Done():
-				return // exit the loop and close the go routine
+				return nil // exit the loop and close the go routine
 			}
 		}
-	}(ctx, sh, s)
+	})
+
+	go func() {
```
roosterfish: Running the `.Wait` inside a routine doesn't guarantee we are actually waiting for all the routines in the errgroup `g` to finish as this routine might itself be killed when the daemon is shutting down. 

I was about to suggest we need to hook into something like a `OnStop` hook and put the `Wait` there but just realized Microcluster doesn't even have such functionality yet.
The daemon's shutdownCtx gets cancelled but there is no way for the daemon to wait until the routines spawned by the user are done or not.

Looks like something that we should add to Microcluster so we can properly use it here.

test/suites/cluster-manager.sh
```diff
@@ -15,19 +15,60 @@ test_cluster_manager() {
...
+        accept = base64.b64encode(hashlib.sha1((key + WS_MAGIC).encode()).digest()).decode()
+        response = (
+            'HTTP/1.1 101 Switching Protocols\r\n'
+            'Upgrade: websocket\r\n'
+            'Connection: Upgrade\r\n'
+            f'Sec-WebSocket-Accept: {accept}\r\n\r\n'
+        )
+        conn.sendall(response.encode())
+        open(HIT_FILE, 'w').write('hit')
+        import time; time.sleep(30)
```
roosterfish: Why the sleep here? If the conn is upgraded to websocket, can't the server simply continue looping on `accept()`?

roosterfish (changes requested):
Thanks @edlerd, I like the approach with the small dummy webserver to the the ws upgrade. 

Whilst going through the routines logic, I spotted that Microcluster currently doesn't seem to have any option to wait for the user spawned processes. So instead of running the `Wait` for the errgroup also in a routine, I have the feeling there should be another hook which users like MicroCloud can instrument to block the Microcluster daemon from shutting down too early so we can perform a wait on our spawned routines. I can take a look at this. 

Also can you please split commits so that it's easier to go through and the changes are logically grouped together? Especially now since the config keys also got changed to use underscore.

test/suites/cluster-manager.sh
```diff
@@ -15,19 +15,60 @@ test_cluster_manager() {
...
+    else:
+        conn.sendall(b'HTTP/1.1 200 OK\r\nContent-Length: 0\r\n\r\n')
+        import time; time.sleep(1)
+    conn.close()
+PYEOF"
+  lxc exec micro01 -- sh -c "nohup sh -c 'nohup python3 ws_server.py > ws_server.log 2>&1'" &
```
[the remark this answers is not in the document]
edlerd: inner call is blocking and that is fine.

test/suites/cluster-manager.sh
edlerd: The sleep keeps the WebSocket connection open for the test. Without it, the server would send 101 and immediately close the connection, making it hard to verify a successful upgrade. In a real implementation, the connection would stay open and process WebSocket traffic instead.

database/cluster_manager_config.go
```diff
@@ -14,31 +16,106 @@ import (
...
+
+	return &clusterManagerConfig[0], nil
+}
+
+// LoadClusterManagerUpdateIntervalSeconds loads the cluster manager update interval configuration from the database.
+func LoadClusterManagerUpdateIntervalSeconds(state types.State, ctx context.Context, clusterManagerId int64) (time.Duration, error) {
+	defaultInterval := time.Duration(UpdateIntervalDefaultSeconds) * time.Second
+	updateIntervalConfig, err := LoadClusterManagerSingleConfig(state, ctx, clusterManagerId, UpdateIntervalSecondsKey)
+	if err != nil {
+		return defaultInterval, err
```
roosterfish: In the error case we should we instead return `nil` duration? Or is this the "default" case when nothing is explicitly configured in the DB? 

I think a better pattern would be to return a 404 status error from `LoadClusterManagerSingleConfig` and then check `api.StatusErrorCheck(err, http.StatusNotFound)` so we can safely ignore the error, return a `nil` err to the caller together with the default duration.

cmd/microcloudd/cluster_manager_tunnel.go
```diff
@@ -0,0 +1,303 @@
...
+		return
+	}
+
+	leaderClient, err := s.Database().Leader(ctx)
+	if err != nil {
+		logger.Error("Failed to get database leader client", logger.Ctx{"err": err})
+		return
+	}
+
+	leaderInfo, err := leaderClient.Leader(ctx)
```
roosterfish, line 86: At the moment when the leadership moves to another member, there can potentially be multiple websockets be open at the same time until the next iteration of the loop.

Can this be a problem on cluster manager side? I mean how do you know which tunnel is the one established with the current leader? Maybe this isn't even an issue, but still wanted to raise it.

cmd/microcloudd/cluster_manager_tunnel.go
```diff
@@ -0,0 +1,303 @@
...
+	if err != nil {
+		logger.Error("Failed to get database leader info", logger.Ctx{"err": err})
+		return
+	}
+
+	if leaderInfo.Address != s.Address().Host {
+		ensureTunnelClosed(tunnel)
```
roosterfish, line 93: In the outer loop you are waiting on the inner waitgroup of the tunnel after calling `ensureTunnelClosed`. Wouldn't this be also necessary here? There could be a central helper func for this purpose or the wait can be performed within `ensureTunnelClosed`.

cmd/microcloudd/cluster_manager_tunnel.go
```diff
@@ -0,0 +1,303 @@
...
+		return
+	}
+
+	clusterManager, err := database.LoadClusterManager(s, ctx, database.ClusterManagerDefaultName)
+	if err != nil {
+		if err.Error() == "Cluster manager not found" {
```
roosterfish: Can we please use 404 status errors here instead of string comp? This aligns best with the code base.

cmd/microcloudd/cluster_manager_tunnel.go
```diff
@@ -0,0 +1,303 @@
...
+	conn, err := clusterManagerClient.ConnectTunnel(ctx, clusterCert)
+	if err != nil {
+		logger.Error("Failed to connect cluster manager tunnel", logger.Ctx{"err": err})
+		return
+	}
+
+	tunnel.Mu.Lock()
+	if tunnel.WsConn != nil {
+		tunnel.Mu.Unlock()
+		_ = conn.Close()
```
roosterfish: Why not checking first if `tunnel.WsConn != nil` and only perform `ConnectTunnel` if `== nil`?

cmd/microcloudd/cluster_manager_tunnel.go
```diff
@@ -0,0 +1,303 @@
...
+	for {
+		var req types.ClusterManagerTunnelRequest
+		err = conn.ReadJSON(&req)
+		if err != nil {
+			logger.Error("Cluster manager tunnel read error:", logger.Ctx{"err": err})
+			ensureTunnelClosed(tunnel)
+			return
+		}
+
+		logger.Debug("Cluster manager tunnel request received:", logger.Ctx{"path": req.Path})
```
roosterfish: Just a nit, worth logging the payload too when in debugging mode?

cmd/microcloudd/cluster_manager_tunnel.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+}
+
+func handleTunnelRequest(req types.ClusterManagerTunnelRequest, lxdServerCert string, lxdHttpsAddress string) types.ClusterManagerTunnelResponse {
+	if req.Path != "/1.0" && !strings.HasPrefix(req.Path, "/1.0/") {
+		logger.Warn("Received tunnel request with invalid path prefix", logger.Ctx{"path": req.Path})
+		return types.ClusterManagerTunnelResponse{UUID: req.UUID, Status: http.StatusBadRequest}
+	}
+
+	if req.Method != http.MethodGet && req.Method != http.MethodPost && req.Method != http.MethodPut && req.Method != http.MethodPatch && req.Method != http.MethodDelete {
```
roosterfish: Can we please do a `slices.Contains` check instead for readability purposes? Or a switch?

roosterfish: In general can't we just forward the request to LXD? If the respective API endpoint isn't implemented for the given method, it errors anyway.

roosterfish: Same with the path check above.

cmd/microcloudd/cluster_manager_tunnel.go
```diff
@@ -0,0 +1,303 @@
...
+		logger.Warn("Received tunnel request with unsupported HTTP method", logger.Ctx{"method": req.Method})
+		return types.ClusterManagerTunnelResponse{UUID: req.UUID, Status: http.StatusMethodNotAllowed}
+	}
+
+	if lxdHttpsAddress == "[::]:8443" || lxdHttpsAddress == ":8443" {
+		lxdHttpsAddress = "127.0.0.1:8443"
```
roosterfish: Mh that check seems very narrow. Instead of appending the 127 address, can we instead do this mangling on the calling site after retrieving the `core.https_address`? There you also have access to the `sh` handler struct which might allow you to get the address used for MicroCloud/Microcluster and LXD will use the same.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -20,7 +21,8 @@ import (
 
 // SendClusterManagerStatusMessageTask starts a go routine, that sends periodic status messages to cluster manager.
 func SendClusterManagerStatusMessageTask(ctx context.Context, sh *service.Handler, s microTypes.State) {
-	go func(ctx context.Context, sh *service.Handler, s microTypes.State) {
+	g, ctx := errgroup.WithContext(ctx)
```
roosterfish: Can you please merge this with the original commit that adds `g, _ :=`

cmd/microcloudd/main.go
```diff
@@ -184,6 +184,7 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 				return setHandlerAddress(state.Address().Host)
 			},
 			OnStart: func(ctx context.Context, state microTypes.State) error {
+				ReconcileClusterManagerTunnel(ctx, s, state)
```
roosterfish: Likely once we have a `OnStop` we either pass the errgroup into both funcs here so we can then also have it available in `OnStop` and properly call `Wait`.

roosterfish (changes requested):
Thanks, a few more smaller comments from my side after properly looking through.

cmd/microcloudd/cluster_manager_tunnel.go
edlerd: Good point!

Cluster manager is storing which cluster manager member (it is a cluster itself) holds the tunnel to a remote microcloud in its database. We clear that member value in the database on disconnect of the tunnel. We need to be careful not to clear a value of a new tunnel that landed on another member in case of tunnel reconnection, as the timing can smear. I revised the logic and it should be sound now within cluster manager.

roosterfish:
I have just added https://github.com/canonical/microcluster/pull/766 to address the `OnStop`.

database/cluster_manager_config.go
edlerd: changing this to return nil if it is not found and handling that case in the caller.

cmd/microcloudd/cluster_manager_tunnel.go
edlerd: yes! moving the wait into the ensure closed is best!

cmd/microcloudd/cluster_manager_tunnel.go
edlerd: This causes a bit longer locking time. We keep the lock while connection. But as there is only a single connection, it is okay and even desired to check earlier and lock longer. :+1:

cmd/microcloudd/cluster_manager_tunnel.go
edlerd: Both are safeguards to not allow arbitrary paths / methods. Limiting the attack surface.

I'll change this to use slices though.

edlerd:
@roosterfish thanks for the review! I resolved the open issues, please have another pass at the proposed changes.

cmd/microcloudd/main.go
roosterfish: Just landed in Microcluster v3 https://github.com/canonical/microcluster/pull/767.

edlerd: Hah, that was fast :rocket: 

Shall we update microcluster in this PR and start using it right away? Or do you want the update in a dedicated PR and usagein a follow up?

roosterfish: If you can use it starting with this PR that would be great.

edlerd:
@roosterfish I updated microcluster v3 and adjusted to use the on stop hook for the background tasks. please give this changeset a pass at review.

cmd/microcloudd/main.go
```diff
@@ -247,6 +256,22 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
...
+					backgroundTasksCancel()
+				}
+
+				if backgroundTasks == nil {
+					return nil
+				}
+
+				err := backgroundTasks.Wait()
+				if err != nil {
+					return fmt.Errorf("Cannot stop cluster manager background tasks: %w", err)
```
roosterfish: To be in line with gerund style errors `Failed stopping ...`.

cmd/microcloudd/main.go
```diff
@@ -184,7 +188,12 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 				return setHandlerAddress(state.Address().Host)
 			},
 			OnStart: func(ctx context.Context, state microTypes.State) error {
-				SendClusterManagerStatusMessageTask(ctx, s, state)
+				backgroundTasksCtx, cancel := context.WithCancel(ctx)
```
roosterfish: I see the pain with the context.
But let's try the following:

1) Don't define an extra context and cancel. When the `OnStop` hook fires, the daemon's shutdown context already got cancelled. This means it cancelled also the context passed into `OnStart` which will cause context cancellation of the one passed into the errgroup. So we don't have to manually cancel the context.
2) In `OnStop` return early if `backgroundTasks == nil` or wait as you are already doing.

This should allow doing the same with less lines.

roosterfish (changes requested):
Thanks for applying those changes so quickly. Just one more comment as I think we don't have to manually cancel the context inside the `OnStop` as this is already done by the daemon.

roosterfish:
Did you test stopping the MicroCloud daemon to see it doesn't block for some reason we wouldn't expect? I am pretty sure we have some `snap stop microcloud` in the test suite but just to double check when the cluster manager routines are spawned.

test/suites/cluster-manager.sh
roosterfish: Ah yes I see thanks.

roosterfish: Please add a comment next time you are at it.

edlerd:
Did simplifications as suggested above.

> Did you test stopping the MicroCloud daemon to see it doesn't block for some reason we wouldn't expect?

Yes, a simple snap stop worked well. With and without cluster manager configured or with/without tunnel.

test/suites/cluster-manager.sh
edlerd: added an inline comment

cmd/microcloudd/cluster_manager_tunnel.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	lxdHttpClient := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsConfig,
+		},
+		Timeout: 30 * time.Second,
+	}
+
+	lxdResponse, err := lxdHttpClient.Do(newReq)
```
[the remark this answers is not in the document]
roosterfish: The attach vector of this "user provided input" is limited because it's 

1) created by cluster manager
2) send through the tunnel which is proactively opened by MicroCloud
3) target is fixed as we craft the `lxdHttpsAddress` server address without user input

I am confident we can dismiss this alert.

But would it make sense to put a whitelist of allowed endpoints? AFAIK you can trigger anyway only a limited set of functionality from cluster manager.

cmd/microcloudd/cluster_manager_tunnel.go
```diff
@@ -0,0 +1,321 @@
...
+	tunnel.Mu.Unlock()
+
+	if needsTunnel && hasConnection {
+		logger.Debug("Tunnel already connected")
+		return
+	}
+
+	if needsTunnel && !hasConnection {
+		logger.Debug("Tunnel not connected, opening")
+		tunnel.Wg.Add(1)
```
roosterfish: Figured the other day you can now do `wg.Go` which spawns a routine and does the `Add(1)`. Just a nit. Looks this was added in one of the recent Go versions.

cmd/microcloudd/cluster_manager_tunnel.go
```diff
@@ -0,0 +1,321 @@
...
+	if err != nil {
+		logger.Error("Failed to get LXD server info", logger.Ctx{"err": err})
+		return
+	}
+
+	lxdServerCert := server.Environment.Certificate
+	lxdHttpsAddress := fmt.Sprint(server.Config["core.https_address"])
+	if lxdHttpsAddress == "[::]:8443" || lxdHttpsAddress == ":8443" {
+		cloud := sh.Services[types.MicroCloud].(*service.CloudService)
+		lxdHttpsAddress = cloud.Address() + ":8443"
```
roosterfish: Nice, glad this worked out this way.

roosterfish: For a follow up, there is a `service.LXDPort` var too :)

roosterfish (dismissed):
LGTM! Just a few comments for follow ups and about the flagged security issue where we might want to have a whitelist of LXD endpoints to be safe?

edlerd:
> few comments for follow ups 

Addressed those just now.

> about the flagged security issue where we might want to have a whitelist of LXD endpoints to be safe?

If I recall correctly, we discussed this previously, and @tomponline raised a valid concern that we would then have to maintain that list of allowed endpoints, which could become cumbersome or even blocking in the future. Hence my current approach to allow anything with `/1.0` prefix and a concrete list of methods.

cmd/microcloudd/cluster_manager_tunnel.go
roosterfish: Also you don't have to `defer wg.Done()`. This happens automatically too once the func exits.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
```
roosterfish, line 4: Just saw this commit seems to have two commits squashed into one? See the commit message.

roosterfish (changes requested):

cmd/microcloudd/cluster_manager_task.go
edlerd: I though that is what you requested with one of the comments. But I might have misunderstood the intention back then.

Entangled those two into dedicated commits.

api/cluster_manager.go
```diff
@@ -191,13 +197,21 @@ func clusterManagerPut(state types.State, r *http.Request) types.Response {
...
 		if err != nil {
 			return types.SmartError(err)
 		}
 	}
 
+	if args.ReverseTunnel != nil {
+		reverseTunnelValue := strconv.FormatBool(*args.ReverseTunnel)
+		err = database.StoreClusterManagerConfig(state, r.Context(), name, database.ReverseTunnelKey, reverseTunnelValue)
+		if err != nil {
+			return response.SmartError(err)
```
roosterfish: `types.SmartError`. Looks like this one creep-ed in. Saw the import at the top.

cmd/microcloudd/cluster_manager_tunnel.go
```diff
@@ -0,0 +1,321 @@
+package main
```
roosterfish, line 1: Would you mind moving this commit right before the one that starts using `ReconcileClusterManagerTunnel`?

roosterfish (changes requested):

roosterfish (approved):
Thanks!

PR #821, opened 2025-06-10
ensure app is initialized for any cluster manager command

### LINKED ISSUE #811: Cluster manager subcommand should check if MicroCloud is initialized
All the other `microcloud *` subcommands check whether or not MicroCloud is initialized. 
This ensures the user gets presented with a clear error message in case MicroCloud isn't yet ready. 

When running the `microcloud cluster-manager *` commands, currently there is no extra check which causes the user to see the underyling `Error: Database is not yet initialized` in case MicroCloud isn't yet initialized.

To fix it add a check equal to https://github.com/canonical/microcloud/blob/main/cmd/microcloud/status.go#L127-L129.

edlerd:
# Done
* ensure app is initialized for any cluster manager command

fixes #811 and WD-22691

### DIFF
--- a/cmd/microcloud/cluster_manager.go
+++ b/cmd/microcloud/cluster_manager.go
@@ -26,6 +26,14 @@ func (c *cmdClusterManager) Command() *cobra.Command {
 	cmd := &cobra.Command{}
 	cmd.Use = "cluster-manager"
 	cmd.Short = "Manage cluster manager connection"
+	cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+		// Skip the check if this is the root command itself
+		if cmd.Use == "cluster-manager" {
+			return nil
+		}
+
+		return checkInitialized(c.common.FlagMicroCloudDir, true, false)
+	}
 
 	// Join
 	clusterManagerJoinCmd := cmdClusterManagerJoin{common: c.common, alias: c}


cmd/microcloud/cluster_manager.go
```diff
@@ -75,6 +75,11 @@ func (c *cmdClusterManagerJoin) command() *cobra.Command {
 }
 
 func (c *cmdClusterManagerJoin) run(_ *cobra.Command, args []string) error {
+	err := ensureAppInitialized(c.common)
```
roosterfish: Are you aware of `checkInitialized` which is used for the other commands? https://github.com/canonical/microcloud/blame/main/cmd/microcloud/ask.go#L32

roosterfish: If you call it as `checkInitialized(path, false, false)` it should provide what you are looking for and also print a nice message what do to next.

It looks the manual call to the ready endpoint is only done in cases where the action itself wouldn't fit to the reply of `checkInitialized`. But in this case you are actually looking for an initialized MicroCloud.

roosterfish (changes requested):

cmd/microcloud/cluster_manager.go
edlerd: Thanks for the hint, this simplifies the PR. See changes below.

edlerd: I think the second argument should be `true`, because we expect the cloud to be initialized.

roosterfish: This made me thinking if `cobra` allows running some preliminary code for all subcommands. It looks like there is `cmd.PersistentPreRun` func which can potentially be used to run the `checkInitialized` for all cluster manager subcommands.

edlerd: Ah very nice, replacing this to avoid the duplication :+1:

cmd/microcloud/cluster_manager.go
```diff
@@ -26,6 +26,9 @@ func (c *cmdClusterManager) Command() *cobra.Command {
 	cmd := &cobra.Command{}
 	cmd.Use = "cluster-manager"
 	cmd.Short = "Manage cluster manager connection"
+	cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
```
roosterfish, line 29: Mh, I was just trying this because I wondered if it will overwrite the default behavior of displaying the help text. And it seems it does :(

When you just run `microcloud cluster-manager` you would expect to see the `-h` output, but instead you get the error saying MicroCloud is not initialized.

edlerd: Good catch! safeguarding the check now against the root command.

PR #823, opened 2025-06-15
doc: add information about using a single network interface

minaelee:
Replaces https://github.com/canonical/microcloud/pull/596 to clarify that a configuration with a single network interface per cluster member can be used in testing/development environments.

### DIFF
--- a/doc/reference/requirements.md
+++ b/doc/reference/requirements.md
@@ -77,7 +77,7 @@ For further information, see the Prerequisites section of this page: {doc}`micro
 (reference-requirements-network-interfaces)=
 ### Required network interfaces
 
-For networking, each machine in the MicroCloud cluster requires at least two dedicated network interfaces: one for intra-cluster communication and one for external connectivity. These can be connected to the same network, or different networks.
+For networking, use at least two dedicated network interfaces for each machine in a MicroCloud cluster: one for intra-cluster communication and one for external connectivity. These can be connected to the same network, or different networks.
 
 In production environments, we recommend dual-port network cards with a minimum 10 GiB capacity, or higher if low latency is essential.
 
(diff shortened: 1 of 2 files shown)


doc/reference/requirements.md
```diff
@@ -98,6 +98,13 @@ Network interface to connect to the uplink network
    You can specify a different interface to be used as the uplink interface for each cluster member.
    MicroCloud requires that all uplink interfaces are connected to the uplink network, using the gateway and IP address range information that you provide during the MicroCloud initialization process.
 
+(reference-requirements-network-interface-single)=
+### Single network interface configuration
+
+For testing or development environments, it is possible to use a single network interface per cluster member, with a bridge in front of each for external connectivity to the uplink network. An OVS bridge is preferable to a native bridge to avoid an extra hop for OVN uplink traffic.
+
```
mseralessandri, line 105: Hi @minaelee thank you for this update! Should we explain (not here but maybe in a page under the "How-to" guides) how to use this configuration with MicroCloud, still clarifying that it is not recommended for production environment? I've seen several times questions around this topic.

minaelee: The guidance I've received has been to avoid going into depth about this topic in the official documentation since we do not support it, and it seems to me that a how-to guide would fall under "going into depth".

roosterfish: Maybe we can add a link to the respective netplan configuration without going too much into detail? https://netplan.readthedocs.io/en/stable/examples/#how-to-configure-network-bridges.
This way we give a hint but its still up to the user to apply the necessary changes (identify the required config for their netplan config file) and how to use a OVS bridge instead of a native one.

minaelee: That sounds like a great compromise, thanks Julian. I've updated with the following:

>However, this configuration is neither supported nor recommended, because it prevents traffic segregation of intra-cluster and uplink traffic. If you wish to experiment with this setup, you can find basic examples of bridge configuration using Netplan in the [Netplan documentation](https://netplan.readthedocs.io/en/stable/examples/#how-to-configure-network-bridges). Note that this is a general reference and does not cover the use of OVS or MicroCloud. You must determine the appropriate configuration for your environment.

Issue #649, opened 2025-02-19
Make the LXD refresh steps more clear


roosterfish:
Currently the docs suggest to evacuate cluster members when performing an update/upgrade.
However this is not necessary when refreshing LXD.

Also document that LXD will potentially perform an auto refresh on the other cluster members after some time if the refresh command isn't performed on all cluster members.

ismailkayi:
There are warnings for MicroCeph and MicroOVN to be done one by one. We should make "The refresh will block until each of the LXD cluster members is updated, so make sure to perform the following command on all machines in parallel" message more significant. Also, as @roosterfish explained, the refresh will not block the update. 

<img width="848" alt="Image" src="https://github.com/user-attachments/assets/5f75160c-ba3f-440b-acda-8c73284a2507" />

PR #824, opened 2025-06-16
doc: clarify lxd refresh process

minaelee:
This PR:
- Clarifies that evacuating running instances is not necessary for LXD snap refreshes.
- Emphasizes and further explains why LXD snap updates should be run in parallel, rather than in sequence like the other snaps for MicroCloud.

Closes https://github.com/canonical/microcloud/issues/649


roosterfish (approved):
LGTM

Issue #781, opened 2025-05-09
preseed storage filtering


wideawakening:
i'm not able to filter the two available disks: nvme1n1, nvme3n1
microcloud version: 2/stable, 2.1.0-3e8b183, rev 1144

```
ubuntu@pc7b-ra8-n2:~$ lsblk -d -o NAME,ROTA,TYPE,MODEL
NAME    ROTA TYPE MODEL
... loops
nvme2n1    0 disk INTEL SSDPF2KE064T1O
nvme1n1    0 disk INTEL SSDPF2KE064T1O
nvme0n1    0 disk INTEL SSDPF2KE064T1O
nvme3n1    0 disk INTEL SSDPF2KE064T1O

ubuntu@pc7b-ra8-n2:~$ lsblk
NAME                  MAJ:MIN RM   SIZE RO TYPE  MOUNTPOINTS
...
nvme2n1               259:0    0   5.8T  0 disk  
├─nvme2n1p1           259:1    0   488M  0 part  /boot/efi
└─nvme2n1p2           259:2    0   5.8T  0 part  
  └─md1                 9:1    0   5.8T  0 raid1 
    ├─vg0-lvephemeral 252:0    0   5.3T  0 lvm   
    ├─vg0-lvlog       252:1    0  93.1G  0 lvm   /var/log
    └─vg0-lvroot      252:2    0 465.7G  0 lvm   /
nvme1n1               259:3    0   5.8T  0 disk  
nvme0n1               259:4    0   5.8T  0 disk  
├─nvme0n1p1           259:6    0   488M  0 part  
└─nvme0n1p2           259:7    0   5.8T  0 part  
  └─md1                 9:1    0   5.8T  0 raid1 
    ├─vg0-lvephemeral 252:0    0   5.3T  0 lvm   
    ├─vg0-lvlog       252:1    0  93.1G  0 lvm   /var/log
    └─vg0-lvroot      252:2    0 465.7G  0 lvm   /
nvme3n1               259:5    0   5.8T  0 disk  
```

i tried several approaches

1 - partitioned/partitions attribute not usable

```
## seems partitioned attribute is not there anymore
## partitions does exist though https://github.com/canonical/lxd/blob/c86603236167a43836c2766647e2fac97d79f899/shared/api/resource.go#L591
## seems this attributed should not be used eitherway, since the partitioned disks are ignored https://github.com/canonical/microcloud/blob/main/service/system_information.go#L105
[...]

PR #825, opened 2025-06-16
doc: remove partitioned filter in preseed example

minaelee:
Per https://github.com/canonical/microcloud/issues/781, removes ceph filter example of `partitioned==false`. MicroCloud already ignores partitioned disks, so there's no need to filter them as this example suggests.


roosterfish (approved):
Thanks!

PR #829, opened 2025-06-23
doc: update snaps and releases documentation

minaelee:
This PR:
- Adds a new Releases and snaps reference page, similar to the recently added LXD documentation [Releases and snap](https://documentation.ubuntu.com/lxd/latest/reference/releases-snap/) page. Updates/adds links to this page from other pages.
- Enables use of a sustitutions.yaml file similar to LXD's for facilitating documentation updates of the current LTS and feature tracks for MicroCloud.
- Moves the information about holding and synchronizing updates from the Manage the snap page to the Updates and upgrades page.
- Improves and adds to the Manage the snap and Updates and upgrades how-to guides.
- Updates the How to get support guide to align with LXD's [how to get support guide](https://documentation.ubuntu.com/lxd/latest/support/).
- Updates the How to install MicroCloud guide to use the recommended defaults for installation, indicate that specifying a channel is optional, clarify how to handle a previously installed snap, etc.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -112,5 +112,8 @@ serve: html
 serve-microcloud: microcloud
 	cd $(current_dir)/_build/; python3 -m http.server --bind 127.0.0.1 8000
 
+install:
+	$(MAKE) -f Makefile.sp sp-install
+	
 %:
 	$(MAKE) -f Makefile.sp sp-$@
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -82,7 +82,7 @@ html: integrate
 	cd integration/lxd/doc/ && $(MAKE) html BUILDDIR=$(current_dir)/_build/lxd
 	cd integration/microceph/docs/ && $(MAKE) html BUILDDIR=$(current_dir)/_build/microceph
 	cd integration/microovn/docs/ && $(MAKE) html BUILDDIR=$(current_dir)/_build/microovn
-	$(MAKE) -f Makefile.sp sp-html BUILDDIR=$(current_dir)/_build/microcloud
+	$(MAKE) -f Makefile.sp sp-html BUILDDIR=$(current_dir)/_build/microcloud ADDPREREQS='pyyaml'
 
 # `html-rtd` builds the integrated docs, with the correct paths for Read the Docs.
 # This target is used by the Read the Docs build.
@@ -90,7 +90,7 @@ html-rtd:
 	PATH_PREFIX=$(PATH_PREFIX) $(MAKE) -C integration/lxd/doc/ html-rtd BUILDDIR=$(READTHEDOCS_OUTPUT)/html/lxd
 	PATH_PREFIX=$(PATH_PREFIX) $(MAKE) -C integration/microceph/docs/ html BUILDDIR=$(READTHEDOCS_OUTPUT)/html/microceph
 	PATH_PREFIX=$(PATH_PREFIX) $(MAKE) -C integration/microovn/docs/ html BUILDDIR=$(READTHEDOCS_OUTPUT)/html/microovn
-	PATH_PREFIX=$(PATH_PREFIX) $(MAKE) -f Makefile.sp sp-html BUILDDIR=$(READTHEDOCS_OUTPUT)/html/microcloud
+	ADDPREREQS='pyyaml' PATH_PREFIX=$(PATH_PREFIX) $(MAKE) -f Makefile.sp sp-html BUILDDIR=$(READTHEDOCS_OUTPUT)/html/microcloud
 
 # `spelling` checks only the MicroCloud docs.
 spelling: clean-doc microcloud
@@ -113,7 +113,7 @@ serve-microcloud: microcloud
 	cd $(current_dir)/_build/; python3 -m http.server --bind 127.0.0.1 8000
 
 install:
-	$(MAKE) -f Makefile.sp sp-install
+	$(MAKE) -f Makefile.sp sp-install ADDPREREQS='pyyaml'
 	
 %:
 	$(MAKE) -f Makefile.sp sp-$@
--- a/doc/custom_conf.py
+++ b/doc/custom_conf.py
@@ -1,5 +1,6 @@
 import datetime
 import os
+import yaml
 
 # Custom configuration for the Sphinx documentation builder.
 # All configuration specific to your project should be done in this file.
@@ -221,6 +222,7 @@ custom_extensions = [
 # sphinxext-opengraph
 custom_required_modules = [
     'sphinx-sitemap',
+    'pyyaml',
 ]
 
 # Add files or directories that should be excluded from processing.
@@ -294,3 +296,8 @@ if not ('SINGLE_BUILD' in os.environ and os.environ['SINGLE_BUILD'] == 'True'):
     custom_templates_path = ['integration/microcloud/_templates']
     redirects['../index'] = 'microcloud/'
     custom_tags.append('integrated')
+
+# Load substitutions from YAML file
+if os.path.exists('./substitutions.yaml'):
+    with open('./substitutions.yaml', 'r') as fd:
+        myst_substitutions = yaml.safe_load(fd.read())
--- a/doc/custom_conf.py
+++ b/doc/custom_conf.py
@@ -164,7 +164,9 @@ redirects = {
 # Links to ignore when checking links
 linkcheck_ignore = [
     'http://127.0.0.1:8000',
-    'http://localhost:8000'
+    'http://localhost:8000',
+    # These links may fail from time to time
+    'https://ceph.io',
     ]
 
 # Pages on which to ignore anchors
--- a/doc/how-to/install.md
+++ b/doc/how-to/install.md
@@ -86,27 +86,73 @@ For detailed information, see: {ref}`reference-requirements`.
 ```{youtube} https://www.youtube.com/watch?v=M0y0hQ16YuE
 ```
 
-To install MicroCloud, install all required {ref}`snaps <reference-requirements-software-snaps>` on all machines that you want to include in your cluster.
+To install MicroCloud, install all required {ref}`snaps <reference-requirements-software-snaps>` on all machines that you want to include in your cluster. You can {ref}`optionally specify a channel <howto-install-specify-channel>` for each snap, but generally, you can leave out the channel to use the current recommended default. 
 
 To do so, enter the following commands on all machines:
 
-    sudo snap install lxd --channel=5.21/stable --cohort="+"
-    sudo snap install microceph --channel=squid/stable --cohort="+"
-    sudo snap install microovn --channel=24.03/stable --cohort="+"
-    sudo snap install microcloud --channel=2/stable --cohort="+"
-
-```{note}
-Make sure to install the same version of the snaps on all machines.
-See {ref}`howto-snap` for more information.
-
-If you don't want to use MicroCloud's full functionality, you can install only some of the snaps.
-However, this is not recommended.
+```bash
+sudo snap install lxd --cohort="+"
+sudo snap install microceph --cohort="+"
+sudo snap install microovn --cohort="+"
+sudo snap install microcloud --cohort="+"
 ```
 
-```{note}
-It's possible that a required snap is already installed on your machine. For example, it might be a version of Ubuntu that comes with LXD pre-installed.
-In this case, run `sudo snap refresh <snap> --channel=<track>/stable --cohort="+"` to refresh (update) the installed snap.
+The `--cohort` flag ensures that versions remain {ref}`synchronized during later updates <howto-update-sync>`.
+
+Following installation, make sure to {ref}`hold updates <howto-update-hold>`.
+
+### Previously installed snaps
+
+If a required snap is already installed on your machine, you will receive a message to that effect. In this case, check the version for the installed snap:
+
+```bash
+snap list <snap>
 ```
 
-After installing the snaps make sure to hold any automatic updates to keep the used snap versions across MicroCloud in sync.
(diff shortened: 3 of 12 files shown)


doc/Makefile
```diff
@@ -81,15 +81,15 @@ html: integrate
 	cd integration/lxd/doc/ && $(MAKE) html BUILDDIR=$(current_dir)/_build/lxd
 	cd integration/microceph/docs/ && $(MAKE) html BUILDDIR=$(current_dir)/_build/microceph
 	cd integration/microovn/docs/ && $(MAKE) html BUILDDIR=$(current_dir)/_build/microovn
-	$(MAKE) -f Makefile.sp sp-html BUILDDIR=$(current_dir)/_build/microcloud
+	$(MAKE) -f Makefile.sp sp-html BUILDDIR=$(current_dir)/_build/microcloud ADDPREREQS='gitpython pyyaml'
```
roosterfish: I am trying to follow what this change is all about as it doesn't match the commit message.

Can you please move this change (and probably the other `ADDPREREQS`) into a separate commit?

doc/custom_conf.py
```diff
@@ -163,7 +164,9 @@
 # Links to ignore when checking links
 linkcheck_ignore = [
     'http://127.0.0.1:8000',
-    'http://localhost:8000'
+    'http://localhost:8000',
+    # These links may fail from time to time
+    'https://ceph.io',
```
roosterfish, line 169: Can this please also be included in its own commit?

doc/custom_conf.py
```diff
@@ -294,3 +298,8 @@
     custom_templates_path = ['integration/microcloud/_templates']
     redirects['../index'] = 'microcloud/'
     custom_tags.append('integrated')
+
+# Load substitutions from YAML file
+if os.path.exists('./substitutions.yaml'):
```
roosterfish, line 303: Are the substitutions just regular YAML key/value pairs you can set to arbitrary contents so you can reference them using `{{ }}` inside the docs? 
I understand as long as it's assigned to `myst_substitutions` you can then make use of it.

doc/Makefile
roosterfish: Ah now I see, you require the `pyyaml` to load the substitutions file. Can you maybe add a comment somewhere?

For what is `gitpython` used?

doc/reference/releases-snaps.md
```diff
@@ -0,0 +1,161 @@
...
+
+(ref-releases-matrix)=
+## Supported and compatible releases
+
+```{table}
+:align: center
+
+| Ubuntu LTS | MicroCloud | LXD        | MicroCeph       | MicroOVN |
+| ---------- | ---------- | ---------- | --------------- | -------- |
+| 22.04      | 2.1        | 5.0        | Quincy          | 22.03    |
```
roosterfish: LXD 5.0 is not supported by MicroCloud 2 LTS. On 22.04 you should also use LXD 5.21.

roosterfish: Also MicroCeph `squid` and MicroOVN `24.03` is enforced by MicroCloud 2 LTS.

doc/reference/releases-snaps.md
```diff
@@ -0,0 +1,161 @@
...
+
+- **Full support**: Some new features, frequent bugfixes, and security updates are provided regularly.
+- **Maintenance support**: High impact bugfixes and critical security updates are provided as needed.
+
+The only currently supported MicroCloud LTS release is 2.1._z_. This LTS is supported until June 2029 and is currently in full support phase.
+
+(ref-releases-microcloud-feature)=
+### Feature releases
+
+MicroCloud feature releases are pushed out more often and contain the newest features and bugfixes. Since they are less tested than LTS releases, they are **not recommended for production use**.
```
roosterfish: Compared to LXD there are no direct feature releases in MicroCloud. 

MicroCloud has no LXD 6.x equivalent meaning we only have the LTS and `main`/`edge` which receives continuous but not officially released/tagged updates.

Only the LTS receives tagged updates using the point releases.

doc/reference/releases-snaps.md
```diff
@@ -0,0 +1,161 @@
...
+```bash
+snap info microcloud
+```
+
+(ref-snaps-microcloud-tracks)=
+### Tracks
+
+MicroCloud releases are grouped under [snap tracks](https://snapcraft.io/docs/channels#heading--tracks).
+
+The current feature track is {{current_feature_track}}, and the currently supported LTS track is {{current_lts_track}}. The `1` track reached {abbr}`EOL (End of Life)` at the end of April 2025.
```
roosterfish, line 85: I think it's fine that we call the `3` feature track but there are no releases under this track.

I was about to suggest to call it development track but to keep it consistent with LXD keeping the naming but just not doing feature releases makes most sense to me.

doc/reference/releases-snaps.md
```diff
@@ -0,0 +1,161 @@
...
+#### The default track
+
+If you {ref}`install the MicroCloud snap <installing-snap-package>` without specifying a track, the recommended default is used. The default track always points to the most recent LTS track, which is currently {{current_lts_track}}.
+
+(ref-snaps-microcloud-track-latest)=
+#### The `latest` track
+
+In the list of channels shown by `snap info microcloud`, you might see channels with a track named `latest`. This track typically points to the latest feature release.
+
+Since `latest` is a continuously rolling release track, it might become incompatible with your host OS version over time. Due to this, this track is _not recommended for general use_ and might be removed in the future. Instead, use a feature or LTS track.
```
roosterfish: Unlike LXD we only ever update `latest/edge` in MicroCloud according to https://discourse.ubuntu.com/t/microcloud-1-reaches-end-of-life-in-april-2025/57905#p-147753-snap-tracks-and-channels.

Maybe we should mention this here too explicitly?

doc/how-to/install.md
```diff
@@ -84,27 +84,73 @@ For detailed information, see: {ref}`reference-requirements`.
...
+```
 
-```{note}
-Make sure to install the same version of the snaps on all machines.
-See {ref}`howto-snap` for more information.
+#### Upgrade
 
-If you don't want to use MicroCloud's full functionality, you can install only some of the snaps.
-However, this is not recommended.
+If you need to upgrade the channel, run:
```
roosterfish, line 124: Worth linking to the upgrade guide to create awareness that if somebody decides to upgrade one of the services most likely the others need to be upgraded too (according to the compatibility matrix).

doc/how-to/update_upgrade.md
```diff
@@ -31,48 +34,124 @@ To avoid such possible effects entirely, use the live migration approach describ
...
+
+You can hold snap updates either indefinitely or for a specific duration. For full control of snap updates, set an indefinite hold.
+
+To indefinitely hold all updates to the snaps needed for MicroCloud, run:
+
+```bash
+sudo snap refresh --hold lxd microceph microovn microcloud
+```
+
+When you want to update, remove the holds:
```
roosterfish: I don't think you have to `--unhold` just to perform the refresh. The hold can stay as it is and you can trigger the refresh manually.

doc/how-to/snaps.md
```diff
@@ -1,87 +1,78 @@
 (howto-snap)=
 # How to manage the snaps
 
-MicroCloud is distributed as a [snap](https://snapcraft.io/docs).
-The benefit of packaging MicroCloud as a snap is that it makes it possible to include the required dependencies, and that it allows MicroCloud to be installed on many different Linux distributions.
-The snap ensures that MicroCloud runs in a consistent environment.
+Manage MicroCloud and its components (LXD, MicroCeph, and MicroOVN) through their snap packages.
```
roosterfish, line 4: Usually they are called services instead of components. We also have a `microcloud service` command which allows adding an already existing MicroCeph to the MicroCloud.

doc/how-to/update_upgrade.md
```diff
@@ -82,83 +161,136 @@ Furthermore, LXD might automatically update its snaps on cluster members if they
...
+```bash
+sudo snap refresh microceph --channel=<new channel> [--cohort="+"]
+```
 
-If the command succeeds, run the same command on the next machine, and so on.
+Example (for a MicroCeph snap not already `in-cohort`):
 
-```{note}
-Make sure to validate the health of the recently upgraded dependency before continuing with the next one.
+```bash
```
roosterfish: We introduced the upgrade guide for the users being able to switch from MicroCloud 1 to 2. 

That was the reason for having the actual channels in there as a reference. 

Do you think it would make sense to give hints on what should be set here? Because if somebody is just looking for a guide on what to set in order to upgrade from 1 to 2 he has to craft the commands himself by checking the compatibility matrix.

roosterfish: Maybe the example line can just say going from MicroCloud 1 to 2 or similar.

doc/how-to/snaps.md
minaelee: There was discussion a while back about terminology and the term "components" was agreed for use when referring to the snaps/services that make up MicroCloud. This can be seen in the [docs homepage](where it says "The three components of MicroCloud (LXD, MicroCeph, and https://documentation.ubuntu.com/microcloud/v2-edge/microcloud/) MicroOVN) ... "

doc/custom_conf.py
minaelee: Yes — [myst_substitions by default](https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#substitutions-with-jinja2) can use Python key/value pairs in the configuration file or YAML front matter directly in the Markdown files. LXD uses custom code to pull out the substitutions list into a .yaml file and parse it back in within `custom_conf.py`, and I've adopted the same approach here for consistency so it's easy to find.

doc/custom_conf.py
minaelee: Yes, done!

doc/how-to/install.md
minaelee: The compatibility matrix is linked above (where it says "View the {ref}`matrix of compatible versions <ref-releases-matrix>` to determine whether you need to upgrade the snap to a different channel.") and it links to the update/upgrade guide from there. Users should be able to tell from viewing the matrix what snaps they need to upgrade. (Update/upgrade guide is also linked from the matrix.)

doc/how-to/update_upgrade.md
minaelee: I added a new section (called "Example: Upgrade from MicroCloud 1 to 2")  that gives all commands for upgrading from 1 to 2. This can be kept updated whenever we have a new LTS.

doc/how-to/update_upgrade.md
minaelee: You are right. I had copied over previous instructions. Tested to verify that there's no need to --unhold and updated the docs. I'll need to fix this in LXD docs as well.

doc/reference/releases-snaps.md
minaelee: So for `22.04`  all the compatible snap versions should be the same as for 24.04. Fixed.

doc/reference/releases-snaps.md
minaelee: And `main/edge` is published to `3/edge`?

doc/reference/releases-snaps.md
minaelee: Per recent agreement to remove the `latest` track, I removed the section on `latest`.

minaelee:
Test won't pass until https://github.com/canonical/microcloud/pull/839 is merged. Will rebase after.

doc/Makefile
minaelee: Moved into a separate commit without `gitpython` but can't confirm that it will work without until after https://github.com/canonical/microcloud/pull/839 is merged (per tests currently failing due to MicroCeph changes).

roosterfish: PR can be rebased now, the other PR just went in.

doc/custom_conf.py
roosterfish: Ok thanks!

doc/reference/releases-snaps.md
roosterfish: Jep, but it's kind of a rolling version which always receives what we are currently working on. But there might be breaking changes too if we decide something needs further cleanup/enhancements or need to be removed again.
A good example is the addition of cluster manager to MicroCloud. Whilst it is on `3/edge` we can experiment with it and add changes but we won't officially create a release on `3/edge` which would require is to support it going forward.
Doing it like this allows us to still put in breaking changes to the cluster manager integration as we are free to do so on `3/edge`.

doc/how-to/install.md
roosterfish: I see, that looks good like it is.

doc/how-to/snaps.md
roosterfish: Yeah I remember the discussion we had about renaming `microcloud services` command as we intended to use `services` as the term for the now established `cluster links` across the suite of products.
But we didn't progress with this as `services` as a candidate went off the table.

To be consistent I have created https://github.com/canonical/microcloud/issues/840 so we don't forget about deprecating `services` and renaming it to `components`.

doc/how-to/update_upgrade.md
roosterfish: That sounds good.

roosterfish (dismissed):
LGTM!

doc/reference/releases-snaps.md
minaelee: Per Maria, there will be feature releases for MicroCloud, the same as LXD, in the future once there are features to release. So we can leave that in the documentation. I added a couple of sentences to clarify that there are currently no feature releases (to both the `Feature releases > Support` and `Feature track` sections).

PR #830, opened 2025-06-23
doc: Use new stable version of MicroCeph docs (v2-edge)

roosterfish:
The MicroCeph docs version has been moved from squid-stable to v19.2.0-squid which breaks our builds (last two commits).

The PR also backports some necessary commits to further unblock the build of the new version.


markylaing (approved):
LGTM

Issue #797, opened 2025-05-20
Use consistent warning messages and colors


roosterfish:
With the new tui framework we use red for errors and yellow for warning. However this collides with the general highlighting of user provided input which is also using yellow text.

To not get confused the colors yellow and red should only ever be used for warnings and errors. 

Furthermore warning and error messages should be printed consistently using either the `x` for errors or `!` for warnings followed by the actual text in red or yellow letters.

roosterfish:
Hi @n4vxn, thanks for the request. The issue seems to be more involved so I have removed the label. But if you like, would you mind working on https://github.com/canonical/microcloud/issues/695 instead? I have just linked an example in the issue.

PR #834, opened 2025-06-26
CLI: Use consistent colors

roosterfish:
Fixes https://github.com/canonical/microcloud/issues/797

### Overview

The reason for this PR is to be more cautious with colors in the new tui. 
We are using three central colors green, yellow and red to indicate:

* green: positive security related aspects (fingerprints, steps, passphrase)
* yellow: warning messages
* red: error messages

Especially user provided input should not be repeated using the warning color to not confuse the user and to really make warnings and errors stand out.
The new tui styling allows highlighting input using bold text which already is enough as it stands out from regular text.

In addition the way we output errors and warnings is now consistently performed through dedicated functions.
Furthermore we use our own asker (not anymore the one from LXD) so we can influence the error message using our own styling.

### Single node MicroCloud deployment

Screenshot from 2025-06-26 18-39-37

### Multi node MicroCloud deployment

Screenshot from 2025-06-26 18-42-24

### Standard errors

Screenshot from 2025-06-26 18-40-27

### DIFF
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -645,7 +645,7 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
 		}
 
 		if availableDiskCount == 0 {
-			fmt.Println(tui.WarningColor("Warning: No disks available for distributed storage. Skipping configuration", false))
+			tui.PrintWarning("No disks available for distributed storage. Skipping configuration")
 
 			return nil
 		}
@@ -1125,7 +1125,7 @@ func (c *initConfig) askOVNNetwork(sh *service.Handler) error {
 	canOVNUnderlay := true
 	for peer, system := range c.systems {
 		if len(c.state[system.ServerInfo.Name].AvailableOVNInterfaces) == 0 {
-			fmt.Println(tui.WarningColor(fmt.Sprintf("Not enough interfaces available on %s to create an underlay network, skipping", peer), false))
+			tui.PrintWarning(fmt.Sprintf("Not enough interfaces available on %s to create an underlay network. Skipping configuration", peer))
 			canOVNUnderlay = false
 			break
 		}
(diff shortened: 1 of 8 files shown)


cmd/tui/style.go
```diff
@@ -111,7 +111,7 @@ func (*ColorErr) Write(p []byte) (n int, err error) {
 
 // PrintWarning calls Println but it appends "! Warning:" to the front of the message.
 func PrintWarning(s string) {
-	fmt.Printf("%s %s: %s\n", WarningSymbol(), WarningColor("Warning", true), s)
+	fmt.Printf("%s %s %s\n", WarningSymbol(), WarningColor("Warning:", true), WarningColor(s, false))
```
markylaing: Micro optimisation but since these are strings you can just concatenate :)

roosterfish: Good point, replaced it with their `Println` equivalents which adds the space by default.

cmd/microcloud/ask.go
```diff
@@ -1323,7 +1323,7 @@ func (c *initConfig) askCephNetwork(sh *service.Handler) error {
 	availableCephNetworkInterfaces := map[string]map[string]service.DedicatedInterface{}
 	for name, state := range c.state {
 		if len(state.AvailableCephInterfaces) == 0 {
-			fmt.Println(tui.WarningColor(fmt.Sprintf("No network interfaces found with IPs on %q to set a dedicated Ceph network, skipping Ceph network setup", name), false))
+			tui.PrintWarning(fmt.Sprintf("No network interfaces found with IPs on %q to set a dedicated Ceph network. Skipping Ceph network setup", name))
```
minaelee:
```suggestion
			tui.PrintWarning(fmt.Sprintf("No network interfaces found with IPs on %q to set up a dedicated Ceph network. Skipping Ceph network setup", name))
```

roosterfish: Ty, fixed.

PR #838, opened 2025-06-27
CLI: Exit `add` command in case no new system got selected

roosterfish:
In case no new system got selected with the `init` subcommand, the questionnaire falls back to creating a single node MicroCloud. In case of the `add` subcommand, there has to be at least one new system to continue.

The check itself cannot be done inside the initiating session because the same session func is used for both the init and add subcommands.

### DIFF
--- a/cmd/microcloud/add.go
+++ b/cmd/microcloud/add.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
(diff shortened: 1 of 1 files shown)


roosterfish:
The e2e test failure is unrelated to this change.

markylaing (approved):
Makes sense. Thanks.

PR #839, opened 2025-06-27
doc: update to sync with microceph adoption of sphinx-starter-pack

minaelee:
MicroCeph's recent adoption of the [Sphinx Starter Pack](https://github.com/canonical/sphinx-docs-starter-pack) meant changes to its config file to where our MicroCeph integration was no longer compatible. This PR updates our MicroCeph integration, including:
- Updates the Makefile to create folders that no longer exist in the standalone but are needed by the integration
- Updates the Makefile to use `conf.py` instead of `custom_conf.py` for the MicroCeph configuration file
- Adds a new `override-header.css` file because sphinx-starter-pack overwrites the `header.css` file (via the `canonical-sphinx` extension that it uses). This `override-header.css` file only includes the CSS rules we need to change, modified for higher specificity to override the sphinx-starter-pack's `header.css`. 
- Updates `add_config.py` to use new variable names in `conf.py`.

### DIFF
--- a/doc/.custom_wordlist.txt
+++ b/doc/.custom_wordlist.txt
@@ -62,6 +62,7 @@ SVG
 TinyPNG
 uplink
 VLAN
+vlatest
 VM
 VMs
 VPC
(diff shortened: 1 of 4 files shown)


roosterfish (approved):
Thank you, looking great!

PR #851, opened 2025-07-04
Preseed: Fix IPv6 only uplink

roosterfish:
In preseed mode in case only an IPv6 gateway is set on the uplink, instead of the actual address MicroCloud was checking the IPv4 gateway (bug).
This caused MicroCloud not being able to configure an IPv6 only gateway on the uplink when in preseed mode.

Furthermore another test case is added for this scenario.


markylaing (approved):
Good catch :)

PR #857, opened 2025-07-08
CLI: Consistent warning when skipping storage

roosterfish:
When no disks are available, MicroCloud notifies the user that neither local nor remote storage can be configured. For remote storage this was displayed to the user through a warning. Local storage was using a log message which is not directly visible to the user.

This PR ensures that both local and remote storage configuration consistently use the same warning message in case the configuration get skipped:

```bash
root@micro01:~# microcloud init
Do you want to set up more than one cluster member? (yes/no) [default=yes]: no

 Using address 10.87.144.48 for MicroCloud

! Warning: Discovered non-LTS version "6.4" of LXD
Gathering system information ...
! Warning: No disks available for local storage. Skipping configuration
! Warning: No disks available for distributed storage. Skipping configuration
```


roosterfish:
The e2e test failure is unrelated to this change.

Issue #812, opened 2025-06-06
MicroCloud offers to set up distributed storage after adding existing microceph cluster


MggMuggins:
Snap version `git-8a2e3cd` on `latest/edge`

When a MicroCeph cluster already exists, MicroCloud will offer:
```
 "hp-g3-001" is already part of a MicroCeph cluster
Do you want to add this cluster to MicroCloud? (yes/no) [default=yes]: yes
! "hp-g3-001" is already part of a MicroOVN cluster
Do you want to add this cluster to MicroCloud? (yes/no) [default=yes]: no
Would you like to set up local storage? (yes/no) [default=yes]: no
Would you like to set up distributed storage? (yes/no) [default=yes]: 
```

If the user answers "yes" to `Would you like to set up distributed storage? (yes/no) [default=yes]`, it will offer the disks that are already present in the MicroCeph cluster, and then attempt to set up MicroCeph with those disks (& fails).

It looks like the same behavior happens with MicroOVN.

Reported by @desrod

roosterfish:
Hey @MggMuggins, thanks for reporting this. 

I now ran a few tests to better understand the behavior.
At this point I can only speculate why MicroCloud behaves like this, but there seems to be a good reason.
Let's take MicroCeph as an example. In case any of the members is already part of the MicroCeph cluster, the cluster might already have OSDs (even on that member) or there can also just be the cluster without any OSDs.
As MicroCloud doesn't know the users intention, it asks you to configure "distributed storage" which means selecting the disks you want to use for adding new OSDs to the MicroCeph cluster.

However I see where this gets problematic: If you select a disk which is already in use by MicroCeph.
Currently MicroCloud filters out only disks with partitions and considers all other disks to be "free".

Given the fact that you could also have other disks on your system that contain data, I think it's not MicroCloud's responsibility to determine on its own whether or not a disk can be considered "free". Rather the admin should carefully select the disks that should be used.

As MicroCloud is opinionated, the configuration step of the "distributed storage" is not split and we just ask the user for disks and do the remainder of the steps under the hood (bootstrap/join the cluster, add OSD disks, configure the LXD storage pools etc.). 

But there is one bug we can and should address in regards to this report. If the MicroCeph cluster is already bootstrapped and has OSDs and you select `no` to distributed storage, MicroCloud won't setup any of the remote storage pools which feels like a bug to me. Requiring to select at least one disk to create an OSD defeats the purpose of reusing an existing cluster.

[...]

PR #859, opened 2025-07-08
Service: Allow distributed storage config without adding new disks

roosterfish:
Preceded by:
* https://github.com/canonical/microcloud/pull/864
* https://github.com/canonical/microcloud/pull/863

Fixes https://github.com/canonical/microcloud/issues/812.

The PR ensures that an existing (configured) MicroCeph cluster can be reused by MicroCloud without the need of having to add additional disks during the process.

In a scenario with multiple nodes running MicroCeph, a user might decide to leverage the MicroCeph cluster which is already configured with OSDs for MicroCloud. MicroCloud allows this re-use and asks during the interactive questionnaire if the user wants to reuse this existing cluster.
However until now MicroCloud required to add additional disks (OSDs) to be able to include this existing cluster.
This limitation is now removed and the re-use of disks is made more transparent to the user by printing the already configured disks if the user decides to re-use the existing MicroCeph cluster.

### DIFF
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -624,10 +624,68 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
 	}
 	var selectedDisks map[string][]string
 	var wipeDisks map[string]map[string]bool
-	availableDiskCount := 0
+
 	if len(askSystemsRemote) != 0 {
+		// existingClusterDisks contains a slice of disks configured on each of the MicroCeph cluster members.
+		// That allows checking whether or not the configured disks meet the recommendations.
+		existingClusterDisks := map[string][]string{}
+
+		// availableDiskCount is the total amount of unconfigured disks across all of the remote systems.
+		availableDiskCount := 0
+
+		// availableDisks contains a map of unconfigured disks on each of the remote systems.
+		// Those disks aren't yet used for neither local nor remote storage.
 		availableDisks := map[string]map[string]api.ResourcesStorageDisk{}
+
+		// Set to true after checking one of the existing MicroCeph cluster members for existing disks.
+		existingClusterDisksChecked := false
+
 		for name, state := range c.state {
+			// Collect list of already existing remote storage disks.
+			// This allows understanding which disks on which cluster members are already configured for remote storage.
+			// The information can then be yielded to the user and allows skipping the selection of additional disks
+			// in case the user only wants to configure distributed storage without adding additional disks to MicroCeph.
+			// That scenario is important when adding an existing MicroCeph cluster to MicroCloud.
+			if state.ServiceClustered(types.MicroCeph) && !existingClusterDisksChecked {
+				cephService := sh.Services[types.MicroCeph].(*service.CephService)
+				system, ok := c.systems[name]
+				if !ok {
+					return fmt.Errorf("Failed to find system %q", name)
+				}
+
+				var cert *x509.Certificate
+				var address string
+
+				// When asking for remote pool configuration the MicroCloud cluster isn't yet formed.
+				// But we have already established temporary trust with all of the remote systems.
+				// Use the temporary trust store certificate only in case we are not trying to request
+				// the disks from the local system itself.
+				if name != sh.Name {
+					cert = system.ServerInfo.Certificate
+					address = state.ClusterAddress
+				}
+
+				disks, err := cephService.GetDisks(context.TODO(), address, cert)
+				if err != nil {
+					return fmt.Errorf("Failed to get disks of existing %s cluster on %q: %w", types.MicroCeph, name, err)
+				}
+
+				// Only initialize the slice if there are existing disks on this MicroCeph cluster member.
+				// We consolidate the length of the map later to indicate whether or not there are already existing disks.
+				if len(disks) > 0 {
+					existingClusterDisks[name] = []string{}
+				}
+
+				// Fetching the disks from one of the existing MicroCeph cluster members is sufficient.
+				// As the disks are known cluster wide, each member should respond with the same number.
+				for _, disk := range disks {
+					existingClusterDisks[disk.Location] = append(existingClusterDisks[disk.Location], disk.Path)
+				}
+
+				// Skip checking every other MicroCeph cluster member as we have already collected the existing disks.
+				existingClusterDisksChecked = true
+			}
+
 			if askSystemsRemote[name] {
 				availableDisks[name] = state.AvailableDisks
 
--- a/cmd/microcloud/ask.go
+++ b/cmd/microcloud/ask.go
@@ -744,7 +744,6 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
 			insufficientDisks = !useJoinConfigRemote && len(targetDisks) < RecommendedOSDHosts
 
 			if insufficientDisks {
-				// This error will be printed to STDOUT as a normal message, so it includes a new-line for readability.
 				return fmt.Errorf("Disk configuration does not meet recommendations for fault tolerance. At least %d systems must supply disks. Continuing with this configuration will inhibit MicroCloud's ability to retain data on system failure", RecommendedOSDHosts)
 			}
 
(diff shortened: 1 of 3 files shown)


cmd/microcloud/ask.go
```diff
@@ -625,9 +625,46 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
...
 	var wipeDisks map[string]map[string]bool
 	availableDiskCount := 0
+
 	if len(askSystemsRemote) != 0 {
+		totalExistingClusterDisks := 0
+		existingClusterDisks := map[string][]string{}
 		availableDisks := map[string]map[string]api.ResourcesStorageDisk{}
+
 		for name, state := range c.state {
+			if state.ServiceClustered(types.MicroCeph) {
```
markylaing: Can we have a comment above this loop that explains what it does and why? Essentially this can be the contents of your commit message.

cmd/microcloud/ask.go
```diff
@@ -625,9 +625,46 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
...
+				if !ok {
+					return fmt.Errorf("Failed to find system %q", name)
+				}
+
+				var cert *x509.Certificate
+				var address string
+
+				// Use the temporary trust store certificate only in case we are not trying to request
+				// the disks from the local system itself.
+				if name != sh.Name {
```
markylaing, line 663: The certificate is temporary because it hasn't been added to the trust store yet?

Can we change this comment a little as I found it hard to understand what was going on. Maybe:
```
// For connecting to ceph on a different cluster member we need the address and certificate. 
// We are using the temporary certificate because XXX...
// To connect to ceph on this member, we don't need a certificate because we can connect on the unix socket.

File: cmd/microcloud/ask.go [hunk 3]
```diff
@@ -625,9 +625,46 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
...
+				}
+
+				disks, err := cephService.GetDisks(context.TODO(), address, cert)
+				if err != nil {
+					return fmt.Errorf("Failed to get disks of existing %s cluster on %q: %w", types.MicroCeph, name, err)
+				}
+
+				// Fetching the disks from one of the existing MicroCeph cluster members is sufficient.
+				// As the disks are known cluster wide, each member should respond with the same number.
+				if totalExistingClusterDisks == 0 {
```
markylaing: If fetching the disks from one member is sufficient, why are we looping through all of them? Is this because microceph might not be installed on all members? If that's the case, can we figure out where the service is enabled and then query once?

Also should this if statement include a `break`?

cmd/microcloud/ask.go
```diff
@@ -685,115 +685,124 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
...
 		}
 
-		// Ask if the user is okay with fully remote ceph on some systems.
-		if len(askSystemsRemote) != availableDiskCount && wantsDisks {
-			warning := "Unable to find disks on some systems"
-			question := "Continue anyway?"
-			wantsDisks, err = c.asker.AskBoolWarn(warning, question, true)
-			if err != nil {
-				return err
+		if totalExistingClusterDisks > 0 && wantsDisks {
```
markylaing: I'm wondering if we need to keep track of both `totalExistingClusterDisks` and `existingClusterDisks`. It seems we're only ever checking if we have zero or non-zero cluster disks, so maybe we can just have `len(existingClusterDisks) == 0` (or `> 0`).

cmd/microcloud/ask.go
```diff
@@ -779,7 +779,29 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
 					return errors.New("No disks were selected")
 				}
 
-				insufficientDisks = !useJoinConfigRemote && len(targetDisks) < RecommendedOSDHosts
+				mergedDisks := map[string][]string{}
+
+				// Merge both the already existing disks and the new selected ones.
+				// This allows identifying if the selection follows the recommendations.
+				for target, disks := range existingClusterDisks {
```
markylaing: Is `target` here a cluster member? If so, can we either call it that or call it "host" please.

markylaing: Just because then it makes more sense when comparing it to "recommended OSD hosts".

cmd/microcloud/ask.go
roosterfish: Fixed by adding a comment.

cmd/microcloud/ask.go
roosterfish: I updated the comment and added more context.

cmd/microcloud/ask.go
roosterfish: You are right, continuing to loop over the remainder of MicroCeph cluster members is not required (as indicated by the comment).

We still have to continue the loop though to collect available unconfigured disks. But to skip the Ceph check I have added another variable to make this section more performant. Thanks.

cmd/microcloud/ask.go
roosterfish: Good point, I have removed the `totalExistingClusterDisks` now. 

Also I have added comments to the other variables as it might not be clear on a first glance why are all the various maps/slices of disks are used for.

cmd/microcloud/ask.go
roosterfish: `target` is normally used across the code base in equivalent loops. But since we already use "system" (usually for new joiners) and "cluster member" (component specific systems that are already part of the MicroCloud), we can start using different loop var names to make this clearer.

In this case I have used `clusterMember` as it's specific to the MicroCeph component cluster members.

cmd/microcloud/ask.go
```diff
@@ -624,10 +624,63 @@ func (c *initConfig) askRemotePool(sh *service.Handler) error {
...
+
+		// availableDiskCount is the total amount of unconfigured disks across all of the remote systems.
+		availableDiskCount := 0
+
+		// availableDisks contains a map of unconfigured disks on each of the remote systems.
+		// Those disks aren't yet used for neither local nor remote storage.
 		availableDisks := map[string]map[string]api.ResourcesStorageDisk{}
+
+		// Set to true after checking one of the existing MicroCeph cluster members for existing disks.
+		existingClusterDisksChecked := false
```
markylaing, line 641: Sorry this still looks a bit odd to me. Is there a reason this needs to be done inside the loop? i.e. can we just check over the local unix socket for ceph disks once, before getting into this loop?

roosterfish: This `askRemotePool` function gets invoked in various situations:

1. Create MicroCloud with MicroCeph on all cluster members
2. Add an additional system to MicroCloud and add MicroCeph to the MicroCloud
  a. MicroCeph might already be bootstrapped on the local system
  b. MicroCeph might already be bootstrapped on any of the remote systems (requires temporary trust store!)
  c. MicroCeph isn't bootstrapped on any of the systems
3. Add MicroCeph service to the MicroCloud (`microcloud service add` command)

Cases 2 a) and b) are what seems to be confusing here. 
When asking for the remote storage configuration, the underlying Micro* clusters aren't yet formed. That means we cannot just query the local MicroCeph as it might not be a part of the MicroCeph cluster yet. 
That means we can only query for the disks configured in the MicroCeph cluster by reaching out to them. If its our local MicroCeph is clustered, easy as we can just connect to it via the socket.
But if any of the remote system's MicroCeph is clustered we can only reach out to it using the temporary trust store as the MicroCeph cluster isn't yet formed between our local MicroCeph and the remote one. 

The call to `state.ServiceClustered(types.MicroCeph)` returns true for the system which contains an already bootstrapped MicroCeph. In (3) this will be `true` for all systems. 

Also note there cannot be two different MicroCeph clusters (like split brain). MicroCloud will error out if multiple systems contain individual clusters as we cannot easily merge them together.

roosterfish: > Is there a reason this needs to be done inside the loop?

I just reused the loop here. We require it to gather a list of available unconfigured disks. 
Now in addition to the first task we try to get the first system which has an already clustered MicroCeph so we can gather the already configured disks.

markylaing: Ok that makes sense, thank you.

markylaing (approved):
LGTM thanks :)

markylaing:
Looks like there is a failed test though, I'm not sure if that one was already flaky?

PR #860, opened 2025-07-09
doc: backport config and URL updates (v2-edge)

minaelee:
This PR cherry-picks the following commits from `main` to v2-edge`:
* `9d17880` update absolute URL links to LXD docs
* `e5f96c2` doc: update config for move to documentation.ubuntu.com
* `125f226` doc: fix sitemap URL scheme
* `808f420` doc: add sphinx-sitemap to generate sitemap for SEO

A few manual edits were added to use URLs corresponding to the v2-edge docs version (e.g., `stable-5.21` instead of `latest` docs version for LXD)

### DIFF
--- a/doc/custom_conf.py
+++ b/doc/custom_conf.py
@@ -50,7 +50,7 @@ copyright = '%s, %s' % (datetime.date.today().year, author)
 # don't know yet)
 # NOTE: If no ogp_* variable is defined (e.g. if you remove this section) the
 # sphinxext.opengraph extension will be disabled.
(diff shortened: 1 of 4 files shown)


doc/custom_conf.py
```diff
@@ -140,6 +140,22 @@
 else:
     sitemap_url_scheme = '{link}'
 
+#######################
```
roosterfish: This section is duplicated as it was also added in the previous commit. Is that intentional? The only difference seems to be the `sitemap_url_scheme`.

minaelee: Not intentional, just that two commits were misordered. Fixed now!

PR #863, opened 2025-07-10
CLI: Fix profile replacement for storage pools

roosterfish:
Preceded by https://github.com/canonical/microcloud/pull/864

The question to ask for profile device replacement was only applicable for networks because the storage section was missed when migrating the code into a helper in https://github.com/canonical/microcloud/commit/29764bcd6a4e54819a2a575f657970cfb7cc7f7a.
This had the effect that only in case distributed network got configured the question got asked to replace the FAN with OVN network.
However when configuring distributed storage the profile's root device got replaced in any case without asking the user.

Another check is added in this PR to explicitly check the behavior which went unnoticed when refactoring the code.

### DIFF
--- a/test/includes/microcloud.sh
+++ b/test/includes/microcloud.sh
@@ -634,17 +634,17 @@ validate_system_lxd() {
     if [ -n "${profile_pool}" ]; then
       lxc profile device get default root pool | grep -q "${profile_pool}"
     elif [ "${has_microceph}" = 1 ] && [ "${remote_disks}" -gt 0 ] ; then
-       lxc profile device get default root pool | grep -q "remote"
+      lxc profile device get default root pool | grep -q "remote"
     elif [ -n "${local_disk}" ] ; then
-       lxc profile device get default root pool | grep -q "local"
+      lxc profile device get default root pool | grep -q "local"
     else
-       ! lxc profile device list default | grep -q "root" || false
+      ! lxc profile device list default | grep -q "root" || false
     fi
 
     if [ "${has_microovn}" = 1 ] && [ -n "${ovn_interface}" ] ; then
-       lxc profile device get default eth0 network | grep -q "default"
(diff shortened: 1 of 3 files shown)


test/suites/basic.sh
```diff
@@ -1099,7 +1099,9 @@ test_add_services() {
   unset SETUP_OVN
   export REPLACE_PROFILE="no"
   microcloud_interactive "service add" micro01
-  services_validator
+  # The initial cluster got setup with only local storage.
+  # Due to REPLACE_PROFILE="no" the default profile's device doesn't get replaced.
+  services_validator local
```
markylaing, line 1104: `local` is a bash builtin, so it might be a good idea to wrap it in quotes to prevent syntax highlighting or confusing the reader :)

test/includes/microcloud.sh
```diff
@@ -634,17 +634,17 @@ validate_system_lxd() {
...
-       ! lxc profile device list default | grep -q "root" || false
+      ! lxc profile device list default | grep -q "root" || false
     fi
 
     if [ "${has_microovn}" = 1 ] && [ -n "${ovn_interface}" ] ; then
-       lxc profile device get default eth0 network | grep -q "default"
+      lxc profile device get default eth0 network | grep -q "default"
     else
-       lxc profile device get default eth0 network | grep -q "lxdfan0"
+      lxc profile device get default eth0 network | grep -q "lxdfan0"
```
markylaing, line 647: Similar to previous reviews, it would be good to try and update these so that we are testing for an exact match. Not urgent though.

markylaing (approved):
Just some test nits but I think these can all be lumped together in a later PR :)

test/includes/microcloud.sh
roosterfish: Ack, will address it in the follow up PR mentioned on the other PR.

PR #864, opened 2025-07-10
Storage: Fix backup and image vols for new systems

roosterfish:
This fixes a bug when adding new systems to MicroCloud.

If the already existing MicroCloud only has local storage configured, you can decide to configure distributed storage when adding new systems.
In this situation the new systems haven't had their LXD `storage.images_volume` and `storage.backups_volume` settings configured due to a limitation which was setup by MicroCloud which either tracked new or existing pool's.

Another test is added to explicitly check this behavior.
Also a race when enabling optional services is fixed by adding a consequent restart of MicroCloud (see last commit).

### DIFF
--- a/test/suites/add.sh
+++ b/test/suites/add.sh
@@ -91,6 +91,7 @@ test_add_interactive() {
   for m in micro01 micro02 micro03 ; do
     lxc exec "${m}" -- snap disable microovn || true
     lxc exec "${m}" -- snap disable microceph || true
+    lxc exec "${m}" -- snap restart microcloud
   done
 
   lxc exec micro04 -- snap disable microcloud
--- a/test/suites/add.sh
+++ b/test/suites/add.sh
@@ -118,7 +118,7 @@ test_add_interactive() {
   done
 
   reset_systems 4 2 1
-  echo "Test growing a MicroCloud when storage & networks were not already set up"
+  echo "Test growing a MicroCloud when remote storage & networks were not already set up"
   unset_interactive_vars
   export MULTI_NODE="yes"
   export LOOKUP_IFACE="enp5s0"
--- a/test/suites/add.sh
+++ b/test/suites/add.sh
@@ -168,4 +168,63 @@ test_add_interactive() {
     validate_system_microceph "${m}" 1 "${default_cluster_subnet}" disk2
     validate_system_microovn "${m}"
   done
+
+  reset_systems 4 2 1
+  echo "Test growing a MicroCloud when remote storage was not already set up (REPLACE_PROFILE=yes)"
+  unset_interactive_vars
+  export MULTI_NODE="yes"
+  export LOOKUP_IFACE="enp5s0"
+  export EXPECT_PEERS=2
+  export SETUP_ZFS="yes"
+  export ZFS_FILTER="lxd_disk1"
+  export ZFS_WIPE="yes"
+  export SETUP_CEPH="no"
+  export SETUP_OVN="yes"
+  export OVN_FILTER="enp6s0"
+  export IPV4_SUBNET="10.1.123.1/24"
+  export IPV4_START="10.1.123.100"
+  export IPV4_END="10.1.123.254"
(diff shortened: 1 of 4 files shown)


roosterfish:
This bug was never noticed because the `validate_system_lxd` validation func has one drawback.
When setting:

`validate_system_lxd "${m}" 3 disk1 0`

it sets that 0 remote disks are used which falls back to not checking anything Ceph related and also checks if the profile contains the local storage pool. 
In addition it checks for the `storage.*` config settings.

If it's set to 1, it checks all things Ceph related and confirms the profile contains the `remote` storage pool.
However in this case the `storage.*` keys aren't anymore checked even though they are still valid in case a local storage pool gets configured.

To copy with this a manual check is added to the new test case to overcome this limitation of `validate_system_lxd`. Besides this niche case it's behaving fine so I have decided to not modify it for this specific situation.

test/suites/add.sh
```diff
@@ -168,4 +168,63 @@ test_add_interactive() {
...
+  export OVN_FILTER="enp6s0"
+  export IPV4_SUBNET="10.1.123.1/24"
+  export IPV4_START="10.1.123.100"
+  export IPV4_END="10.1.123.254"
+  export IPV6_SUBNET="fd42:1:1234:1234::1/64"
+  export DNS_ADDRESSES="10.1.123.1,fd42:1:1234:1234::1"
+  export OVN_UNDERLAY_NETWORK="no"
+
+  join_session init micro01 micro02 micro03
+  lxc exec micro01 -- tail -1 out | grep "MicroCloud is ready" -q
```
markylaing, line 193: I think we normally test with `grep -F` and without the `-q` flag to make it clearer what is failing when the command fails

test/suites/add.sh
```diff
@@ -91,6 +91,7 @@ test_add_interactive() {
   for m in micro01 micro02 micro03 ; do
     lxc exec "${m}" -- snap disable microovn || true
     lxc exec "${m}" -- snap disable microceph || true
+    lxc exec "${m}" -- snap restart microcloud
```
markylaing, line 94: > MicroCloud has an internal loop which tries to check every second which services are present.
When enabling disabled services during the tests, ensure that MicroCloud picks up the enabled services by restarting it.
>
>This might cause issues from time to time on faster runner machines.

I'm not sure I follow this one. 

Firstly, why does Microcloud check every second which services are available, that seems wasteful. Can we figure this out via snap interfaces? 

Second, if it truly is necessary to poll every second, can we instead just sleep for 1s after this loop is complete - at which point we would expect Microcloud to have picked up available services? This would at least test that the polling works and might even be quicker that restarting the snap.

markylaing (changes requested):
Code changes look fine - just a couple of nits about the tests.

test/suites/add.sh
roosterfish: Happy to land this in a follow up PR and fix it not only for this test but also for all the others (which feel too much for this PR).

test/suites/add.sh
roosterfish: Depending on when MicroCloud gets installed (before or after the other Micro*) it might not have the information about the currently installed services. Therefore the continuous loop.

The daemon starts and it checks if a service exists by reaching out to the service's Microcluster which is possible due to the content interfaces. Then it enters the loop to check regularly and to update this internal list.

That process bugs me too, because it's a waste of resources after the cluster is formed. Afterwards MicroCloud will still continue doing this.
We should rather perform this operation on demand instead of having it as a background operation.

The deamon's `s.Services` is heavily used across the code base so let's create an issue to come up with an improved approach.

roosterfish: Issue created https://github.com/canonical/microcloud/issues/870.

roosterfish: There are already places in the test suite where we restart MicroCloud because of that so this commit just makes the use of this pattern consistent across all the test suites.

PR #865, opened 2025-07-10
GitHub: Split tests into smaller suites

roosterfish:
This allows getting faster test feedback.
Currently the `interactive` test suite runs for over 2h which is tedious.

The PRs finally splits tests into more granular suites.

### DIFF
--- a/test/main.sh
+++ b/test/main.sh
@@ -266,19 +266,11 @@ run_instances_tests() {
   run_test test_instances_launch "instances launch"
 }
 
-run_basic_tests() {
-  run_test test_reuse_cluster "reuse_cluster"
-  run_test test_add_services "add_services"
-  run_test test_remove_cluster_member "remove_cluster_member"
-  run_test test_non_ha "non_ha"
-}
-
-run_recover_tests() {
-  run_test test_recover "recover"
-}
-
 run_interactive_tests() {
   run_test test_interactive "interactive"
+}
+
+run_interactive_combinations_tests() {
   run_test test_interactive_combinations "interactive combinations"
 }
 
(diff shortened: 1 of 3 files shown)


roosterfish:
The e2e failure is unrelated to this change.

test/main.sh
```diff
@@ -287,10 +279,30 @@ run_mismatch_tests() {
...
+}
+
+run_reuse_cluster_tests() {
+  run_test test_reuse_cluster "reuse_cluster"
+}
+
+run_service_tests() {
+  run_test test_add_services "add_services"
+}
+
```
markylaing: Why do we need separate functions for this? Can we just call `run_test` from the if-else block?

I'm just thinking to follow this pattern when adding new tests in the future it's more places to keep track of.

markylaing (changes requested):
Just a small nit to keep things similar to the LXD suite.

test/main.sh
roosterfish: I have left this "extra" layer as there is for example `run_mismatch_tests` which runs both the service mismatch and disk mismatch tests.

So in case we can or have to further split a test it still makes sense to have this grouping in my opinion.

roosterfish: Ah nvm I see what you mean. Let me push a fix for this.

roosterfish: Fixed

roosterfish: Restructured as discussed. Also I found we already have the `run_test "test_${1}"` line so I removed the extra conditions and just left two that combine the instances and mismatch tests.

markylaing: Great stuff thanks!

PR #866, opened 2025-07-11
doc: Remove notion about `--wipe`

roosterfish:
Before releasing 2 LTS the flag was already removed and it's a left over in the docs.


roosterfish:
@minaelee mh, the same cheat sheet error as on `v2-edge`, so probably unrelated to the last doc PR on the `v2-edge` branch: https://github.com/canonical/microcloud/actions/runs/16214365579/job/45780365232?pr=866#step:4:342.

PR #871, opened 2025-07-11
Test: Run joiners in isolated terminals

roosterfish:
The test output is very convoluted as it prints together the output of all systems joining the cluster (as well as the initiator).
By running it in separate terminals we can ensure the outputs are not mixed.

To make this work for each joiner a new tmux session gets spawned to which we write the interactive test input.

### DIFF
--- a/test/includes/microcloud.sh
+++ b/test/includes/microcloud.sh
@@ -246,13 +246,12 @@ $(true)                                 # workaround for set -e
   microcloud_interactive "${mode}" "${initiator}"
   code="$?"
 
-  # Kill the childs if they are still running.
-  child_processes="$(jobs -pr)"
-  if [ -n "${child_processes}" ]; then
-    for p in ${child_processes}; do
-      kill -9 "${p}"
-    done
-  fi
+  # Kill the tmux sessions if they are still running.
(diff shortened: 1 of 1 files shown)


.github/actions/system-test/action.yml
```diff
@@ -32,6 +32,18 @@ runs:
...
+      shell: bash
+      run: |
+        set -eux
+        sudo apt-get update
+
+        sudo apt-get install --no-install-recommends -y \
+          tmux
+
+        # reclaim some space
+        sudo apt-get clean
```
simondeziel: `tmux` is already present on GHA runners so maybe skip this bit or put it behind a `command -v tmux`?

roosterfish: Thank, I should have checked the test's output:

```
+ sudo apt-get install --no-install-recommends -y tmux
Reading package lists...
Building dependency tree...
Reading state information...
tmux is already the newest version (3.2a-4ubuntu0.2).
0 upgraded, 0 newly installed, 0 to remove and 60 not upgraded.
```

I have removed the commit. If they ever decide to remove it from the image, we can still add an install instruction.

PR #894, opened 2025-07-25
Service: Increase the context deadline for adding MicroCeph disks

roosterfish:
During pipeline execution (e.g. https://github.com/canonical/microcloud/actions/runs/16517071789/job/46710255729?pr=893) we sometimes see that the request to add disks to MicroCeph exceeds the request context's deadline.

By default this deadline is set to 30s by the underlying Microcluster client. However when adding a large amount of disks we should give MicroCeph time to perform the necessary actions.

This should address one of the errors in https://github.com/canonical/microcloud/issues/653:

`⨯ Error: Failed to request disk addition: Post "http://control.socket/1.0/services/microceph/1.0/disks?target=micro01": context deadline exceeded`

### DIFF
--- a/service/microceph.go
+++ b/service/microceph.go
@@ -139,6 +139,14 @@ func (s CephService) AddDisk(ctx context.Context, data cephTypes.DisksPost, targ
 		return response, err
 	}
 
+	var cancel context.CancelFunc
+
+	// Allow MicroCeph time to bootstrap the new OSDs.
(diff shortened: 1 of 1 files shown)


roosterfish:
In the long run MicroCeph could maybe have something like LXD's operations which would allow waiting until the job is done.

markylaing (approved):

PR #895, opened 2025-07-25
Test: Add temporary fix for MicroOVN northd issue

roosterfish:
When restarting MicroOVN a new leader gets elected. If this node happens to be the last one that joined the cluster, it's northd service fails to start.

We can temporarily resolve this by restarting the service on the member that joined the cluster last.

### DIFF
--- a/test/e2e/reboot.local-lxd-vm
+++ b/test/e2e/reboot.local-lxd-vm
@@ -33,3 +33,14 @@ lxc exec "${MEMBER}" -- timeout 90 systemctl is-system-running --wait --quiet ||
 sleep 5
 lxc exec "${MEMBER}" --env TEST_CONSOLE=0 -- microcloud waitready --timeout=90 || true
 echo " DONE"
+
+# Temporary fix to deal with https://github.com/canonical/microovn/pull/249.
+#
+# First try to find the cluster member which joined last and restart its ovn-northd service.
+# We can identify it based on its appearance in Microcluster's `core_cluster_members` table.
+# The CLI's `microovn cluster list` cannot be used as the result is sorted.
+# MicroCloud doesn't guarantee any order in which members are joined and doesn't sort them by name (as they are internally stored inside a map).
+last_member="$(lxc exec "${MEMBER}" -- microovn cluster sql 'SELECT name FROM core_cluster_members ORDER BY ID DESC LIMIT 1' | awk -F" " '/micro/ { print $2 }')"
+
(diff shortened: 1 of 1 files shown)


test/e2e/reboot.local-lxd-vm
```diff
@@ -33,3 +33,14 @@ lxc exec "${MEMBER}" -- timeout 90 systemctl is-system-running --wait --quiet ||
 sleep 5
 lxc exec "${MEMBER}" --env TEST_CONSOLE=0 -- microcloud waitready --timeout=90 || true
 echo " DONE"
+
+# Temporary fix to deal with https://github.com/canonical/microovn/pull/249.
+#
+# First try to find the cluster member which joined last and restart its ovn-northd service.
+# We can identify it based on its appearance in Microcluster's `core_cluster_members` table.
```
simondeziel, line 40:
```suggestion
# We can identify it based on its appearance in MicroOVN's `core_cluster_members` table.
```

test/e2e/reboot.local-lxd-vm
```diff
@@ -33,3 +33,14 @@ lxc exec "${MEMBER}" -- timeout 90 systemctl is-system-running --wait --quiet ||
...
+# The CLI's `microovn cluster list` cannot be used as the result is sorted.
+# MicroCloud doesn't guarantee any order in which members are joined and doesn't sort them by name (as they are internally stored inside a map).
```
simondeziel, line 42: MicroCloud or MicroOVN?

roosterfish: It's MicroCloud who instructs MicroOVN to perform the join procedure (member after member). But as MicroCloud stores the list of all joiners in a map, the order in which they are joined cannot be determined upfront. In this case MicroOVN just does what MicroCloud tells it to do.

test/e2e/reboot.local-lxd-vm
roosterfish: The `core_cluster_members` table is coming from Microcluster. In this case I wanted to highlight the specifics of this table, meaning the last member who joined the cluster will appear as the last one in the list when filtered by id. 

And that is something purely implemented in Microcluster, therefore the comment. MicroOVN (as well as all the other Micro*) just make use of this.

roosterfish:
I only ever saw the issue twice in our pipelines. Let's keep this PR open and only merge if we see it blocks us proceeding forward.

PR #893, opened 2025-07-25
Backports (v2-edge)

roosterfish:
Includes all of the recent improvements to the docs as well as fixes for disk selection, profile replacement and the new e2e tests including the waitready enhancements.

The only major feature excluded this time is cluster manager as it won't be included in v2.

Backports until https://github.com/canonical/microcloud/commit/b700eb2691012b4071013859ddd94f8464ce1f9f (July 25, 2025)
Consider those commits too for the next backport https://github.com/canonical/microcloud/pull/893#issuecomment-3131209956 as they were excluded.

PR #896, opened 2025-07-25
doc: fix internal link targets

minaelee:
This PR fixes link targets that are intended to point to MicroCloud internal docs but are pointing at LXD's docs through intersphinx. Due to the link targets existing in LXD latest branch, the links did not raise an error, but the issue showed up in a [backport PR](https://github.com/canonical/microcloud/pull/893) where the LXD branch used in the backport (stable-5.21) does not yet have those link targets.

### DIFF
--- a/doc/how-to/snaps.md
+++ b/doc/how-to/snaps.md
@@ -3,7 +3,7 @@
 
 Manage MicroCloud and its components (LXD, MicroCeph, and MicroOVN) through their snap packages.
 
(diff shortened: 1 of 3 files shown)


roosterfish (approved):
Good catch! Just to clarify, in case there are two refs with the same name on both LXD and MC, does the one from MC take precedence?

roosterfish:
Can we prefix the ones coming from LXD?

PR #901, opened 2025-07-29
version: Set 2.1.1 LTS (v2-edge)

roosterfish:
This has to be merged as the last PR in line right before cutting the next LTS release.


markylaing (approved):
:rocket:

PR #903, opened 2025-07-29
version: Set 3.0 for latest version

roosterfish:
It looks for the last release we set the version on 'main' and then backported it into 'v2-edge'. However MicroCloud 'main' is the latest which should be indicated by the proper version number. Also 3.0 allows having a feature release (like LXD) at some point in the future.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -3,10 +3,10 @@ package version
 
 // RawVersion is the current daemon version of MicroCloud.
 // LTS versions also include the patch number.
-const RawVersion = "2.1.0"
+const RawVersion = "3.0"
 
 // LTS should be set if the current version is an LTS (long-term support) version.
-const LTS = true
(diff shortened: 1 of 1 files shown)


markylaing (approved):
Just to write down my understanding about this from a quick call with @roosterfish:

This means that `latest/edge` microcloud will report it's version as `3.0` but snap list will continue to output `git-{short_sha}`, and `latest/edge` will no longer report itself as being an LTS when it isn't.

We also discussed setting `RawVersion` to `3.0-{git_short_sha}` at build time, but this is a departure from what LXD does and isn't necessary since the snap version will be set.

roosterfish:
> Also this commit will not be backported and the `2.1.0` version will be incremented via PR directly to `v2-edge`.

Yes! Thanks for the summary :)

Issue #516, opened 2024-11-27
Subnet sharing warning should check interfaces not IPs


roosterfish:
The warning in https://github.com/canonical/microcloud/blob/main/cmd/microcloud/main_init.go#L530 isn't fired in case OVN underlay configuration is using an IPv4 and Ceph public/internal traffic is using an IPv6.
 
If both IP networks are configured on the same interface the warning should still be fired to notify the user that a single interface is used for multiple networks that maybe should have been segregated. 

Starting from https://github.com/canonical/microcloud/pull/515 bridges can be used for both the OVN uplink and network segregation so checking the interfaces too would make it more consistent.

PR #522, opened 2024-12-02
microcloud: Subnet sharing warning should check interfaces not IPs

gabrielmougard:
closes https://github.com/canonical/microcloud/issues/516

PR #927, opened 2025-08-13
Fix/remove subnet sharing warning

kadinsayani:
Closes https://github.com/canonical/microcloud/issues/516.

Includes cherry-picks from https://github.com/canonical/microcloud/pull/522.


cmd/microcloud/ask.go
```diff
@@ -258,13 +258,20 @@ func (c *initConfig) askAddress(filterAddress string) error {
 	}
 
 	if subnet == nil {
-		return fmt.Errorf("Cloud not find valid subnet for address %q", listenAddr)
+		return fmt.Errorf("Could not find valid subnet for address %q", listenAddr)
```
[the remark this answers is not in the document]
kadinsayani: It is.

service/lxd_config.go
```diff
@@ -308,3 +308,37 @@ func (s LXDService) DefaultCephFSStoragePoolJoinConfig() api.ClusterMemberConfig
 		Value:  DefaultCephFSOSDPool,
 	}
 }
+
+// FindInterfaceForSubnet takes a subnet in CIDR notation (e.g., "192.168.1.0/24")
+// and returns the network interface that has an IP address in that subnet.
+func (s LXDService) FindInterfaceForSubnet(cidrSubnet string) (*net.Interface, error) {
+	_, targetNet, err := net.ParseCIDR(cidrSubnet)
+	if err != nil {
+		return nil, fmt.Errorf("Invalid CIDR subnet notation %q: %v", cidrSubnet, err)
```
roosterfish: `%w` instead of `%v`.

service/lxd_config.go
```diff
@@ -308,3 +308,37 @@ func (s LXDService) DefaultCephFSStoragePoolJoinConfig() api.ClusterMemberConfig
...
+	}
+
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to list interfaces: %v", err)
```
roosterfish: `%w`.

service/lxd_config.go
```diff
@@ -308,3 +308,37 @@ func (s LXDService) DefaultCephFSStoragePoolJoinConfig() api.ClusterMemberConfig
...
+	}
+
+	for _, i := range interfaces {
+		addrs, err := i.Addrs()
+		if err != nil {
+			continue
```
roosterfish: I wonder if we should not rather return an error here if this doesn't work?

cmd/microcloud/add.go
```diff
@@ -221,9 +228,20 @@ func (c *cmdAdd) Run(cmd *cobra.Command, args []string) error {
...
+		system := cfg.systems[name]
+		err = populateMicroCloudNetworkFromState(state, name, &system, cfg.lookupSubnet)
+		if err != nil {
+			return err
+		}
+
 		cfg.state[name] = *state
 	}
 
+	adderSystem := cfg.systems[cfg.name]
```
roosterfish: Should we rather call `populateMicroCloudNetworkFromState` to perform this action here?

cmd/microcloud/ask.go
```diff
@@ -1493,40 +1500,105 @@ func (c *initConfig) askCephNetwork(sh *service.Handler) error {
...
+		// Update systems with their internal Ceph network representation.
+		for peer, system := range c.systems {
+			peerCephValidatedInterfaces := internalCephNetworkValidatedInterfaces[peer]
+			if len(peerCephValidatedInterfaces) == 0 {
+				continue
+			}
+
+			// We assume that the first interface is the one that the user wants to use.
+			// peerCephValidatedInterfaces[0] is a slice of strings containing the interface name, IP address, and interface type.
+			if len(peerCephValidatedInterfaces[0]) < 3 {
```
roosterfish: Mh I remember making a comment on this structure on the old PR already.

Can we already return the new `Network` struct directly from the `ValidateCephInterfaces` func so we don't have to deal with those three values inside the slice? Also all of the IP/network parsing could then be handled already there and we assign only the `system.MicroCephInternalNetwork` here.

roosterfish: I am mainly asking because the same logic is repeated below.

cmd/microcloud/join.go
```diff
@@ -67,14 +67,14 @@ func (c *cmdJoin) Run(cmd *cobra.Command, args []string) error {
 		cfg.sessionTimeout = time.Duration(c.flagSessionTimeout) * time.Second
 	}
 
-	err = cfg.askAddress(c.flagInitiatorAddress)
+	cfg.name, err = os.Hostname()
```
roosterfish, line 71: Should we add a comment why this has to come first? IIRC `askAddress` is using the hostname.

cmd/microcloud/main_init.go
```diff
@@ -51,19 +51,24 @@ type InitSystem struct {
...
-	MicroCephPublicNetworkSubnet string
-	// MicroCephClusterNetworkSubnet is an optional subnet (IPv4/IPv6 CIDR notation) for the Ceph cluster network.
-	MicroCephInternalNetworkSubnet string
+	// MicroCephPublicNetwork specifies the optional public network configuration for Ceph.
+	// Includes the subnet (IPv4/IPv6 CIDR), network interface name, and IP address within the subnet.
+	MicroCephPublicNetwork *Network
+	// MicroCephInternalNetwork specifies the optional cluster network configuration for Ceph.
+	// Includes the subnet (IPv4/IPv6 CIDR), network interface name, and IP address within the subnet.
+	MicroCephInternalNetwork *Network
+	// MicroCloudInternalNetwork contains the network configuration for the MicroCloud internal network.
```
roosterfish, line 60: Should this comment also have the line `Includes the subnet...`?

cmd/microcloud/ask.go
roosterfish: Also the same block looks to be repeated twice for preseed.

service/lxd_config.go
kadinsayani: I don't think so because we want to try every interface before returning an error.

cmd/microcloud/add.go
kadinsayani: Yes I think that's better, thanks for the suggestion!

cmd/microcloud/main_init.go
kadinsayani: Yes, nice catch.

cmd/microcloud/add.go
kadinsayani: Actually, how about `microCloudNetworkFromStateSystem`? That way we aren't mirroring the name of the function `populateMicroCloudNetworkFromState()`? The `*System` suffix seems to be used elsewhere as well, e.g. `bootstrapSystem`.

cmd/microcloud/ask.go
kadinsayani: I've refactored `validateCephInterfacesForSubnet()` to return `map[string][]NetworkInterfaceInfo`. I've also moved the IP address validation from the call sites to `validateCephInterfacesForSubnet()`.

The `Network` struct has been renamed to `NetworkInterfaceInfo`. I think this name is more descriptive, and it avoids confusion with LXD's `Network` struct.

Overall, I think both of these changes reduce the complexity of the code and we also avoid checking against the magic number of 3.

Thanks for the suggestion :)

service/lxd_config.go
roosterfish: I was thinking in case https://pkg.go.dev/net#Interface.Addrs returns an error, shouldn't we consider this to be a general problem and return early? I think its fine if the returned list is empty so we continue on.

cmd/microcloud/ask.go
```diff
@@ -510,30 +510,54 @@ func (c *initConfig) askLocalPool(sh *service.Handler) error {
...
+			if cephPeerIP == nil {
+				// Attempt to parse the IP address as a CIDR.
+				cephPeerIP, _, err = net.ParseCIDR(iface[2])
+				if err != nil {
+					return nil, fmt.Errorf("Could not parse either Ceph peer IP nor Ceph peer CIDR notation for address %q: %w", iface[2], err)
+				}
+
+				networkInterfaces = append(networkInterfaces, NetworkInterfaceInfo{Interface: net.Interface{Name: iface[1]}, Subnet: internalCephNet, IP: cephPeerIP})
+
+				if !c.autoSetup {
```
roosterfish: Do I understand it right that `iface[2]` can either be just an IP or CIDR? 
Before the "Interface %q detected on cluster member..." sentence was printed regardless of that. Now it will only be printed in case `net.ParseIP` returns a nil value which means its a CIDR address. 

Is that on purpose?
I was just giving this a try, and when you are getting asked during MicroCloud creation to enter a network for either public or internal traffic, in all cases you have to provide a CIDR network address. So can we get rid of the `cephPeerIP == nil` check and just do `net.ParseCIDR` right from the beginning? Or am I missing something?

cmd/microcloud/ask.go
```diff
@@ -544,9 +551,9 @@ func (c *initConfig) validateCephInterfacesForSubnet(lxdService *service.LXDServ
 					fmt.Printf("Interface %q (%q) detected on cluster member %q\n", iface[1], iface[2], iface[0])
 				}
 			}
-
-			validatedCephInterfaces[peer] = networkInterfaces
 		}
+
+		validatedCephInterfaces[peer] = networkInterfaces
```
roosterfish, line 553: Was this a bug when being located inside the inner loop?

cmd/microcloud/ask.go
```diff
@@ -1517,40 +1524,65 @@ func (c *initConfig) askCephNetwork(sh *service.Handler) error {
...
+		}
+	} else {
+		// This is to avoid the situation where the internal network for Ceph has been skipped, but the public network has been set.
+		// Ceph will automatically set the internal network to the public Ceph network if the internal network is not set, which is not what we want.
+		// Instead, we still want to keep the internal Ceph network to use the MicroCloud internal network as a default.
 		bootstrapSystem := c.systems[sh.Name]
-		bootstrapSystem.MicroCephInternalNetworkSubnet = internalCephSubnet
-		c.systems[sh.Name] = bootstrapSystem
+		for peer, system := range c.systems {
+			system.MicroCephInternalNetwork = &NetworkInterfaceInfo{Interface: bootstrapSystem.MicroCloudInternalNetwork.Interface, Subnet: bootstrapSystem.MicroCloudInternalNetwork.Subnet, IP: bootstrapSystem.MicroCloudInternalNetwork.IP}
```
roosterfish: As its both the same type, can we just do 

`system.MicroCephInternalNetwork = bootstrapSystem.MicroCloudInternalNetwork`

?

roosterfish: There are also a few more of those down below.

roosterfish:
The doc failures are unrelated to your PR.

service/lxd_config.go
kadinsayani: Gotcha, makes sense.

cmd/microcloud/ask.go
kadinsayani: `MicroCephInternalNetwork` is a pointer (`*NetworkInterfaceInfo`); we should create a new `NetworkInterfaceInfo` for each system.

cmd/microcloud/ask.go
kadinsayani: Yes, I'll move that change to the correct commit.

cmd/microcloud/ask.go
roosterfish: But wouldn't that be actually right? If it is the exact same network being used, why can't they both reference the same underlying struct using the pointer?

kadinsayani: Ah yes you're right, no need to allocate extra space since we are filling the `NetworkInterfaceInfo` struct with the exact same properties.

cmd/microcloud/ask.go
kadinsayani: I've made it so that message is printed regardless of the address to avoid a behavior change.

Yes, we can just check with `net.ParseCIDR()`.

kadinsayani: `lxdService.ValidateCephInterfaces()` accepts either IP or CIDR (it tries `net.ParseIP()` first, then `net.ParseCIDR()`). We should expect the address to be CIDR when we check in `validateCephInterfaces()`.

roosterfish (approved):
Thank you, looks good to me.

The test failures are all unrelated. You might want to rebase `main` to at least let the read the docs build pass. The e2e one seems to be something on LXD side, I never saw this one before. I'll log it over at https://github.com/canonical/microcloud/issues/653.

roosterfish (approved):

Issue #889, opened 2025-07-24
Add passphrase auto completion


roosterfish:
The new tui framework allows having dedicated models for various types of inputs.
For the interactive questionnaire MicroCloud should grow a new model which allows auto completion of the to be typed passphrase.
As the list of potential words is already known, we can offer the user an improved interface by offering to auto complete all of the four individual words of the passphrase.

The bubbletea `textinput` model already has support for a [`SetSuggestion`](https://pkg.go.dev/github.com/charmbracelet/bubbles/textinput#Model.SetSuggestions) func which accepts a slice of strings which will then be used for auto completion.

It can either be implemented by having four individual text inputs or by having one and by crafting the list of suggestions on demand (separated by a whitespace).
The latter should be preferred as it seems to be more straightforward. Also we don't have to deal with the specific key strokes that would allow jumping between the text inputs.

If the user decides to not type but CTRL+v the password instead, the model should have support for this too.

PR #928, opened 2025-08-13
`microcloud join` passphrase autocompletion

kadinsayani:
Closes https://github.com/canonical/microcloud/issues/889.


service/wordlist.go
```diff
@@ -1301,3 +1305,23 @@ var wordlist = `1111	aardvark
...
+func WordList() []string {
+	lines := strings.Split(wordlist, "\n")
+	words := make([]string, 0, len(lines))
+	for _, line := range lines {
+		if line == "" {
+			continue
+		}
+
+		// Each line is in the form: "digits<TAB>word".
+		parts := strings.SplitN(line, "\t", 2)
```
simondeziel: Could `strings.Cut()` be used here?

kadinsayani: Thanks again! I always miss this one :laughing:

service/util.go
```diff
@@ -0,0 +1,25 @@
...
+func WordList() []string {
+	lines := strings.Split(wordlist, "\n")
+	words := make([]string, 0, len(lines))
+	for _, line := range lines {
+		if line == "" {
+			continue
+		}
+
+		// Each line is in the form: "number<TAB>word".
+		_, word, found := strings.Cut(line, "\t")
```
simondeziel: Actually, I looked at the `worklist` variable and don't know why we keep the `number<TAB>` bit. Seems to always be discarded.

kadinsayani: It was probably easier to copy + paste :)

simondeziel: `wget -qO- https://www.eff.org/files/2016/09/08/eff_short_wordlist_2_0.txt | awk '{print $2}'`

I'd vote to clean it up and avoid needing functions to do it at runtime. See `generatePassphrase()` for another place where this logic is duplicated.

service/wordlist.go
```diff
@@ -2,1302 +2,1303 @@
...
-6654	zestfully
-6655	zigzagged
-6656	zillion
-6661	zipping
-6662	zirconium
-6663	zodiac
-6664	zombie
-6665	zookeeper
-6666	zucchini`
+const Wordlist string = `aardvark
```
roosterfish: The reason I left the original file contents untouched was due to this line in the CC license text:

https://creativecommons.org/licenses/by/3.0/us/

> indicate if changes were made — In 4.0, you must indicate if you modified the material and retain an indication of previous modifications. In 3.0 and earlier license versions, the indication of changes is only required if you create a derivative. 

It wasn't clear to me at which point it is a derivative when we start removing the numbers so I kept the files contents as they are.

cmd/tui/autocomplete.go
```diff
@@ -0,0 +1,141 @@
...
+	keymap      keymap
+	suggestions []string
+	maxTokens   uint8
+}
+
+// keymap is used to define key bindings for the autocomplete model.
+type keymap struct{}
+
+// ShortHelp returns a slice of key bindings for the autocomplete model.
+func (k keymap) ShortHelp() []key.Binding {
```
roosterfish: I think we don't have to print this short help. As soon as you start typing you should get the twist that you can press tab in order to use auto completion.
Also it does not match to the remainder of the ux of this CLI.

But we could extend the "Specify the passphrase for joining the system:" message with another note mentioning that auto completion is possible?

cmd/tui/autocomplete.go
```diff
@@ -0,0 +1,141 @@
...
+
+// autocompleteModel initializes a new autocomplete model with the given prompt and suggestions.
+func autocompleteModel(prompt string, suggestions []string, maxTokens uint8) model {
+	ti := textinput.New()
+	ti.Prompt = prompt + ": "
+	ti.Cursor.Style = lipgloss.NewStyle().Foreground(Bright)
+	ti.PromptStyle = lipgloss.NewStyle().Foreground(Bright)
+	ti.ShowSuggestions = true
+
+	ti.Focus()
```
roosterfish, line 29: Not sure what is the right place, but when typing the four passwords and hitting enter, the cursor still remains at the end of the fourth password. IIRC this can be fixed by calling `ti.Blur()` to hide the cursor once we exit the model.

cmd/tui/autocomplete.go
```diff
@@ -0,0 +1,141 @@
...
+	maxTokens   uint8
+}
+
+// keymap is used to define key bindings for the autocomplete model.
+type keymap struct{}
+
+// ShortHelp returns a slice of key bindings for the autocomplete model.
+func (k keymap) ShortHelp() []key.Binding {
+	return []key.Binding{
+		key.NewBinding(key.WithKeys("tab"), key.WithHelp("tab", "complete")),
```
roosterfish: Do you think `ctrl+(n|p)` is helpful/required in this case?

Let's say you have just ran `microcloud init` on the first node and you remembered the four passwords. Do you now go through each of them using this key stroke until you find the right one, or do you just start typing what you still have in your head?

cmd/tui/autocomplete.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+	}
+
+	var out []string
+	for _, w := range m.suggestions {
+		if token == "" || strings.HasPrefix(w, token) {
+			out = append(out, prefix+w)
+			// Limit the number of suggestions to maxSuggestions.
+			// This is to prevent overwhelming the user with too many suggestions.
+			if len(out) >= maxSuggestions {
```
roosterfish: Mh, in our case it's not really that the user has a choice to select between different options. Instead the suggestions should be an assistance so that the user picks the right password he is still remembering from the other `microcloud init` system. So I don't think a limit should be applied here. We can just append what matches.

service/wordlist.go
simondeziel: What about if we add a comment explaining that we stripped the first column? We could even include the how: `wget -qO- https://www.eff.org/files/2016/09/08/eff_short_wordlist_2_0.txt | awk '{print $2}'`

roosterfish: Yes that sounds reasonable. And in addition we have to do less computing to strip off the numbers every time.

kadinsayani: Sounds good to me!

cmd/tui/autocomplete.go
kadinsayani: Sure, we can remove the short help bit. I think the autocomplete can be implied by the suggested word appearing.

cmd/tui/autocomplete.go
kadinsayani: I've removed the help menu with instructions for `ctrl+(n|p)`. Here is the default `KeyMap` for `textinput`:

```Go
// DefaultKeyMap is the default set of key bindings for navigating and acting
// upon the textinput.
var DefaultKeyMap = KeyMap{
	CharacterForward:        key.NewBinding(key.WithKeys("right", "ctrl+f")),
	CharacterBackward:       key.NewBinding(key.WithKeys("left", "ctrl+b")),
	WordForward:             key.NewBinding(key.WithKeys("alt+right", "ctrl+right", "alt+f")),
	WordBackward:            key.NewBinding(key.WithKeys("alt+left", "ctrl+left", "alt+b")),
	DeleteWordBackward:      key.NewBinding(key.WithKeys("alt+backspace", "ctrl+w")),
	DeleteWordForward:       key.NewBinding(key.WithKeys("alt+delete", "alt+d")),
	DeleteAfterCursor:       key.NewBinding(key.WithKeys("ctrl+k")),
	DeleteBeforeCursor:      key.NewBinding(key.WithKeys("ctrl+u")),
	DeleteCharacterBackward: key.NewBinding(key.WithKeys("backspace", "ctrl+h")),
	DeleteCharacterForward:  key.NewBinding(key.WithKeys("delete", "ctrl+d")),
	LineStart:               key.NewBinding(key.WithKeys("home", "ctrl+a")),
	LineEnd:                 key.NewBinding(key.WithKeys("end", "ctrl+e")),
	Paste:                   key.NewBinding(key.WithKeys("ctrl+v")),
	AcceptSuggestion:        key.NewBinding(key.WithKeys("tab")),
	NextSuggestion:          key.NewBinding(key.WithKeys("down", "ctrl+n")),
	PrevSuggestion:          key.NewBinding(key.WithKeys("up", "ctrl+p")),
}
```

We can use our own `KeyMap` that excludes `NextSuggestion` and `PrevSuggestion`, however, I think it is a useful feature.

cmd/tui/autocomplete.go
kadinsayani: Thanks, I've added that :)

cmd/tui/autocomplete.go
kadinsayani: Sounds good!

kadinsayani:
> Pipeline is unhappy though :(

This was due to the `service/wordlist.go` file having a `test` build tag, which prevented its compilation during `make build-test` runs. I've removed the tag.

kadinsayani:
> Looks great @kadinsayani! Just tested it and I really like how clean and "simple" the new model is implemented.
> 
> Left a few minor comments. About the CC license topic not sure how we should proceed here, maybe we just add another line mentioning the modifications to be on the safe side.

Thanks for your review! I looked into the CC license and as long as we reasonably mention our changes we're good to go :)

simondeziel (dismissed):
LGTM, thanks for simplifying the wordlist!

service/session.go
```diff
@@ -36,22 +36,17 @@ type Session struct {
 // generatePassphrase returns four random words chosen from wordlist.
 // The words are separated by space.
 func generatePassphrase() (string, error) {
-	splitWordlist := strings.Split(wordlist, "\n")
+	splitWordlist := strings.Split(Wordlist, "\n")
```
simondeziel: The `Wordlist` is always turned into a slice before using it. Should it be exported as a slice instead of a multi-line string?

kadinsayani: I've added an exported `var`. I like that approach better, thanks :)

simondeziel (dismissed):

service/session.go
roosterfish: Mh, now we are doing the splitting just at a different place. If I understood @simondeziel right, he was thinking about exporting it as a slice right from the beginning?

I think it's a matter of taste how we do it, I am happy with just splitting it once (like it is right now) as it might be easier to just swap out the multi line string if we at some point decide to use different words.
We are anyway accessing the list only once so it shouldn't be too resource intensive to iterate over it for splitting.

kadinsayani: > it might be easier to just swap out the multi line string if we at some point decide to use different words.

That's what I was thinking :)

cmd/tui/autocomplete.go
```diff
@@ -0,0 +1,108 @@
...
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
+)
+
+// Based off https://github.com/charmbracelet/bubbletea/blob/main/examples/autocomplete/main.go.
+
+// model is the Bubble Tea model for passphrase autocomplete.
+type model struct {
+	textInput   textinput.Model
+	help        help.Model
```
roosterfish: Do we still need this field now after removing the help message?

cmd/tui/autocomplete.go
```diff
@@ -0,0 +1,108 @@
...
+	var cmd tea.Cmd
+	m.textInput, cmd = m.textInput.Update(msg)
+
+	// Refresh token suggestions on every change so ctrl+n/p work.
+	m = m.refreshTokenSuggestions(m.maxTokens)
+
+	return m, cmd
+}
+
+// View renders the model as a string, including the text input and help information.
```
roosterfish: Comment requires an update now as the help isn't anymore rendered.

service/test_wordlist.go
```diff
@@ -1,9 +0,0 @@
-//go:build test
```
roosterfish: Mhh, not sure if we can delete it. 
For the test pipeline we build MicroCloud using `make build-test` which sets the `test` build tag and makes sure only `a` is populated as a valid word.
This helps running the tests because we don't have to parse the random passphrase but can instead always supply `a a a a` as the passphrase as it cannot be a different one.

Probably you need to revert the change that removed the tag from the `wordlist.go` file and change the name of this variable from `wordlist` to the now exported `Wordlist`.

roosterfish (changes requested):

service/session.go
simondeziel: I was thinking of exporting just a ready made slice but I'm happy with either ways.

cmd/tui/autocomplete.go
kadinsayani: We don't need this field anymore.

service/test_wordlist.go
kadinsayani: Thanks for the info, I've addressed this and tests are passing now :heavy_check_mark:

cmd/tui/handler.go
```diff
@@ -167,3 +168,49 @@ func (i *InputHandler) AskString(question string, defaultAnswer string, validato
...
+
+// AskPassphrase renders an interactive input with autocomplete suggestions and returns the entered passphrase.
+func (i *InputHandler) AskPassphrase(question string, suggestions []string, validator func(string) error, maxTokens uint8) (string, error) {
+	i.setActive(true)
+	defer i.setActive(false)
+
+	for {
+		var answer string
+		var err error
+		if os.Getenv("TEST_CONSOLE") != "1" {
```
roosterfish: Is there a specific reason the new autocomplete model cannot be used when `TEST_CONSOLE=1`?

cmd/tui/autocomplete.go
roosterfish: Makes sense thanks.

cmd/tui/handler.go
kadinsayani: I wanted to keep things simple and avoid running the bubble tea model when terminal interactivity is disabled.

roosterfish: I see. But we might skip over regressions if we skip the entire model during the pipeline tests.

kadinsayani: Good point, I'll remove the guard and get the model working with the test console :)

roosterfish:
Not sure if helpful, but we once had a custom "asker" tui model used for simple text input which got replaced with something more simple but it might contain bits you might find helpful for this autocomplete model. See https://github.com/canonical/microcloud/pull/505/files#diff-df35a010729fac291c6f88376c6332622be677149e0abcc22e32c77ddb9891ae.

test/includes/microcloud.sh
```diff
@@ -219,6 +219,7 @@ join_session(){
 $([ -n "${LOOKUP_IFACE}" ] && printf "table:select")          # select the interface
 $([ -n "${LOOKUP_IFACE}" ] && printf -- "table:done")
 a a a a                                 # the test passphrase
+$(printf "table:done")					# autocomplete model requires carriage return
```
roosterfish: Do we have to print `table:done` or can we just do `""`? As we are not entering inputs for a table in this case (only a text input), we should not create confusion here by supplying the specific `table:*` control commands.

Maybe also this works:

```
a a a a                                 # the test passphrase
<--- just do a blank line here
$(true)                                 # workaround for set -e
```

roosterfish (changes requested):
Looks great, just one more comment on the carriage return to not create confusion.

test/includes/microcloud.sh
roosterfish: Ah now I see what you did there. `table:done` gets interpreted to just do a `\r`. Maybe let's define a dedicated `ctrl:cr` input action in the corresponding switch statement to make this more clear?

```go
case "table:done", "ctrl:cr":
    action = "\r"
```

kadinsayani: Sure, sounds good! I went with `ctrl:m` to match the control character command `^M`.

cmd/microcloud/test_input.go
```diff
@@ -36,8 +36,9 @@ table:wait <time.Duration>  # waits before the next instruction
 table:expect <count>        # waits until exactly <count> peers are available, and errors out if more are found
 table:filter <text>         # applies filtering text to the table output
 table:done                  # confirms the table selection and exits the table
+ctrl:m						# carriage return, used to submit text input
```
simondeziel, line 39: I would probably have gone with `table:submit` or similar to avoid having to deal with another prefix but that's really a tiny nitpick.

simondeziel (approved):

cmd/microcloud/test_input.go
roosterfish: In this case there is no table (like address selection or member selection) but just the text input field that expects some input. 
IMHO using the `table:` ctrl commands I think would be confusing.

PR #934, opened 2025-08-19
doc: Update broken link to MyST syntax guide

roosterfish:


roosterfish:
@minaelee something seems to be broken with the link/spell checkers. I saw it failing for some other PRs and thought it's just this link. But looks like there is more to it. Do you have an idea?

minaelee:
RTD build is failing for another reason, due to a change I made in LXD docs. I will fix that part.

roosterfish:
@minaelee should we exclude the link for now from the checks? It currently blocks all the edge builds as the doc stage is required before continuing on.

minaelee (approved):

PR #945, opened 2025-08-26
doc: Add en.wikipedia.org to link ignore list

roosterfish:
It looks wikipedia is blocking our link checks due to too many requests, see https://github.com/canonical/microcloud/actions/runs/17230818667/job/48884267655.

### DIFF
--- a/doc/custom_conf.py
+++ b/doc/custom_conf.py
@@ -166,6 +166,8 @@ linkcheck_ignore = [
     'https://ceph.io',
     # Cloudflare protection on SourceForge domains often block linkcheck
     r"https://.*\.sourceforge\.net/.*",
+    # Wikipedia may return 403 possibly due to too many requests
+    r"https://en\.wikipedia\.org/.*",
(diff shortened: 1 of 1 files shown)


minaelee:
> @minaelee do you have an idea why the exception isn't picked up? Is it a different kind of check that is not using this ignore list?

For links added to `custom_linkcheck_anchors_ignore_for_url`, sphinx still performs linkchecking on the URL while ignoring any page anchors (e.g., for `https://somedomain.com/somepage#someanchor`, still linkchecks `https://somedomain.com/somepage`). Since the wikipedia links aren't having issues with anchors but seem to be getting blocked with 403 forbidden, add instead to the `linkcheck_ignore` list (starts on line 162) to skip linkchecking altogether.

minaelee:
PR #946 should fix it, and you can close this one.

roosterfish:
> PR #946 should fix it, and you can close this one.

Thank you!

PR #953, opened 2025-09-03
Test: include the Terraform demo in the test suite

mseralessandri:


roosterfish:
Can be rebased once https://github.com/canonical/microcloud/pull/957 is merged.

.github/actions/system-test/action.yml
```diff
@@ -123,7 +123,11 @@ runs:
         export MICROCLOUD_SNAP_CHANNEL="${{ matrix.microcloud }}"
 
         cd test
-        sudo --preserve-env=GOCOVERDIR,DEBUG,GITHUB_ACTIONS,MICROCLOUD_DEBUG_PATH,MICROCLOUDD_DEBUG_PATH,SKIP_VM_LAUNCH,SNAPSHOT_RESTORE,TEST_STORAGE_SOURCE,TESTBED_READY,BASE_OS,LXD_SNAP_CHANNEL,MICROCEPH_SNAP_CHANNEL,MICROOVN_SNAP_CHANNEL,MICROCLOUD_SNAP_CHANNEL ./main.sh setup
+        if [ "${{ matrix.suite}}" = "terraform_demo" ]; then
```
simondeziel: This GHA syntax is close to `bash` but not quite. I think the surrounding spaces are needed:
```suggestion
        if [ "${{ matrix.suite }}" = "terraform_demo" ]; then
```

.github/actions/system-test/action.yml
```diff
@@ -123,7 +123,11 @@ runs:
         export MICROCLOUD_SNAP_CHANNEL="${{ matrix.microcloud }}"
```
roosterfish, line 132: Can you please split this commit into multiple smaller ones? One that modifies the `system-test` action, one that adds the new test suite and so on.

.github/actions/system-test/action.yml
```diff
@@ -123,7 +123,11 @@ runs:
...
 
         cd test
-        sudo --preserve-env=GOCOVERDIR,DEBUG,GITHUB_ACTIONS,MICROCLOUD_DEBUG_PATH,MICROCLOUDD_DEBUG_PATH,SKIP_VM_LAUNCH,SNAPSHOT_RESTORE,TEST_STORAGE_SOURCE,TESTBED_READY,BASE_OS,LXD_SNAP_CHANNEL,MICROCEPH_SNAP_CHANNEL,MICROOVN_SNAP_CHANNEL,MICROCLOUD_SNAP_CHANNEL ./main.sh setup
+        if [ "${{ matrix.suite}}" = "terraform_demo" ]; then
+          echo "Skipping the setup for the Terraform demo"
```
roosterfish: I wonder if we should rather add another input parameter to the action instead?
Something like `skip_testbed` so you can either pass `true` or `false` to indicate whether or not the action should setup the testbed (depending on the current test):

`skip_testbed: "${{ matrix.suite }}" = "terraform_demo"`

Checking for a specific test suite name inside the action doesn't feel right.

See https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#inputs

.github/actions/system-test/action.yml
```diff
@@ -133,7 +137,7 @@ runs:
         echo "MICROCLOUD_SNAP_CHANNEL=${MICROCLOUD_SNAP_CHANNEL}" >> "${GITHUB_ENV}"
 
     - name: Setup Terraform
-      if: ${{ matrix.suite == 'e2e' }}
+      if: ${{ matrix.suite == 'e2e' || matrix.suite == 'terraform_demo' }}
```
roosterfish: Maybe lets also have a new input parameter `skip_terraform` to indicate whether or not we want to setup terraform?

We could also go reverse and name those `setup_testbed` and `setup_terraform`.

test/suites/terraform_demo.sh
```diff
@@ -0,0 +1,154 @@
...
+VOLUMES="local1 local2 local3 local4 remote1 remote2 remote3"
+STORAGE_POOL="default"
+VM_COUNT=4
+CEPH_EXCLUDED_VM="micro4"
+OVN_INTERFACE="enp6s0"
+OVN_IPV4_GATEWAY="192.0.2.1/24"
+OVN_IPV4_RANGE="192.0.2.100-192.0.2.200"
+LOCAL_DISKS="local1,local2,local3,local4"
+CEPH_DISKS="remote1,remote2,remote3"
+
```
roosterfish, line 15: One newline should be enough.

test/suites/terraform_demo.sh
```diff
@@ -0,0 +1,154 @@
...
+  # Use the exact same configuration as the example to test what users would actually use
+  cp terraform.tfvars.example terraform.tfvars
+  
+  echo "==> Initializing Terraform"
+  terraform init
+  
+  echo "==> Validating Terraform configuration"
+  terraform validate
+  
+  echo "==> Planning Terraform deployment"
```
roosterfish: Do we require this `plan` step here? Isn't the same printed as part of `apply` too?

test/suites/terraform_demo.sh
```diff
@@ -0,0 +1,154 @@
...
+  terraform plan
+  
+  echo "==> Applying Terraform deployment"
+  terraform apply -auto-approve
+  
+  # Wait for deployment to complete - terraform demo includes cloud-init setup
+  echo "==> Waiting for MicroCloud initialization to complete"
```
roosterfish: When `terraform apply` is done, isn't the initialization already done as it waits for `cloud-init` to complete? So I wonder if instead of "Waiting for Microcloud ...` this should rather print "MicroCloud initialization is complete".

test/suites/terraform_demo.sh
```diff
@@ -0,0 +1,154 @@
...
+  # Check that all VMs are running 
+  echo "==> Verifying all VMs are running"
+  running_vms=$(lxc list -f csv -c ns | grep -cxE 'micro[1-4],RUNNING')
+  if [ "${running_vms}" -ne "${VM_COUNT}" ]; then
+    echo "ERROR: Expected ${VM_COUNT} running VMs, found ${running_vms}"
+    lxc list micro  # Show current state for debugging
+    exit 1
+  fi
+  
+  # Wait for MicroCloud to be fully ready - terraform demo uses preseed, so check status directly
```
roosterfish: This comment doesn't match to what the command is doing. `microcloud status` simply outputs a status but doesn't wait on anything to complete.

roosterfish: Also doesn't `terraform apply` print the output of `microcloud status` already? I think there is an output configured.

test/suites/terraform_demo.sh
```diff
@@ -0,0 +1,154 @@
...
+  echo "==> Waiting for MicroCloud to be fully ready"
+  lxc exec micro1 -- microcloud status
+  
+  echo "==> Validating MicroCloud cluster configuration"
+  for m in ${VMS}; do
+    # Validate LXD configuration using variables
+    # Parameters: name, num_peers, local_disk, remote_disks, cephfs, ovn_interface, ipv4_gateway, ipv4_ranges, ipv6_gateway, dns_nameservers, profile_pool
+    validate_system_lxd "${m}" "${VM_COUNT}" "${LOCAL_DISKS}" 3 0 "${OVN_INTERFACE}" "${OVN_IPV4_GATEWAY}" "${OVN_IPV4_RANGE}" "" "" "${STORAGE_POOL}"
```
roosterfish: I like that those are being reused here too.

roosterfish:
Also the job seems to exit with non-zero code. I wonder if we need to set the `TEST_RESULT="success"` variable? I remember this from the modified version of the testbed setup script I am using locally for testing.

mseralessandri:
I hit the

test/main.sh
```diff
@@ -274,6 +275,9 @@ elif [ "${1}" = "mismatch" ]; then
   run_test test_disk_mismatch
 elif [ "${1}" = "setup" ]; then
   testbed_setup
+elif [ "${1}" = "terraform_demo" ]; then
+  TESTBED_READY=1
+  run_test "test_${1}"
```
roosterfish: Not a blocker but in all the other cases (except the fallback to run any test) we use `test_<name>` followed by its actual name instead of the variable:

I would prefer `run_test "test_terraform_demo"` for consistency. Maybe you could also move this block under the "mismatch" one so that all the custom test set specifications are together.

test/suites/terraform_demo.sh
```diff
@@ -0,0 +1,181 @@
...
+  echo "Testing Terraform demo deployment workflow"
+  
+  # Set up cleanup trap to ensure cleanup happens even on failure
+  trap terraform_demo_cleanup EXIT
+  
+  lxc remote switch local
+  
+  # Copy terraform demo to a test location to avoid modifying the original
+  cp -r ../demos/terraform /tmp/terraform_demo_test
+  cd /tmp/terraform_demo_test || exit 1
```
roosterfish, line 71: In which case is this `exit 1` required? The directory gets created the line before.

test/suites/terraform_demo.sh
```diff
@@ -0,0 +1,181 @@
...
+    if ! validate_system_microovn "${m}"; then
+      echo "ERROR: MicroOVN validation failed for ${m}"
+      exit 1
+    fi
+    echo "  -> MicroOVN validation passed for ${m}"
+  done
+  
+  # Test basic cluster functionality by creating and managing a container
+  echo "==> Testing basic cluster functionality"
+  if ! lxc exec micro1 -- lxc launch ubuntu:24.04 test-container; then
```
roosterfish: The other tests use `ubuntu-minimal-daily:24.04` as we can pull this image through devlxd as its getting pre-loaded by the testbed. 
In this case there is no testbed but should we still use the same image here too?

mseralessandri: I think so for consistency

test/suites/terraform_demo.sh
mseralessandri: I added `|| exit 1`  because of the `SHELL_CHECK` warning
```
✋ Defects, NEEDS INSPECTION
Error: SHELLCHECK_WARNING:
./test/suites/terraform_demo.sh:71:3: warning[SC2164]: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
#   68|     
#   69|     # Copy terraform demo to a test location to avoid modifying the original
#   70|     cp -r ../demos/terraform /tmp/terraform_demo_test
#   71|->   cd /tmp/terraform_demo_test 
```

roosterfish: Ok thanks! I find it odd why this is flagged by shellcheck.

Because when you intend to change the directory but it doesn't exist, doesn't `cd` return a non-zero exit code already? What is the need for the extra `|| exit 1`.

Not a blocker, just curious.

mseralessandri: `cd` returns a non-zero result in case of error, but the script continues the execution, that's the reason why I think `SHELLCHECK` signals that  `|| exit 1` is needed.

roosterfish: Ah yes you are right, thanks.

Issue #949, opened 2025-08-31
microcloud init fingerprint doesn't match until sending a pass phrase


nobuto-m:
microcloud  2.1.1-202e275  1781  2/stable  canonical✓  in-cohort

To have a working cluster, users must go ahead by ignoring instructions about the fingerprint instruction, which is concerning.

[on node-1]

```bash
ubuntu@mc-1:~$ sudo microcloud init
Waiting for services to start ...
Do you want to set up more than one cluster member? (yes/no) [default=yes]: 

 Using address 192.168.123.194 for MicroCloud

Use the following command on systems that you want to join the cluster:
                                                                       
 microcloud join                                     
                                     
When requested, enter the passphrase:
                                     
 synagogue identical popcorn yiddish                       
                       
Verify the fingerprint 12e80c412f77 is displayed on joining systems.
Systems will appear in the table as they are detected. Select those that should join the cluster:
```

-> It directs users to verify the fingerprint `12e80c412f77` is displayed on the joining system.

[on node-2]
```bash
ubuntu@mc-2:~$ sudo microcloud join
Waiting for services to start ...

 Using address 192.168.123.117 for MicroCloud

Verify the fingerprint 2bb7dc187dda is displayed on the other system.
Specify the passphrase for joining the system:
```

-> At this step, users are requested to type the passphrase but the displayed fingerprint is `2bb7dc187dda`, which is different from the one on node-1.

By ignoring the instruction about the fingerprint and providing the passphrase, both side can see each other's fingerprint and users can verify. However, it happens **after** providing the passphrase.

[on node-1]

```shell
Systems will appear in the table as they are detected. Select those that should join the cluster:          
 Filter | <type to filter rows>
   ┌──────┬─────────────────┬──────────────┐
   │ NAME │     ADDRESS     │ FINGERPRINT  │
   ├──────┼─────────────────┼──────────────┤
 > │ mc-2 │ 192.168.123.117 │ 2bb7dc187dda │
   └──────┴─────────────────┴──────────────┘
 space to select; enter to confirm
 ↑/↓ to move; → to select all; ← to select none
```

[on node-2]

```bash
Specify the passphrase for joining the system: synagogue identical popcorn yiddish
Searching for an eligible system ...

 Found system mc-1 at 192.168.123.194 using fingerprint 12e80c412f77

Select mc-2 on mc-1 to let it join the cluster
```

nobuto-m:
[...]

PR #954, opened 2025-09-03
Questionnaire: Improve flow for fingerprint verification

roosterfish:
Fixes https://github.com/canonical/microcloud/issues/949.

We should not tell the user to verify the joiners fingerprint before this can actually be done on the initiator to avoid confusion.

Furthermore it fixes a few smaller inconsistencies in the new autocomplete tui model.


kadinsayani (approved):
LGTM, thanks!

PR #978, opened 2025-09-08
Autocomplete fixes

### LINKED ISSUE #959: Autocomplete model doesn't recognize newline on copy & paste
Before having the autocomplete model for typing the passphrase, you could simply copy & paste the passphrase including a newline from the initiator and continue on the joiner by pasting the passphrase which then already submitted the passphrase without having the need to press the enter/return key.

In the new model this newline isn't considered anymore and you have to paste and press the enter key. 
The autocomplete model should recognize the newline and proceed further if encountered. This is due to convenience and to keep the old behavior.

kadinsayani:
Closes #959.

### DIFF
--- a/cmd/tui/autocomplete.go
+++ b/cmd/tui/autocomplete.go
@@ -44,6 +44,13 @@ func (m model) Init() tea.Cmd {
 func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	switch msg := msg.(type) {
 	case tea.KeyMsg:
+		// Auto-submit on paste.
+		if msg.Paste {
+			m.textInput.SetValue(m.textInput.Value() + strings.TrimSpace(string(msg.Runes)))
+			m.textInput.Blur()
+			return m, tea.Quit
+		}
+
 		switch msg.Type {
 		case tea.KeyEnter, tea.KeyCtrlJ:
 			m.textInput.Blur()
--- a/cmd/tui/autocomplete.go
+++ b/cmd/tui/autocomplete.go
@@ -45,10 +45,11 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	switch msg := msg.(type) {
 	case tea.KeyMsg:
 		switch msg.Type {
-		case tea.KeyEnter, tea.KeyCtrlC, tea.KeyEsc:
+		case tea.KeyEnter, tea.KeyCtrlJ:
 			m.textInput.Blur()
 			return m, tea.Quit
-
+		case tea.KeyCtrlC, tea.KeyEsc:
+			return m, tea.Interrupt
 		case tea.KeyTab:
 			m = m.refreshTokenSuggestions(m.maxTokens)
 		}
--- a/cmd/tui/handler.go
+++ b/cmd/tui/handler.go
@@ -180,6 +180,10 @@ func (i *InputHandler) AskPassphrase(question string, suggestions []string, vali
 		p := tea.NewProgram(m, tea.WithInput(i.input), tea.WithOutput(i.output))
 		result, err := p.Run()
 		if err != nil {
+			if errors.Is(err, tea.ErrInterrupted) {
+				os.Exit(1)
+			}
+
 			return "", err
 		}
 


cmd/microcloud/session.go
```diff
@@ -85,12 +85,11 @@ func (c *initConfig) initiatingSession(gw *cloudClient.WebsocketGateway, sh *ser
 When requested, enter the passphrase:
 
  %s
-
 Verify the fingerprint %s is displayed on joining systems.
 `
 
 		cmdArg := tui.Fmt{Arg: "microcloud join", Color: tui.Green, Bold: true}
-		passArg := tui.Fmt{Arg: session.Passphrase, Color: tui.Green, Bold: true}
+		passArg := tui.Fmt{Arg: session.Passphrase + "\n", Color: tui.Green, Bold: true}
```
roosterfish: Isn't it the same as leaving the newline under the `%s` in the template?

cmd/tui/autocomplete.go
```diff
@@ -45,10 +45,12 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 	case tea.KeyMsg:
 		switch msg.Type {
-		case tea.KeyEnter, tea.KeyCtrlC, tea.KeyEsc:
+		case tea.KeyEnter, tea.KeyCtrlJ:
 			m.textInput.Blur()
 			return m, tea.Quit
-
+		case tea.KeyCtrlC, tea.KeyEsc:
+			m.textInput.SetValue("")
+			return m, tea.Quit
```
roosterfish: We have to return something else than `tea.Quit` to allow Ctrl+c to exit the `microcloud join` command. Otherwise you end up in a situation where you have to type in the password, otherwise you cannot exit out as the passphrase never matches the validator.

Just checked what other `tea` messages are there, what about `tea.Interrupt`? We need something that produces a non nil error so that we exit out of this retry loop. 
Maybe we can catch the error and just quit to resemble the same behavior as with regular text inputs (e.g. yes/no) which also just quits when you press CTRL+c.

cmd/tui/autocomplete.go
```diff
@@ -44,6 +44,18 @@ func (m model) Init() tea.Cmd {
 func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	switch msg := msg.(type) {
 	case tea.KeyMsg:
+		// Auto-submit if a paste contains ends with newline.
```
roosterfish: Was giving it a try but it doesn't work for me so let me try to explain you what is currently working in `2/stable`.

You can left click three times to highlight and select the entire row (which contains a newline):

Screenshot from 2025-09-09 09-54-01

Then when you paste the passphrase on one of the joiners, its submitting the text input right away as it contains a newline:

Screenshot from 2025-09-09 09-55-10

Maybe this extra whitespace in front of the first word of the passphrase is setting it off?

roosterfish (changes requested):
Thanks for taking a look at this so quickly. Please check my comments, maybe I am missing something.

cmd/tui/autocomplete.go
kadinsayani: `tea.Interrupt` works well, thanks for the suggestion!

image

cmd/microcloud/session.go
kadinsayani: Yes, it should be. However, the `\n` was not being picked up unless appended to `session.Passphrase` when we set `Arg`. 

I think this actually has to do with my terminal emulator stripping the `\n` when selecting text. The `\n` is only detected when I explicitly include an empty line like this:

image

cmd/tui/autocomplete.go
kadinsayani: Due to observed inconsistencies across terminal emulators when selecting text containing `\n` characters, I'm going to rework this PR to follow a different approach. I think we can either simulate a null-terminated C style string by appending `\x00`, or count the number of words and auto-submit if they match `PassphraseWordCount` (4).

kadinsayani: On second thought, to avoid complicating things and running into edge cases for different terminal emulators, we can just auto-submit the trimmed input on paste, regardless of string content. This will suffice because we already print errors and allow for retries when a passphrase is incomplete (<4 words) or incorrect.

cmd/microcloud/session.go
roosterfish: Ah ok that is interesting. I was only testing this with the regular terminal on my Ubuntu system.

cmd/tui/autocomplete.go
roosterfish: Should we omit the `program was killed: ...` text and simply exit out? If you for example quit any of the other text inputs we also don't print any extra error message for Ctrl+c:

```bash
root@micro03:~# microcloud init
Do you want to set up more than one cluster member? (yes/no) [default=yes]: ^C
```

Maybe there is a specific error type you can catch using `errors.Is` and simply exit.

cmd/tui/autocomplete.go
roosterfish: That sounds practical, thanks!

cmd/tui/autocomplete.go
kadinsayani: Sounds good! We can exit on `errors.Is(err, tea.ErrInterrupted)`.

roosterfish (approved):
Thanks!

PR #993, opened 2025-09-15
Demos: use disk-by-id paths for storage devices in the Terraform demo

mseralessandri:
Fixes https://github.com/canonical/microcloud/pull/982#issuecomment-3285120879


roosterfish (approved):
Thanks, but there still seems to be the error with `systemctl is-system-running --wait" failed with an error (1):
│ <nil>`. Is it maybe timing out?

roosterfish:
Another one, maybe you can find something which is common between those two https://github.com/canonical/microcloud/actions/runs/17734616809/job/50393485034?pr=988

PR #994, opened 2025-09-16
 Demos: removed the VM index from the disk-by-id paths for storage devices in the Terraform demo

mseralessandri:
See https://github.com/canonical/microcloud/pull/993#issuecomment-3296136937

### DIFF
--- a/test/suites/terraform_demo.sh
+++ b/test/suites/terraform_demo.sh
@@ -10,8 +10,9 @@ CEPH_EXCLUDED_VM="micro4"
 OVN_INTERFACE="enp6s0"
 OVN_IPV4_GATEWAY="192.0.2.1/24"
 OVN_IPV4_RANGE="192.0.2.100-192.0.2.254"
-LOCAL_DISKS="local1,local2,local3,local4"
-CEPH_DISKS="remote1,remote2,remote3"
+LOCAL_DISKS_PREFIX="local"
+CEPH_DISKS_PREFIX="remote"
+DNS_SERVERS="192.0.2.1,2001:db8:d:200::1"
 
 terraform_demo_cleanup() {
   echo "==> Terraform demo cleanup starting"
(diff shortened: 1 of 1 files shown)


demos/terraform/main.tf
```diff
@@ -160,8 +160,8 @@ resource "lxd_instance" "microcloud" {
       ovn_ipv4_range        = var.ovn_ipv4_range
       ovn_ipv6_gateway      = var.ovn_ipv6_gateway
       ovn_dns_servers       = var.ovn_dns_servers
-      local_disk_device     = "${var.disk_device_base}${lxd_volume.local_disk[count.index].name}"
-      ceph_disk_device      = local.systems[count.index].has_ceph ? "${var.disk_device_base}${lxd_volume.ceph_disk[local.ceph_disk_mapping[count.index]].name}" : ""
+      local_disk_device     = "${var.disk_device_base}${var.local_disk_name_prefix}"
```
roosterfish, line 163: The disks appear as `/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_local1` in the system. At which point do you add the index 1-3? 

In the template it seems to take the `local_disk_device` var right away.

mseralessandri: I am not sure I understood correctly, this seems fine to me 
https://github.com/canonical/microcloud/actions/runs/17759986531/job/50470518497?pr=994#step:3:1531
The device names are the following for all the VMs
```
ceph_disk  = "/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_remote"
local_disk = "/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_local"
```
and the lxd volume names are `local1`, `remote1` etc.

roosterfish: But should the disks appear as `/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_remote(1|2|3)` inside the VM?

roosterfish: Or `_local(1|2|3)`?

roosterfish: Ah nvm, the disk's name is defined as `name = "local"` or `name = "remote"`. I got confused by the indexes.

roosterfish: But then what I don't understand is this line https://github.com/canonical/microcloud/blob/main/test/suites/terraform_demo.sh#L114C58-L114C68.

In there we pass the disks `CEPH_DISKS="remote1,remote2,remote3"` but aren't they actually named just `remote`?

roosterfish (approved):

PR #1014, opened 2025-10-07
Test: print the relevant log files to stdout in case of failure in the Terraform demo test

mseralessandri:


test/main.sh
```diff
@@ -237,6 +237,12 @@ testbed_setup() {
 }
 
 collect_go_cover_files() {
+	# Skip coverage collection for terraform_demo test as it cleans up VMs before collection
```
roosterfish: I have the feeling we should not check for specific test case names inside those generic functions.
Instead what about adding another input var to the `system-test` action (same as we did for e.g. `setup_testbed`) to indicate whether or not we want to collect coverage files for this test run? 
In case it's set to `false`, we can unset `GOCOVERDIR` so it doesn't try to collect something.

test/suites/terraform_demo.sh
```diff
@@ -14,9 +15,68 @@ LOCAL_DISKS_PREFIX="local"
...
+  echo "==> Printing diagnostic logs to stdout"
+  
+  # Try initiator first, then fallback to any other VM
+  local source_vm=""
+  if lxc info "${INITIATOR_VM}" >/dev/null 2>&1; then
+    source_vm="${INITIATOR_VM}"
+    echo "Using initiator VM ${INITIATOR_VM} for logs"
+  else
+    echo "Initiator VM ${INITIATOR_VM} not accessible, trying fallback VMs"
+    for vm in ${VMS}; do
```
roosterfish: Can you just do this in the single loop without the need to define `INITIATOR_VM` as its anyway the first one in the list of `VMS`.
And then inside the loop (as you already do) break after the first successful attempt to collect logs.

But why only collecting the logs from the first machine? It might be helpful to also get the logs from the other machines too. Let's assume we are trying to join `micro2` to `micro1` but there is an error on `micro2`. It's not guaranteed that `micro2` returns every detail back to `micro1` so we would need to check in the logs of `micro2` to investigate this further and to maybe find more information about what has happened on this VM.

We do the same for the other tests. In case of failure loop through micro01-micro04 and get all the logs.

test/suites/terraform_demo.sh
```diff
@@ -14,9 +15,68 @@ LOCAL_DISKS_PREFIX="local"
...
+      if lxc info "${vm}" >/dev/null 2>&1; then
+        source_vm="${vm}"
+        echo "Using fallback VM ${source_vm} for logs"
+        break
+      fi
+    done
+  fi
+
+  if [ -z "${source_vm}" ]; then
```
roosterfish: This check should be fine to not try collecting logs if everything is offline for a reason.

test/suites/terraform_demo.sh
```diff
@@ -14,9 +15,68 @@ LOCAL_DISKS_PREFIX="local"
...
+  echo "==> End of logs"
+}
+
+
 terraform_demo_cleanup() {
   echo "==> Terraform demo cleanup starting"
-  
+
+  # Print logs to stdout immediately for debugging on failure
+  print_logs
```
roosterfish: Can we run this only in case the test failed with an exit code != 0?

test/main.sh
roosterfish: Thanks, your solution looks excellent.

Issue #958, opened 2025-09-04
Switch MicroCloud v3 to Microcluster v3


roosterfish:
MicroCloud v3 should pull Microcluster v3 to catch issues early on.

This should not involve more than switching all of the import statements. The untangling of deps was already done months ago so we can independently update the version of Microcluster even though MicroOVN and MicroCeph are still using Microcluster v2.

Also make sure to pull dqlite from its latest `main` branch (in Makefile). This change also has to be performed in the snap packaging repo inside the `snapcraft.yaml` file.

PR #1018, opened 2025-10-13
Go: Use Microcluster v3

roosterfish:
Fixes https://github.com/canonical/microcloud/issues/958.

The latest version of MicroCloud v3 should use Microcluster v3.
This also bumps to the latest version of LXD and contains some updates for the latest state of this dependency.


service/lxd.go
```diff
@@ -11,11 +11,11 @@ import (
 	"slices"
 	"time"
 
-	"github.com/canonical/lxd/client"
+	lxd "github.com/canonical/lxd/client"
```
simondeziel: Was that bit intentional?

Makefile
```diff
@@ -2,7 +2,7 @@ GOMIN=1.24.8
 GOCOVERDIR ?= $(shell go env GOCOVERDIR)
 GOPATH ?= $(shell go env GOPATH)
 DQLITE_PATH=$(GOPATH)/deps/dqlite
-DQLITE_BRANCH=lts-1.17.x
+DQLITE_BRANCH=main
```
simondeziel, line 5: Was there a specific reason for this change? The LXD `stable-5.21` and the corresponding snaps both use `lts-1.17.x`.

roosterfish: It's the edge version of MicroCloud. With this PR it uses the edge version of Microcluster, which uses latest `go-dqlite` so I thought it makes sense to also use latest version of dqlite.

simondeziel: Ack, thanks.

service/lxd.go
roosterfish: No, thanks, fixed!

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -8,7 +8,7 @@ import (
 	lxd "github.com/canonical/lxd/client"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
-	"github.com/canonical/microcluster/v2/state"
+	"github.com/canonical/microcluster/v3/state"
```
simondeziel, line 11: Line 8 of this file is the only other occurrence of this `lxd` prefix for imports.

roosterfish: https://github.com/canonical/microcloud/pull/1031

PR #1029, opened 2025-10-13
demos/terraform: remove package_upgrade from cloud-init

mseralessandri:
Remove `package_upgrade: true` from cloud-init to prevent recent systemd update from triggering `lxd-agent` restart that kills Terraform exec connections.

### DIFF
--- a/demos/terraform/cloud-init.yaml.tpl
+++ b/demos/terraform/cloud-init.yaml.tpl
@@ -1,10 +1,4 @@
 #cloud-config
-hostname: ${hostname}
-package_update: true
-package_upgrade: true
-
-packages:
-  - snapd
 
 write_files:
   - path: /etc/netplan/99-microcloud.yaml
@@ -21,6 +15,13 @@ write_files:
 runcmd:
   - |
     /usr/sbin/netplan apply && \
+    # Workaround for https://bugs.launchpad.net/snapd/+bug/2104066
+    (
+      /usr/bin/mkdir -p /etc/systemd/system/snapd.service.d && \
+      /usr/bin/printf '%s\n%s\n%s\n' '# Workaround for https://bugs.launchpad.net/snapd/+bug/2104066' '[Service]' 'Environment=SNAPD_STANDBY_WAIT=1m' > /etc/systemd/system/snapd.service.d/override.conf && \
+      /usr/bin/systemctl daemon-reload && \
+      /usr/bin/systemctl try-restart snapd.service
+    ) || echo "snapd workaround failed, continuing with snap installations" && \
     /usr/bin/snap install lxd --channel=${lxd_channel} --cohort="+" && \
     /usr/bin/snap install microceph --channel=${microceph_channel} --cohort="+" && \
     /usr/bin/snap install microovn --channel=${microovn_channel} --cohort="+" && \


demos/terraform/cloud-init.yaml.tpl
```diff
@@ -1,7 +1,7 @@
 #cloud-config
 hostname: ${hostname}
```
roosterfish: Based on the [comment](https://github.com/canonical/microcloud/pull/1029#pullrequestreview-3332639251) from Simon, do we need the `hostname` and `packages` sections? I guess the hostname is already the right one from the instance creation and possibly snapd is already installed on the image?

mseralessandri:
> > Thanks Simon! I can't reproduce the issue locally though, I should have the same error. I'll check the workaround, but it would make the cloud-init config quite complex, so I'd likely reintroduce `packages: - snapd`
> 
> The `snapd` package is already part of `-minimal` images. Maybe `cloud-init` was upgrading `snapd` all along and made the workaround unneeded?

The test now prints also the snap lists in the logs. It seems that `snap` version is already `2.71`. 
```
+ lxc exec micro1 -- snap list
Name        Version                 Rev    Tracking       Publisher    Notes
core22      20250822                2133   latest/stable  canonical**  base
core24      20250829                1196   latest/stable  canonical**  base
lxd         5.21.4-7b6bd68          36153  5.21/stable    canonical**  in-cohort
microceph   19.2.1+snapac85f90153   1554   squid/stable   canonical**  in-cohort
microcloud  2.1.1-202e275           1781   2/stable       canonical**  in-cohort
microovn    24.03.2+snapa2c59c105b  667    24.03/stable   canonical**  in-cohort
snapd       2.71                    25202  latest/stable  canonical**  snapd
```

simondeziel:
> > > Thanks Simon! I can't reproduce the issue locally though, I should have the same error. I'll check the workaround, but it would make the cloud-init config quite complex, so I'd likely reintroduce `packages: - snapd`
> > 
> > 
> > The `snapd` package is already part of `-minimal` images. Maybe `cloud-init` was upgrading `snapd` all along and made the workaround unneeded?
> 
> The test now prints also the snap lists in the logs. It seems that `snap` version is already `2.71`.

Yeah but the situation with `snapd` is a bit complicated. There is the `.deb` package and also the `snapd` snap that is pulled on demand when you first install a snap.

I kinda lost track of what's the issue at hand so maybe we could have a look at it together in a quick Meet?

mseralessandri:
@simondeziel it seems that the workaround is needed, although the difference between the local and remote environment is not the image (the fingerprint is the same)
```
lxc image list
+-------+--------------+--------+---------------------------------------------+--------------+-----------------+-----------+-----------------------------+
| ALIAS | FINGERPRINT  | PUBLIC |                 DESCRIPTION                 | ARCHITECTURE |      TYPE       |   SIZE    |         UPLOAD DATE         |
+-------+--------------+--------+---------------------------------------------+--------------+-----------------+-----------+-----------------------------+
|       | f3aaae82ade3 | no     | ubuntu 24.04 LTS amd64 (release) (20251001) | x86_64       | VIRTUAL-MACHINE | 591.94MiB | Oct 2, 2025 at 1:48pm (UTC) |
+-------+--------------+--------+---------------------------------------------+--------------+-----------------+-----------+-----------------------------+
```

PR #1034, opened 2025-10-20
Add: Prevent skipping distributed networking

roosterfish:
In case the existing cluster already uses distributed networking, new members cannot skip the configuration (e.g. interface selection) as it's required.
Therefore this PR ensures we automatically go into configuration mode for distributed networking in case it's required by the existing cluster when adding new members.

This also prevents MicroCloud setting up FAN networking across the cluster if any of the new members would skip distributed networking setup (this isn't possible anymore with the proposed changes).

The PR also modifies the test suite to allow accommodating both explicit and implicit (add member) distributed networking setups.

### DIFF
--- a/test/suites/add.sh
+++ b/test/suites/add.sh
@@ -35,7 +35,7 @@ test_add_interactive() {
   export CEPH_CLUSTER_NETWORK="${ceph_cluster_subnet_prefix}.0/24"
   export CEPH_PUBLIC_NETWORK="${ceph_public_subnet_prefix}.0/24"
   export CEPH_ENCRYPT="no"
-  export SETUP_OVN="yes"
+  export SETUP_OVN_EXPLICIT="yes"
   export OVN_FILTER="enp6s0"
   export IPV4_SUBNET="10.1.123.1/24"
   export IPV4_START="10.1.123.100"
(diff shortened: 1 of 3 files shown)


cmd/microcloud/ask.go
```diff
@@ -1077,10 +1077,22 @@ func (c *initConfig) askOVNNetwork(sh *service.Handler) error {
...
 
-	// Ask the user if they want OVN.
-	wantsOVN, err := c.asker.AskBool("Configure distributed networking?", true)
-	if err != nil {
-		return err
+	wantsOVN := false
+
+	// When adding a new member it's important to check whether or not the existing cluster already
+	// has distributed networking configured.
+	// If a user opts out of the configuration (e.g. insterface selection) for distributed networking, this can be considered an error.
```
crishoj:
```suggestion
	// If a user opts out of the configuration (e.g. interface selection) for distributed networking, this can be considered an error.
```

This sounds like adding an uplink-less member to a cluster with distributed networking is unsupported.
Could there be a case for supporting this?

roosterfish: I have dropped one line of the comment because with the proposed changes MicroCloud enters an implicit distributed networking setup if enforced by the existing cluster. It doesn't anymore ask to configure the network and jumps right into the follow up questions like interface selection.
Currently for MicroCloud we expect all members to have the uplink network. 

Feel free to raise a feature request in https://github.com/canonical/microcloud/issues.

crishoj: Thanks. Feature request posted: #1037

kadinsayani (approved):

PR #1042, opened 2025-10-23
github: Remove redundant 'directory' statement

roosterfish:
Follow up on https://github.com/canonical/microcloud/pull/1041.

We already have a 'directories' statement so also having 'directory' is wrong. See https://github.com/canonical/microcloud/runs/53495741971


roosterfish:
@simondeziel I have accidentally missed removing the `directory` statement in the initial PR :/. Just a follow up to fix it.

simondeziel (approved):

PR #1068, opened 2025-10-30
GitHub: Stabilize builds and use Canonical runners

roosterfish:
Switches to the Canonical provided runners and removes the weekly job that runs the instances suite on Canonical runners as now all tests are about to be executed in this environment.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -217,34 +217,6 @@ jobs:
       - name: System test
         uses: ./.github/actions/system-test
 
-  system-tests-canonical:
-    env:
-      DEBUG: "1"
-      SKIP_VM_LAUNCH: "1"
-      SNAPSHOT_RESTORE: "1"
-    # Run the tests on the Canonical runners only when scheduled.
-    if: ${{ github.event_name == 'schedule' && github.repository == 'canonical/microcloud' }}
-    name: System (Canonical)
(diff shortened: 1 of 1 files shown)


.github/actions/system-test/action.yml
```diff
@@ -150,9 +150,11 @@ runs:
 
     - name: Setup Terraform
       if: ${{ inputs.setup_terraform == 'true' }}
-      uses: hashicorp/setup-terraform@b9cd54a3c349d3f38e8881555d616ced269862dd # v3.1.2
-      with:
-        terraform_wrapper: false
+      shell: bash
```
simondeziel: Just a comment for later but in LXD we specify `bash` only once at the top of the job:

```
defaults:
  run:
    # Make sure bash is always invoked with `-eo pipefail`
    # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell
    shell: bash
```

roosterfish: Good point, removed this obsolete line as the same is already put in place for the workflow here https://github.com/canonical/microcloud/blob/main/.github/workflows/tests.yml#L26.

roosterfish: https://github.com/canonical/microcloud/pull/1068#issuecomment-3497366793

Issue #695, opened 2025-03-11
Unexport `Command` funcs


roosterfish:
See https://github.com/canonical/microcloud/pull/681#discussion_r1987940735.

PR #1070, opened 2025-11-02
Unexport Command & Run funcs

oversampling:
https://github.com/canonical/microcloud/issues/695

I believe that you don't want to export the command function, so I make some change `Command` -> `command`.

### DIFF
--- a/cmd/microcloudd/main.go
+++ b/cmd/microcloudd/main.go
@@ -60,13 +60,13 @@ func (c *cmdDaemon) command() *cobra.Command {
 		Version: version.Version(),
 	}
 
-	cmd.RunE = c.Run
+	cmd.RunE = c.run
 
 	return cmd
 }
 
-// Run runs the main microcloudd command.
-func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
+// run runs the main microcloudd command.
+func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 	if len(args) != 0 {
 		return cmd.Help()
 	}


cmd/microcloud/add.go
```diff
@@ -26,7 +26,7 @@ type cmdAdd struct {
 	flagSessionTimeout int64
 }
 
-// Command returns the subcommand to add new systems to MicroCloud.
+// command returns the subcommand to add new systems to MicroCloud.
```
roosterfish, line 29: Please squash this commit into the previous commits that actually modify the function's name.

roosterfish: We don't squash commits when merging the PR.

oversampling: Ok, fixed

cmd/microcloud/main_init.go
roosterfish, line 145: Can you please also unexport all of the `Run*` funcs? Just saw those whilst checking for any leftovers.

roosterfish (changes requested):

cmd/microcloud/main_init.go
oversampling: @roosterfish Yep. I also saw there is `RunInteractive` function which only called inside cmdInit, I believe it also need to be unexported. see this commit [commit](https://github.com/canonical/microcloud/pull/1070/commits/d0c910989320d8d8ad2dcc4fce684320fda5f759)

cmd/microcloud/add.go
```diff
@@ -31,16 +31,16 @@ func (c *cmdAdd) command() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "add",
 		Short: "Add new systems to an existing MicroCloud cluster",
-		RunE:  c.Run,
+		RunE:  c.run,
```
roosterfish, line 34: Thanks! 

However the commit message says `cmd/microcloud` but is has changed files from `cmd/microcloudd` package.

oversampling: Sorry, I fixed it, it should be fine now.

Issue #888, opened 2025-07-24
Rewrite LXD's service `WaitReady` to use `timeout` query parameter


roosterfish:
In the `service` package on the LXD service we have a `WaitReady` function which currently accepts a LXD client and a context.

Ultimately this function should receive only a context and timeout so that the overall wait procedure is governed by the context but we are waiting on the server using the `timeout` flag to get notified why the timeout expired and which resource still isn't ready.
Therefore we cannot only rely on the context because in case our context expires, we close the connection without waiting for a response from the server.
On the same page we cannot derive the timeout from the context's deadline as this would cause non deterministic behavior. Sometimes our context might expire and sometimes the server's `timeout` might be faster.

After https://github.com/canonical/lxd/pull/15495 made its way into `5.21/stable` we can solely rely on passing the `timeout` query parameter.

PR #1075, opened 2025-11-05
Wait ready timeout

oversampling:
https://github.com/canonical/microcloud/issues/888

Regarding this issue. I referenced this implementation https://github.com/canonical/lxd/blob/d0b5449e16905a55687ea8bd8ad7b8a912c183da/lxd/main_waitready.go#L95

If the api response with http.StatusServiceUnavailable, I added the error msg to `errLast` var.
What do you think, should I do something like this.

### DIFF
--- a/cmd/microcloudd/main.go
+++ b/cmd/microcloudd/main.go
@@ -29,6 +29,9 @@ import (
 // heartbeat to 0.
 const MinimumHeartbeatInterval = time.Millisecond * 200
 
+// LXDInitializationTimeout is the time limit for LXD initialization for microcloud daemon.
+const LXDInitializationTimeout time.Duration = 1 * time.Minute
+
 // Debug indicates whether to log debug messages or not.
 var Debug bool
 
@@ -201,7 +204,12 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 					return nil
 				}
 
-				initialized, err := s.Services[types.LXD].IsInitialized(context.Background())
+				// Create initialization context with timeout defined in LXDInitializationTimeout.
+				initializationCtx, cancel := context.WithTimeout(ctx, LXDInitializationTimeout)
+				defer cancel()
+
+				// Check if LXD is initialized using initializationCtx.
+				initialized, err := s.Services[types.LXD].IsInitialized(initializationCtx)
 				if err != nil {
 					return err
 				}


service/lxd.go
```diff
@@ -724,9 +724,6 @@ func (s *LXDService) WaitReady(ctx context.Context, c lxd.InstanceServer, networ
 	ctx, cancel := context.WithCancelCause(timeoutCtx)
 	defer cancel(context.Canceled)
```
roosterfish: Please make sure your commits are prefixed with the respective package name.

service/lxd.go
```diff
@@ -739,14 +736,13 @@ func (s *LXDService) WaitReady(ctx context.Context, c lxd.InstanceServer, networ
...
 			_, _, err := c.RawQuery("GET", url.String(), nil, "")
 			if err != nil {
-				if api.StatusErrorCheck(err, http.StatusNotFound) {
-					if unixCtx.Err() != nil {
-						cancel(err)
-
-						return
-					}
+				// LXD is reachable but is internally reporting as not ready after the specified timeout.
+				if api.StatusErrorCheck(err, http.StatusServiceUnavailable) {
```
roosterfish: This changes the original behavior.

In case the API returns `https.StatusServiceUnavailable` after 5s (when the timeout expires), the Go routine is returned so you won't be able to wait for the entire minute which is set as deadline at the beginning of the func.

Instead why don't you `continue` here instead of `return`? That would also be equivalent to what we have in LXD https://github.com/canonical/lxd/blob/d0b5449e16905a55687ea8bd8ad7b8a912c183da/lxd/main_waitready.go#L115.

I am ok getting rid of the extra 5s check for the availability of the socket but then please make sure it's also not anymore listed in the docstring of the `WaitReady` function.

service/lxd.go
```diff
@@ -9,9 +9,10 @@ import (
 	"net/http"
 	"net/url"
 	"slices"
+	"strconv"
 	"time"
 
-	"github.com/canonical/lxd/client"
+	lxd "github.com/canonical/lxd/client"
```
roosterfish: This change is not required.

service/lxd.go
```diff
@@ -736,12 +739,13 @@ func (s *LXDService) WaitReady(ctx context.Context, c lxd.InstanceServer, networ
 				url.WithQuery("storage", "1")
 			}
 
-			url.WithQuery("timeout", "5")
+			url.WithQuery("timeout", strconv.Itoa(apiTimeoutSeconds))
```
roosterfish: Can we make the timeout a function parameter in `WaitReady`?

roosterfish (changes requested):
Thanks for taking a look at this. Please have a look at my comments.

service/lxd.go
oversampling: Hmm, not sure why everytime I save the code it automatically add this one. Noted, I will fix this.

service/lxd.go
oversampling: I added the "apiTimeoutSeconds" as `WaitReady` function parameter with type of int.

service/lxd.go
oversampling: I am also thinking to sleep another 500ms before the continue.

service/lxd.go
roosterfish: Whilst going over this again I realized we don't need to have an extra "timeout" parameter.
As we already pass the context into `lxdService.Client(ctx)` (it's `lxd.ConnectLXDUnixWithContext(ctx, ...) under the hood) when initializing the LXD client, we already have a possibility to set a deadline which ultimately also affects the call to `c.RawQuery` inside the `WaitReady` function. 
So there is an upper limit in place. 

Still passing the context into `WaitReady` allows us to ensure we can retry in case the socket isn't yet present. Because if we call `c.RawQuery` without the socket being present, we will get `Failed to connect to local LXD: Get "http://unix.socket/1.0": dial unix /var/snap/lxd/common/lxd/unix.socket: connect: connection refused`. 
In this case, if the context hasn't yet expired, we can retry. But only until the context expires which will either cause an ongoing `c.RawQuery` to exit early or us not retrying another time. Something similar to this might work:

```go
func (s *LXDService) WaitReady(ctx context.Context, c lxd.InstanceServer, network bool, storage bool) error {
	url := api.NewURL().Path("internal", "ready")
	if network {
		url.WithQuery("network", "1")
	}

	if storage {
		url.WithQuery("storage", "1")
	}

	for {
		_, _, err := c.RawQuery("GET", url.String(), nil, "")
		if err != nil {
			if ctx.Err() == nil {
				time.Sleep(500 * time.Millisecond)
				continue
			}

			return fmt.Errorf("Failed waiting for LXD to start: %w", err)
		}

		break
	}

	return nil
}
```

Also ensure that all callers of `WaitReady` actually set a deadline on the context to prevent the wait from running too long.

service/lxd.go
roosterfish: Yep please see https://github.com/canonical/microcloud/pull/1075/files#r2522660895, we should for sure give some time in the middle before retrying.

roosterfish (changes requested):

service/lxd.go
oversampling: `waitready` is actually used in cobra [command](https://github.com/canonical/microcloud/blob/2012dafb83a6d2f6c79cb94ee46c7a85a5ecbbf8/cmd/microcloud/waitready.go#L74) and [isInitialize](https://github.com/canonical/microcloud/blob/2012dafb83a6d2f6c79cb94ee46c7a85a5ecbbf8/service/lxd.go#L673) func.

In cobra command, if user didn't specify the timeout flag it will not set the context timeout, so in this case I set it to 60 sec timeout by default if user didn't specify the flag. For this part, the ctx is actually shared in `isInitialized` and `waitReady`, however I also default set the timeout to 60 sec, not sure if it neccessary to refresh the timeout after `isInitialized` func, [reference](https://github.com/canonical/microcloud/blob/88e22683086aad417f8186e144e9fec1c2d760e6/cmd/microcloud/waitready.go#L76). 

In isInitialize func, I just set the parent context to have context timeout of 60 sec.

cmd/microcloud/ask.go
```diff
@@ -56,7 +56,12 @@ func checkInitialized(stateDir string, expectInitialized bool, preseed bool) err
 	}
 
 	return s.RunConcurrent("", "", func(s service.Service) error {
-		initialized, err := s.IsInitialized(context.Background())
+		// Create initialization context with 60 secs timeout.
+		initializationCtx, cancel := context.WithTimeout(context.Background(), time.Second*60)
```
roosterfish: Small nit, you can do `time.Minute` instead of `time.Second*60`.

cmd/microcloudd/main.go
```diff
@@ -201,7 +201,12 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 					return nil
 				}
 
-				initialized, err := s.Services[types.LXD].IsInitialized(context.Background())
+				// Create initialization context with 60 secs timeout.
+				initializationCtx, cancel := context.WithTimeout(context.Background(), time.Second*60)
```
roosterfish: Instead of `context.Background()` rather use the `ctx` passed to the `OnStart` func.

cmd/microcloud/ask.go
oversampling: @roosterfish Actually I just realize there is `main_init.go` where inside defined various timeout for microcloud package, do you think I should create an variable `const DefaultInitializationTimeout time.Duration = 1 * time.Minute` and reuse it only in microcloud package

roosterfish: Good idea, but let's call it `LXDInitializationTimeout`. The `default` suggests it can be modified but in this case it's a fixed value.

cmd/microcloudd/main.go
```diff
@@ -201,7 +201,12 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 					return nil
 				}
 
-				initialized, err := s.Services[types.LXD].IsInitialized(context.Background())
+				// Create initialization context with 1 minute timeout.
+				initializationCtx, cancel := context.WithTimeout(ctx, time.Minute)
```
roosterfish: Can you please also use the `LXDInitializationTimeout` const here?

oversampling: @roosterfish Actually the `LXDInitializationTimeout` is in `cmd/microcloud`, is microcloud program, I don't think it should import the variable in `cmd/microcloudd` program. I think is better I defined another `LXDInitializationTimeout` in `cmd/microcloudd/main.go`. Correct me if I am wrong. 

Maybe there is another possible way where I define `lxdInitializationTimeout` in `NewLXDService`, then in IsInitialized method
```
func NewLXDService(name string, addr string, cloudDir string) (*LXDService, error) {
	client, err := microcluster.App(microcluster.Args{StateDir: cloudDir})
	if err != nil {
		return nil, err
	}

	return &LXDService{
		...
		lxdInitializationTimeout: 1 * time.Minute,
	}, nil
}

func (s LXDService) IsInitialized(ctx context.Context) (bool, error) {
	// If context deadline is less than initialization timeout, use initialization timeout.
	if deadline, ok := ctx.Deadline(); ok && deadline.Sub(time.Now()) < s.InitializationTimeout() {
                // Use initializationTimeout create in NewLXDService
		ctx, cancel := context.WithTimeout(ctx, s.lxdInitializationTimeout)
		defer cancel()
	}
        ...
	err = s.WaitReady(ctx, c, false, false)
        ...

	return isInit, nil
}
```

cmd/microcloudd/main.go
```diff
@@ -201,7 +204,12 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 					return nil
 				}
 
-				initialized, err := s.Services[types.LXD].IsInitialized(context.Background())
+				// Create initialization context with 1 minute timeout.
```
roosterfish: Just a nit, if we ever change the value in the `const`, this comment will be false.

cmd/microcloudd/main.go
roosterfish: I am fine in having another const in the daemon package.

PR #1087, opened 2025-11-19
Fixup e2e tests

MggMuggins:
Minor e2e test tweaks for Ubuntu Pro handovers.

### DIFF
--- a/doc/doc-cheat-sheet-myst.md
+++ b/doc/doc-cheat-sheet-myst.md
@@ -19,7 +19,7 @@ myst:
 This file contains the syntax for commonly used Markdown and MyST markup.
 Open it in your text editor to quickly copy and paste the markup you need.
 
-See the [MyST syntax guide](https://canonical-starter-pack.readthedocs-hosted.com/latest/reference/myst-syntax-reference/) for detailed information and conventions.
+See the [MyST syntax guide](https://canonical-starter-pack.readthedocs-hosted.com/stable/reference/myst-syntax-reference/) for detailed information and conventions.
 
 Also see the [MyST documentation](https://myst-parser.readthedocs.io/en/latest/index.html) for detailed information on MyST, and the [Canonical Documentation Style Guide](https://docs.ubuntu.com/styleguide/en) for general style conventions.
 


simondeziel (dismissed):
LGTM, thanks!

roosterfish (approved):
Looks good!

e2e job was failing due to snap download error `  - Fetch and check assertions for snap "snapd" (25577) (cannot fetch assertion: got unexpected HTTP status code 408 via GET to "https://api.snapcraft.io/v2/assertions/snap-revision/zN4jkDG1_fiE8kC094y-9eQaBF4upRyJt7tV-or1F3YjJsXQS-jS_Xe577vW1Sai?max-format=0")`. Just kicked it off again.

test/e2e/run
```diff
@@ -58,6 +58,15 @@ init() {
     fi
 }
 
+info() {
+    set -x
+    microcloud status
```
roosterfish: It fails with `+ microcloud status
./run: line 63: microcloud: command not found`, see https://github.com/canonical/microcloud/actions/runs/19513589460/job/55911380859?pr=1087.

I suspect you have to enter the `micro*` machines to run the commands.

roosterfish (changes requested):

test/e2e/run
MggMuggins: I did a poor job of reading the manual here: `./run` is expected to be run with a lxc remote configured, not necessarily on one of the metal MicroCloud cluster members. I've just dropped these info commands for now.

test/e2e/run
```diff
@@ -58,6 +58,12 @@ init() {
     fi
 }
 
+info() {
+    set -x
+    lxc cluster list
```
roosterfish: This will run `lxc cluster list` on the single node LXD (on the runner) which we use to provision the MicroCloud VMs which themselves host a LXD cluster (MicroCloud) that is used for the e2e tests.

The e2e test uses a `REMOTE` variable used to access the MicroCloud. So if you want to show the final output of `"${REMOTE}:"` after the evacuation tests, make sure to run `lxc cluster list "${REMOTE}:"`. 
There already is a `clusterMembers` func you can use as reference.

MggMuggins: Thanks for this - trying to do too many things at once. Sorry about the noise :)

roosterfish:
Kicked off the tests again as some failed during artifact download.

PR #1115, opened 2025-12-03
Start backports to unblock pipeline (v2-edge)

roosterfish:
Cleanly backport all the changes necessary to switch v2 to using Canonical runners which (once scheduled) produce almost deterministic results for the MicroCloud test suites as long there are no download issues during setup.

### DIFF
--- a/doc/tutorial/get_started.md
+++ b/doc/tutorial/get_started.md
@@ -1079,9 +1079,4 @@ Now that your MicroCloud is up and running, you can start using it!
 
 If you're already familiar with LXD, see {ref}`howto-commands` for a reference of the most common commands.
 
-If you're new to LXD, check out the {ref}`LXD tutorials <lxd:tutorials>` to familiarize yourself with what you can do in LXD:
-
-- {ref}`lxd:tutorial-ui` guides you through common operations in LXD, using the UI.
-- {ref}`lxd:first-steps` goes through the same functionality, but using the CLI.
-
-In both tutorials, you can skip the first section about installing and initializing LXD, because LXD is already operational as part of your MicroCloud setup.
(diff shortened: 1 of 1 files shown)


roosterfish:
The `non_ha` test failed with `E: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/noble-updates/restricted/i18n/Translation-en.xz  File has unexpected size (525768 != 526172). Mirror sync in progress? [IP: 91.189.91.83 80]` which is a specific error that happens rather often with the Canonical runners during setup (download). Retried it.

simondeziel (approved):
LGTM, thanks

.github/workflows/commits.yml
```diff
@@ -11,6 +11,7 @@ permissions:
 
 jobs:
   commits:
+    if: ${{ github.repository == 'canonical/microcloud' }}
     name: Branch target and CLA
     runs-on: ubuntu-latest
```
kadinsayani, line 16: We started using the `ubuntu-slim` runner in LXD for triaging, CLA checks and branch target checks. [ubuntu-slim](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#single-cpu-runners) runners are lightweight 1vCPU runners with a 15min maximum execution time. They might work well here too!

kadinsayani: The motivation for using slim runners is less energy consumption and faster timeouts if anything hangs.

simondeziel: TIL, `ubuntu-slim` are in fact a container instances running on top of a VM. I like them even better now ;)

kadinsayani: Cool! I wonder which container runtime GitHub is using.

roosterfish: > They might work well here too

Sounds great, can you please add a small PR when you find time?

kadinsayani: https://github.com/canonical/microcloud/pull/1116

simondeziel: > Cool! I wonder which container runtime GitHub is using.

Docker: https://github.com/canonical/microcloud/actions/runs/19933937963/job/57153384632#step:1:11

PR #1116, opened 2025-12-04
github: Use ubuntu-slim runners

kadinsayani:
ubuntu-slim runners are lightweight 1vCPU with a 15min maximum execution time. They consume less energy and will timeout faster if anything hangs.


roosterfish (commented):
Thanks! Can we also run the `changes` job in `tests.yml` with the slim runner? It also just runs for a few seconds to check whether or not something else in the repo besides the `doc/*` folder was modified.

PR #1118, opened 2025-12-04
github: Use ubuntu-slim runners

kadinsayani:


simondeziel:
> > Excellent! We'll only know if it works when merged as we cannot trigger a manual run til then.
> 
> I took a look at the average run times for these jobs and I think the slim runner should work well, but we shall see :)

Indeed, we have enough headroom for them to be ~15x slower ... and other than hitting the time limit of 15 minutes, we don't care as there is no human waiting at the results.

PR #1131, opened 2025-12-08
doc: add and emphasize UI documentation

minaelee:
Adds a how-to guide for using and accessing the UI in MicroCloud, including links to UI documentation in LXD. Also links to and emphasizes the presence of a UI and its documentation from other parts of the MicroCloud docs, to make sure that the availability of a UI is visible to anyone skimming the documentation.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -1,4 +1,7 @@
 ---
+myst:
+  html_meta:
+    description: MicroCloud is a lightweight, open source cloud platform built on LXD, MicroCeph, and MicroOVN, ideal for private cloud, edge computing, and test labs.
 discourse: lxc:[Introducing&#32;MicroCloud](15871)
 relatedlinks: "[Install&#32;MicroCloud&#32;on&#32;Linux&#32;|&#32;Snap&#32;Store](https://snapcraft.io/microcloud)"
 ---
--- a/doc/explanation/microcloud.md
+++ b/doc/explanation/microcloud.md
@@ -1,7 +1,7 @@
 ---
 myst:
   html_meta:
-    description: 
+    description: An overview of MicroCloud including its component snaps of LXD, MicroCeph, and MicroOVN, and its major features.
 ---
 
 (exp-microcloud)=
--- a/doc/how-to/commands.md
+++ b/doc/how-to/commands.md
@@ -1,3 +1,9 @@
+---
+myst:
+  html_meta:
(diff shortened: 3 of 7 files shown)


doc/index.md
```diff
@@ -6,11 +6,11 @@ relatedlinks: "[Install&#32;MicroCloud&#32;on&#32;Linux&#32;|&#32;Snap&#32;Store
 (home)=
 # MicroCloud
 
-Deploy a scalable, low-touch cloud platform in minutes with MicroCloud.
+Deploy a low-touch, open source cloud platform in minutes with MicroCloud.
```
roosterfish, line 12: Should we also align the desc in https://snapcraft.io/microcloud? It's now managed through the `snapcraft.yml` file directly here https://github.com/canonical/microcloud-pkg-snap/blob/latest-edge/snapcraft.yaml#L10.

doc/index.md
```diff
@@ -33,7 +33,7 @@ MicroCloud is designed for small-scale private clouds and hybrid cloud extension
 
 ```{grid-item} [How-to guides](/how-to/index)
 
-**Step-by-step guides** covering key operations and common tasks such as {ref}`installing MicroCloud <howto-install>` and {ref}`adding <howto-member-add>` and {ref}`removing <howto-member-remove>` cluster members
+**Step-by-step guides** covering key operations and common tasks such as {ref}`installing MicroCloud <howto-install>`, {ref}`adding <howto-member-add>` and {ref}`removing <howto-member-remove>` cluster members, and {ref}`accessing the UI <howto-ui>`
```
roosterfish, line 39: Shouldn't there be also a comma in between `<howto-member-add>` and `<howto-member-remove>`?

doc/index.md
minaelee: Yes, let's. I can open a PR there.

doc/index.md
minaelee: No, because the "and" in "adding and removing cluster members" is used to join a compound modifier where both "adding" and "removing" modify "cluster members". A comma should not be used to separate a modifier from what it modifies, even if it's used in a series with other commas. The above renders like this:

>**Step-by-step guides** covering key operations and common tasks such as installing MicroCloud, adding and removing cluster members, and accessing the UI.

Adding a comma would look like this:
>**Step-by-step guides** covering key operations and common tasks such as installing MicroCloud, adding, and removing cluster members, and accessing the UI.

roosterfish: Ah nvm, probably I would have caught it when checking the rendered docs instead. Thanks!

PR #1135, opened 2025-12-10
Preseed: Fix unicast mode

roosterfish:
This is not affecting MicroCloud v2, only v3.

Whilst exploring https://chat.canonical.com/canonical/pl/wwq89ehmkj8bugya8drxqwa7ew I found that https://github.com/canonical/microcloud/pull/927 potentially broke preseed in unicast mode as it always requires the preseed's `LookupSubnet` to be set even though this is only required in multicast mode.

The PR refactors the code a little bit to and also fixes a warning to use the standardized way of logging warnings via the tui interface.

### DIFF
--- a/cmd/microcloud/preseed.go
+++ b/cmd/microcloud/preseed.go
@@ -1123,11 +1123,11 @@ func (p *Preseed) Parse(s *service.Handler, c *initConfig, installedServices map
 				return nil, err
 			}
 
-			if targetPublicCephNetwork.String() != c.lookupSubnet.String() {
+			if targetPublicCephNetwork != nil && targetPublicCephNetwork.String() != c.lookupSubnet.String() {
 				customTargetCephPublicNetwork = targetPublicCephNetwork.String()
 			}
 
-			if targetInternalCephNetwork.String() != c.lookupSubnet.String() {
+			if targetInternalCephNetwork != nil && targetInternalCephNetwork.String() != c.lookupSubnet.String() {
 				customTargetCephInternalNetwork = targetInternalCephNetwork.String()
 			}
 		}
@@ -1199,14 +1199,14 @@ func (p *Preseed) Parse(s *service.Handler, c *initConfig, installedServices map
 			return nil, err
 		}
 
-		if localInternalCephNetwork.String() != "" && localInternalCephNetwork.String() != c.lookupSubnet.String() {
+		if localInternalCephNetwork != nil && localInternalCephNetwork.String() != "" && localInternalCephNetwork.String() != c.lookupSubnet.String() {
 			_, err = c.validateCephInterfacesForSubnet(lxd, addressedInterfaces, localInternalCephNetwork.String())
 			if err != nil {
 				return nil, fmt.Errorf("Failed to validate Ceph internal network: %w", err)
 			}
 		}
 
-		if localPublicCephNetwork.String() != "" && localPublicCephNetwork.String() != c.lookupSubnet.String() {
+		if localPublicCephNetwork != nil && localPublicCephNetwork.String() != "" && localPublicCephNetwork.String() != c.lookupSubnet.String() {
 			_, err = c.validateCephInterfacesForSubnet(lxd, addressedInterfaces, localPublicCephNetwork.String())
 			if err != nil {
 				return nil, fmt.Errorf("Failed to validate Ceph public network: %w", err)


cmd/microcloud/preseed.go
```diff
@@ -725,21 +725,9 @@ func (p *Preseed) Parse(s *service.Handler, c *initConfig, installedServices map
...
-		}
-
-		ipLookupSubnet, lookupSubnet, err := net.ParseCIDR(p.LookupSubnet)
-		if err != nil {
-			return nil, err
-		}
-
 		bootstrapSystem := c.systems[s.Name]
-		bootstrapSystem.MicroCloudInternalNetwork = &NetworkInterfaceInfo{Interface: *microCloudSubnetInterface, Subnet: lookupSubnet, IP: ipLookupSubnet}
+		bootstrapSystem.MicroCloudInternalNetwork = &NetworkInterfaceInfo{Interface: *c.lookupIface, Subnet: c.lookupSubnet, IP: c.lookupSubnet.IP}
```
kadinsayani, line 739: Since we're assuming these fields are already set, should we take a defensive approach here and ensure we don't try to dereference a nil pointer?

cmd/microcloud/preseed.go
```diff
@@ -614,50 +614,55 @@ func (d *DiskFilter) Match(disks []lxdAPI.ResourcesStorageDisk) ([]lxdAPI.Resour
...
+		if system.Name == s.Name {
+			continue
 		}
+
+		expectedSystems = append(expectedSystems, system.Name)
 	}
 
-	if c.lookupIface == nil {
+	var err error
+	c.lookupIface, c.lookupSubnet, err = p.findInterfaceAndNetworkForAddress(c.address)
```
kadinsayani, line 664: This may be problematic since there is no set of return values from `findInterfaceAndNetworkForAddress` where `c.lookupIface` can be populated while `c.lookupSubnet`, which is the scenario we want to accomodate for unicast, right?

cmd/microcloud/preseed.go
```diff
@@ -1095,7 +1090,7 @@ func (p *Preseed) Parse(s *service.Handler, c *initConfig, installedServices map
 		}
 
 		if osdHosts < RecommendedOSDHosts {
-			fmt.Printf("Warning: OSD host count is less than %d. Distributed storage is not fault-tolerant\n", RecommendedOSDHosts)
+			tui.PrintWarning(fmt.Sprintf("OSD host count is less than %d. Distributed storage is not fault-tolerant", RecommendedOSDHosts))
```
kadinsayani:
```suggestion
			tui.PrintWarning(fmt.Sprintf("OSD host count (%d) is less than %d (recommended). Distributed storage is not fault-tolerant", osdHosts, RecommendedOSDHosts))
```

kadinsayani (commented):
Thanks for fixing this and adding a test! I have a few questions and suggestions.

cmd/microcloud/preseed.go
roosterfish: The naming of those variables is not right which causes confusion I believe. 
At the end these two track MicroCloud's internal network subnet (+address) and interface on this system.

And the purpose of `findInterfaceAndNetworkForAddress` is to get the those information based on the address which was selected to be used for the internal network, e.g. `10.1.2.22`.

cmd/microcloud/preseed.go
roosterfish: Yes good point, I have added a check to verify both the interface and subnet. In this case we get the IP from the net definition so we don't have to specifically check this too.

cmd/microcloud/preseed.go
roosterfish: Updated this to use the same wording as when doing the interactive questionnaire.

roosterfish:
The proposed changes have raised another issue with the parsing/checking of the Ceph public/internal networks which is now fixed too as part of this PR.

cmd/microcloud/preseed.go
roosterfish: Let me fix the naming in a follow up PR. I would change the variable names on the struct in general.

kadinsayani: Thanks for clarifying, sounds good.

cmd/microcloud/preseed.go
kadinsayani: Awesome thanks!

PR #1140, opened 2025-12-12
Test: Retry apt and snap downloads

roosterfish:
Try to better deal with stability issues in the infrastructure when performing apt/snap downloads.

Also added a commit to drop the `check_snap_channels` func to reduce noise in the pipeline summary page.

### DIFF
--- a/test/includes/check.sh
+++ b/test/includes/check.sh
@@ -34,24 +34,3 @@ check_empty() {
 		false
 	fi
 }
-
-check_snap_channels() {
-    if [ "${LXD_SNAP_CHANNEL}" != "5.21/edge" ]; then
-        echo "::warning::lxd channel not set to 5.21/edge, continuing anyway"
-    fi
-
-	non_edge=""
-    if [ "${MICROCEPH_SNAP_CHANNEL}" != "latest/edge" ]; then
-        non_edge="${non_edge} microceph"
-    fi
-    if [ "${MICROCLOUD_SNAP_CHANNEL}" != "latest/edge" ]; then
-        non_edge="${non_edge} microcloud"
-    fi
-    if [ "${MICROOVN_SNAP_CHANNEL}" != "latest/edge" ]; then
-        non_edge="${non_edge} microovn"
-    fi
-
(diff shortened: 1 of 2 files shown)


test/includes/microcloud.sh
```diff
@@ -1258,15 +1280,16 @@ setup_system() {
...
+      retry lxc exec "${name}" -- apt-get install --no-install-recommends -y jq zfsutils-linux htop
     else
-      lxc exec "${name}" -- apt-get install --no-install-recommends -y jq
+      retry lxc exec "${name}" -- apt-get install --no-install-recommends -y jq
     fi
 
-    lxc exec "${name}" -- snap install snapd
-    lxc exec "${name}" -- snap install yq
+    retry lxc exec "${name}" -- snap install snapd
+    retry lxc exec "${name}" -- snap install yq
```
simondeziel, line 1292: For another PR, `yq` is best installed with `apt-get` on 24.04 as it's faster than pulling the snap.

For now, I think just `snap install yq` will also bring `snapd`

test/suites/basic.sh
```diff
@@ -723,7 +723,7 @@ test_service_mismatch() {
   reset_systems 1 3 1
 
   lxc exec micro01 -- sh -c "echo 1 > /proc/sys/net/ipv6/conf/enp5s0/disable_ipv6"
```
simondeziel, line 725: Unrelated but the `sh` wrapper could be avoided:
```suggestion
  lxc exec micro01 -- sysctl net.ipv6.conf.eth0.disable_ipv6=1
```

test/includes/microcloud.sh
roosterfish: Ah yes, I remember you providing this hint some time ago, most likely in the LXD CI repo when I was adding checks for the backup configs which use YAML.
Will fix in a follow up PR.

test/suites/basic.sh
roosterfish: Thanks, let me fix this in a follow up.

PR #1161, opened 2025-12-16
doc: backport fix to LXD configuration (v2-edge)

minaelee:
Backports fix in https://github.com/canonical/microcloud/pull/1160 to align with updates to docs configuration in the LXD repository.

### DIFF
--- a/doc/.sphinx/_integration/add_config.py
+++ b/doc/.sphinx/_integration/add_config.py
@@ -11,8 +11,8 @@ html_context['microovn_tag'] = "../microovn/_static/microovn.png"
 
 if project == "LXD":
     html_baseurl = "https://documentation.ubuntu.com/lxd/stable-5.21/"
-    custom_html_js_files.append('rtd-search.js')
-    custom_tags.append('integrated')
+    html_js_files.append('rtd-search.js')
+    tags.add('integrated')
 elif project == "MicroCeph":
     html_baseurl = "https://canonical-microceph.readthedocs-hosted.com/en/latest/"
     # Override default header templates
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -62,7 +62,7 @@ integrate:
 	cp .sphinx/_integration/rtd-search.js integration/microceph/docs/.sphinx/_static/
 	cp .sphinx/_integration/rtd-search.js integration/microovn/docs/.sphinx/_static/
 	# Add information about where to find the tags/docs to the doc sets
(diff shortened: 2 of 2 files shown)


doc/.sphinx/_integration/add_config.py
```diff
@@ -10,9 +10,9 @@
 html_context['microovn_tag'] = "../microovn/_static/microovn.png"
 
 if project == "LXD":
-    html_baseurl = "https://documentation.ubuntu.com/lxd/stable-5.21/"
-    custom_html_js_files.append('rtd-search.js')
-    custom_tags.append('integrated')
+    html_baseurl = "https://documentation.ubuntu.com/lxd/latest/"
```
roosterfish: I guess this should be `stable-5.21` for `v2-edge`?

minaelee: It should, and the one for MicroOVN should also be https://canonical-microovn.readthedocs-hosted.com/en/24.03/. Do you mind if I submit a separate PR for that since that's unrelated to the change in this one?

roosterfish: But before applying the backport it already says `stable-5.21`. So we should rather modify this commit instead of adding another one on top.

roosterfish: Why was it again that we cannot rely on the pulled repos for which we specify the right version in the `doc/Makefile` but have to instead reach out to the respective docs to fetch something?

> Do you mind if I submit a separate PR

Yes please for MicroOVN.

minaelee: > But before applying the backport it already says `stable-5.21`. So we should rather modify this commit instead of adding another one on top.

Ah, got it, that got overwritten when I was cherrypicking. Fixed.

minaelee: > Why was it again that we cannot rely on the pulled repos for which we specify the right version in the `doc/Makefile` but have to instead reach out to the respective docs to fetch something?
> 
> > Do you mind if I submit a separate PR
> 
> Yes please for MicroOVN.

The `html_baseurl`s within those repos rely on an RTD environmental variable so if we don't set it manually in add_config.py, it would be set to the MicroCloud base URL instead of the respective docs'.

roosterfish (approved):
Ty!

doc/.sphinx/_integration/add_config.py
roosterfish: Ok thanks.

PR #1163, opened 2025-12-17
cmd/microcloud/cluster_manager: send pool usages instead of aggregated disk usage

edlerd:
# Done
- in cluster manager heartbeats, send pool usages instead of aggregated disk usage
- relates to https://github.com/canonical/microcloud-cluster-manager/pull/358

fixes https://github.com/canonical/microcloud-cluster-manager/issues/343

### DIFF
--- a/cmd/microcloudd/cluster_manager_task.go
+++ b/cmd/microcloudd/cluster_manager_task.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"slices"
 	"time"
 
 	"github.com/canonical/lxd/client"
@@ -188,6 +189,11 @@ func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *types.Clus
 		result.UIURL = lxdMembers[0].URL
 	}
 
+	localPools, err := getLocalPools(lxdClient)
+	if err != nil {
+		return fmt.Errorf("Failed to get local LXD storage pools: %w", err)
+	}
+
 	var cpuLoad1 float64
 	var cpuLoad5 float64
 	var cpuLoad15 float64
@@ -210,10 +216,7 @@ func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *types.Clus
 		cpuLoad5 += memberState.SysInfo.LoadAverages[1]
 		cpuLoad15 += memberState.SysInfo.LoadAverages[2]
 
-		for _, poolsState := range memberState.StoragePools {
-			result.DiskTotalSize += int64(poolsState.Space.Total)
-			result.DiskUsage += int64(poolsState.Space.Used)
-		}
+		enrichStoragePoolMetrics(member, memberState, localPools, result)
 	}
 
 	for status, count := range statusFrequencies {
@@ -234,4 +237,75 @@ func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *types.Clus
 	}
 
 	return nil
+}
+
+func enrichStoragePoolMetrics(member api.ClusterMember, memberState *api.ClusterMemberState, localPools []string, result *types.ClusterManagerPostStatus) {
+	for name, poolState := range memberState.StoragePools {
+		if poolState.Space.Total == 0 || poolState.Space.Used == 0 {
+			// Error state or no available info on this pool.
+			logger.Info("Missing usage information from LXD storage pool", logger.Ctx{"member": member.ServerName, "pool": name})
+			continue
+		}
+
+		isLocalPool := slices.Contains(localPools, name)
+		if isLocalPool {
+			result.StoragePoolUsages = append(result.StoragePoolUsages, types.StoragePoolUsage{
+				Name:   name,
+				Member: member.ServerName,
+				Total:  poolState.Space.Total,
+				Usage:  poolState.Space.Used,
+			})
+
+			continue
+		}
+
+		if hasStoragePool(result.StoragePoolUsages, name) {
+			// We have already recorded this remote pool from another member.
+			continue
+		}
+
+		result.StoragePoolUsages = append(result.StoragePoolUsages, types.StoragePoolUsage{
+			Name:  name,
+			Total: poolState.Space.Total,
+			Usage: poolState.Space.Used,
+		})
+	}
+}
+
+func hasStoragePool(usages []types.StoragePoolUsage, name string) bool {
+	for _, u := range usages {
+		if u.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
+func getLocalPools(lxdClient lxd.InstanceServer) ([]string, error) {
+	server, _, err := lxdClient.GetServer()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get LXD server info: %w", err)
+	}
+
+	var localDrivers []string
+	for _, d := range server.Environment.StorageSupportedDrivers {
+		if !d.Remote {
+			localDrivers = append(localDrivers, d.Name)
+		}
+	}
+
+	storagePools, err := lxdClient.GetStoragePools()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get LXD storage pools: %w", err)
+	}
+
+	var localPools []string
+	for _, pool := range storagePools {
+		poolDriver := pool.Driver
+		if slices.Contains(localDrivers, poolDriver) {
+			localPools = append(localPools, pool.Name)
+		}
+	}
+
+	return localPools, nil
 }


api/types/cluster_manager.go
```diff
@@ -71,8 +78,7 @@ type ClusterManagerPostStatus struct {
 	CPULoad15         string               `json:"cpu_load_15"`
 	MemoryTotalAmount int64                `json:"memory_total_amount"`
 	MemoryUsage       int64                `json:"memory_usage"`
-	DiskTotalSize     int64                `json:"disk_total_size"`
-	DiskUsage         int64                `json:"disk_usage"`
+	StoragePoolUsages []StoragePoolUsage   `json:"storage_pool_usages"`
 	MemberStatuses    []StatusDistribution `json:"member_statuses"`
```
tomponline, line 83: Should storage pool usages be in `MemberStatus` as local pools will have different usage per-member?

edlerd: `MemberStatus` is a representation of the distribution of member status. It can not hold storage pool information as it is a high level aggregate:
```
type StatusDistribution struct {
	Status string `json:"status"`
	Count  int64  `json:"count"`
}
```

edlerd: I think this is the right structure, as we have also cluster wide pools, so pushing the pool usage into a cluster member hierarchy wouldn't allow representing such pools.

edlerd:
> Is this data submission to cluster manager on a per-member basis?

The heartbeat is sent for the whole cluster from one well-defined node. Not from multiple members.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -235,3 +236,27 @@ func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *types.Clus
...
+
+	for i := range storagePools {
+		pool := storagePools[i]
+		poolResources, err := lxdClient.GetStoragePoolResources(pool.Name)
+		if err != nil {
+			logger.Warn("Failed to get LXD storage pool usage", logger.Ctx{"pool": pool.Name, "err": err})
+			continue
+		}
+
+		result.StoragePoolUsages = append(result.StoragePoolUsages, types.StoragePoolUsage{
```
roosterfish, line 267: Adding to https://github.com/canonical/microcloud/pull/1163/changes/20ef69993820d9f41fbeca014a0c6d67af66ee77#r2627124790, we might have a general issue here.

Depending on which cluster member is the current leader, we might report different use/total stats. 
Imagine a MicroCloud with a ZFS pool `local`. Each cluster member contributes a disk to this pool but there is no guarantee that these disks have the same size (total) and neither do each of these cluster member local pools have the same volumes sitting on top of it (usage). 

This is of course different for remote pools which would report the same values on all cluster members.
So likely we should do something like in `enrichClusterMemberMetrics` which accounts for all cluster members.

roosterfish: Currently `StoragePoolUsages` tracks all storage pools visible from the cluster member who pushes the metrics (leader) but the output might vary heavily in case another node takes the leadership.

roosterfish: This is not an issue in the current version as we are looping over all members and just sump up the values from all of them.

edlerd: > Currently StoragePoolUsages tracks all storage pools visible from the cluster member who pushes the metrics (leader) but the output might vary heavily in case another node takes the leadership.

Yes, that is a problem, also we hide usage of local pools on any other cluster member than the one handling the heartbeats.

In the UI, we report per member usage for local pools. I think we should do the same here as well:
* one entry in`StoragePoolUsages` for cluster wide pools like cephfs driver
* multiple entries in `StoragePoolUsages` for local pools. There will be one entry per cluster member reporting the local usage

To keep all information in check, we add an optional Member field to `StoragePoolUsages`. The `Member` field is not used for cluster wide pools.

To give an example:
```
{
  {Name: "ceph-pool", Total: 1024, Usage: 512},
  {Name: "zfs-pool", Member: "node1", Total: 2048, Usage: 1024},
  {Name: "zfs-pool", Member: "node2", Total: 2048, Usage: 24},
  {Name: "zfs-pool", Member: "node3", Total: 2048, Usage: 8},
}
```

See the updated PR with these changes.

edlerd: > This is not an issue in the current version as we are looping over all members and just sump up the values from all of them.

I think there was an issue in the old version because we would count cluster wide pools multiple times.

roosterfish: > I think there was an issue in the old version because we would count cluster wide pools multiple times.

Yeah I see, good catch.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -235,3 +247,35 @@ func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *types.Clus
 
 	return nil
 }
+
+func enrichStoragePoolMetrics(member api.ClusterMember, memberState *api.ClusterMemberState, localPools []string, knownRemotePools []string, result *types.ClusterManagerPostStatus) {
+	for name, poolState := range memberState.StoragePools {
+		if poolState.Space.Total == 0 || poolState.Space.Used == 0 {
```
roosterfish, line 244: When is this the case? Could you please add a comment?

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -235,3 +247,35 @@ func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *types.Clus
 
 	return nil
 }
+
+func enrichStoragePoolMetrics(member api.ClusterMember, memberState *api.ClusterMemberState, localPools []string, knownRemotePools []string, result *types.ClusterManagerPostStatus) {
```
roosterfish: Is there any reason for tracking the `knownRemotePools` outside of the func? It looks this can also be a local variable inside the func. It also has the benefit that you don't have to pass it as an argument.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -188,6 +189,20 @@ func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *types.Clus
 		result.UIURL = lxdMembers[0].URL
 	}
 
+	storagePools, err := lxdClient.GetStoragePools()
+	if err != nil {
+		return fmt.Errorf("Failed to get storage pools: %w", err)
+	}
+
+	memberLocalStorageDrivers := []string{"btrfs", "dir", "lvm", "zfs"}
```
roosterfish: Do we have to hardcode the drivers? This smells like something that breaks once we would have other local drivers.

roosterfish: Why not just reporting it the same way for both local and remote drivers? This would mean for remote driver we would duplicate the information but at least it's consistent and from the outside we don't have learn the concept of local/remote drivers but treat them all the same way.

cmd/microcloudd/cluster_manager_task.go
edlerd: The `knownRemotePools` are tracking which cluster wide pools we already added. To avoid reporting cluster wide pool states multiple times. Because we call this function for each cluster member.

Another option is to check if the remote pool is already in the `result *types.ClusterManagerPostStatus`. Then we could eliminate the `knownRemotePools` variable and maybe that is anyway more straight forward to follow what is happening. I'll change it accordingly.

cmd/microcloudd/cluster_manager_task.go
edlerd: I think we need to know which pools are remote or not. Without this information, it gets hard to report correctly in cluster manager.

Luckily there is the `/1.0` response from LXD giving us the `environment.storage_supported_drivers` list with a `Remote` flag that holds exactly this information. I'll change it to rely on that instead of having a list here.

cmd/microcloudd/cluster_manager_task.go
edlerd: Error status pools or if the resource tracking is not disabled.

cmd/microcloudd/cluster_manager_task.go
tomponline: yes please, no hard coded driver lists please

edlerd: See update below. Now relying on the api for this info :)

api/types/cluster_manager.go
```diff
@@ -63,6 +63,14 @@ type StatusDistribution struct {
 	Count  int64  `json:"count"`
```
roosterfish, line 63: Please split the changes into smaller commits, at least separated by package.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -210,10 +216,7 @@ func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *types.Clus
 		cpuLoad5 += memberState.SysInfo.LoadAverages[1]
 		cpuLoad15 += memberState.SysInfo.LoadAverages[2]
 
-		for _, poolsState := range memberState.StoragePools {
-			result.DiskTotalSize += int64(poolsState.Space.Total)
-			result.DiskUsage += int64(poolsState.Space.Used)
-		}
+		enrichStoragePoolMetrics(member, memberState, localPools, result)
```
roosterfish, line 219: Let's move `localPools, err := getLocalPools(lxdClient)` into `enrichStoragePoolMetrics` as it seems to be only used there so there is no need to pass it as a function arg.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -235,3 +238,73 @@ func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *types.Clus
 
 	return nil
 }
+
+func enrichStoragePoolMetrics(member api.ClusterMember, memberState *api.ClusterMemberState, localPools []string, result *types.ClusterManagerPostStatus) {
+	for name, poolState := range memberState.StoragePools {
+		if poolState.Space.Total == 0 || poolState.Space.Used == 0 {
+			// error state or no available info on this pool
```
roosterfish: Not a blocker but we tend to write code comments like this `// Start with uppercase .... and end with dot.`.

api/types/cluster_manager.go
```diff
@@ -63,6 +63,14 @@ type StatusDistribution struct {
...
 }
 
+// StoragePoolUsage represents the storage usage distribution.
+type StoragePoolUsage struct {
+	Name   string `json:"name"`
+	Member string `json:"member"`
+	Total  int64  `json:"total"`
```
roosterfish: Can we also use `uint64` here? This way we don't need the `int64()` cast below. And I think we will never have negative values anyway?

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -188,6 +189,11 @@ func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *types.Clus
 		result.UIURL = lxdMembers[0].URL
 	}
 
+	localPools, err := getLocalPools(lxdClient)
+	if err != nil {
+		return fmt.Errorf("Failed to get local storage pools: %w", err)
```
roosterfish: In belows error you say `Failed to get LXD server info`. Should this rather be `Failed to get local LXD storage pools` ?

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -235,3 +238,73 @@ func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *types.Clus
...
+
+func getLocalPools(lxdClient lxd.InstanceServer) ([]string, error) {
+	server, _, err := lxdClient.GetServer()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get LXD server info: %w", err)
+	}
+
+	var localDrivers []string
+	for _, d := range server.Environment.StorageSupportedDrivers {
+		if !d.Remote {
```
roosterfish, line 292: Thanks, I like this approach much better.

roosterfish (changes requested):
Only a few minor comments.

cmd/microcloudd/cluster_manager_task.go
edlerd: I want to avoid calling it multiple times. We are in a loop over the members here and only want to collect the pool drivers once.

edlerd:
@roosterfish this is ready from my side for another review / merge.

cmd/microcloudd/cluster_manager_task.go
roosterfish: Makes sense thanks.

cmd/microcloudd/cluster_manager_task.go
roosterfish: Is it worth adding a log statement too?

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -235,3 +238,73 @@ func enrichClusterMemberMetrics(lxdClient lxd.InstanceServer, result *types.Clus
...
+			localDrivers = append(localDrivers, d.Name)
+		}
+	}
+
+	storagePools, err := lxdClient.GetStoragePools()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get storage pools: %w", err)
```
roosterfish: "Failed to get LXD storage pools"

roosterfish (changes requested):
Thanks, only one small comment. If we settle on this data structure (see comment on the other PR) then this one looks almost ready to me.

roosterfish (approved):
LGTM!

PR #1166, opened 2025-12-19
cmd/microcloud/cluster_manager: send metrics from all cluster members

edlerd:
# Done
- extend cluster manager metrics reporting:
- send metrics from all cluster members
- and prepare model for metrics from other services

relates to https://github.com/canonical/microcloud-cluster-manager/pull/360

### DIFF
--- a/service/lxd.go
+++ b/service/lxd.go
@@ -292,6 +292,29 @@ func (s LXDService) DeleteToken(ctx context.Context, tokenName string, address s
 	return fmt.Errorf("No corresponding join token operation found for %q", tokenName)
 }
 
+// Metrics fetches the metrics from the LXD daemon at the given address with a remote client.
+func (s LXDService) Metrics(ctx context.Context, address string) (string, error) {
+	var c lxd.InstanceServer
+	var err error
+	if address != "" {
+		c, err = s.remoteClient(nil, address, CloudPort)
+	} else {
+		c, err = s.Client(ctx)
+	}
+
+	if err != nil {
+		return "", err
+	}
+
+	// Get the cluster member metrics.
+	metrics, err := c.GetMetrics()
+	if err != nil {
+		return "", err
+	}
+
+	return metrics, nil
+}
+
 // RemoteClusterMembers returns a map of cluster member names and addresses from the MicroCloud at the given address.
 // Provide the certificate of the remote server for mTLS.
 func (s LXDService) RemoteClusterMembers(ctx context.Context, cert *x509.Certificate, address string) (map[string]string, error) {


api/types/cluster_manager.go
```diff
@@ -63,6 +63,13 @@ type StatusDistribution struct {
 	Count  int64  `json:"count"`
```
roosterfish, line 63: Please split the changes into smaller commits by package.

api/types/cluster_manager.go
```diff
@@ -63,6 +63,13 @@ type StatusDistribution struct {
...
 }
 
+// ServerMetrics represents the metrics from one cluster member service.
+type ServerMetrics struct {
+	Member  string `json:"member"`
```
roosterfish: In the other PR you have added a new type `StoragePoolUsage` which also has a `member` field. 
Why not having a map indexed by member names to track all the other items like `StoragePoolUsage` and `ServerMetrics` underneath? 

Otherwise we have to duplicate the `Member` field various times.

Something like this maybe:

```go
type MemberStatus struct {
    ServerMetrics
    StoragePoolUsage
    ...
}

type ClusterManagerPostStatus struct {
    ...
    Members map[string]MemberStatus
}
```

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -167,23 +173,27 @@ func enrichInstanceMetrics(lxdClient lxd.InstanceServer, result *types.ClusterMa
...
-	}
+func enrichServerMetrics(lxdClient lxd.InstanceServer, lxdMembers []api.ClusterMember, result *types.ClusterManagerPostStatus) error {
+	for i := range lxdMembers {
+		member := lxdMembers[i]
 
-	result.Metrics = metrics
+		lxdClient.UseTarget(member.ServerName)
+		metrics, err := lxdClient.GetMetrics()
+		if err != nil {
+			return fmt.Errorf("Failed to get LXD metrics: %w", err)
```
roosterfish: In case one member is offline, do we still want to report the metrics of the others? This could be the case if there is some maintenance going on and some cluster members were shut down.

api/types/cluster_manager.go
edlerd: We have also cluster wide pools, so pushing the pool usage into a cluster member hierarchy wouldn't allow representing such pools.

cmd/microcloudd/cluster_manager_task.go
edlerd: Yes good idea.

api/services_proxy.go
```diff
@@ -49,7 +49,7 @@ func proxy(sh *service.Handler, name, path string, handler endpointHandler) rest
 		Name:              name,
 		Path:              path,
 
-		Get:    rest.EndpointAction{Handler: authHandlerMTLS(sh, handler), ProxyTarget: true},
+		Get:    rest.EndpointAction{Handler: authHandlerMTLS(sh, handler), ProxyTarget: false},
```
edlerd: Changing this here because the metrics response is not a json structure and the proxy would trip over in [microclusters rest.go](https://github.com/canonical/microcluster/blob/v3/internal/rest/rest.go#L159) otherwise. Problem is this proxy is intercepting the LXD `GET /1.0/metrics?target=member2` request and forwards it itself and fails to parse the text only response. We can rely on LXD to do the forwarding, though, so disabling fixes the issue at hand.

I'd like a more granular way to disable this instead of globally. Ideas for how to do that are very welcome :)

api/types/cluster_manager.go
```diff
@@ -63,6 +63,13 @@ type StatusDistribution struct {
...
+	Metrics string `json:"metrics"`
+	Service string `json:"service"`
```
roosterfish: We already have constants for the various different services types in MicroCloud. Could you please use those instead? It's in `api/types/services.go` and the type is `ServiceType`.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -167,23 +174,35 @@ func enrichInstanceMetrics(lxdClient lxd.InstanceServer, result *types.ClusterMa
 	return err
 }
 
-func enrichServerMetrics(lxdClient lxd.InstanceServer, result *types.ClusterManagerPostStatus) error {
-	metrics, err := lxdClient.GetMetrics()
-	if err != nil {
-		return fmt.Errorf("Failed to get LXD metrics: %w", err)
-	}
+func enrichServerMetrics(ctx context.Context, lxdService *service.LXDService, lxdMembers []api.ClusterMember, result *types.ClusterManagerPostStatus) error {
+	for i := range lxdMembers {
```
roosterfish: You can do `for _, member := range lxdMembers` directly.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -167,23 +174,35 @@ func enrichInstanceMetrics(lxdClient lxd.InstanceServer, result *types.ClusterMa
...
+		if err != nil {
+			// If we can't get the metrics of a member, skip it but continue with others.
+			logger.Error("Could not fetch metrics for cluster member", logger.Ctx{"member": member, "err": err})
+			continue
+		}
+
+		result.ServerMetrics = append(result.ServerMetrics, types.ServerMetrics{
+			Member:  member.ServerName,
+			Metrics: metrics,
+			Service: "lxd",
```
roosterfish: Here you can then use the `LXD` const of type `ServiceType`. 
It's actual value is `"LXD"` in uppercase so we ensure we always use the same scheme when talking about services in the MicroCloud API. See my comment from before on using the constants/type.

service/lxd.go
```diff
@@ -292,6 +292,29 @@ func (s LXDService) DeleteToken(ctx context.Context, tokenName string, address s
 	return fmt.Errorf("No corresponding join token operation found for %q", tokenName)
```
roosterfish, line 292: Can you please add a headline to the commit message and put the explanation in a new line? 
This helps better reading through the list of commits.

api/types/cluster_manager.go
roosterfish: Mh I know thought about it a bit. 

1) `MemberStatues`, `InstancesStatuses` and `ServerMetrics` report per cluster member info. However the `CPU*` fields seem to be pre-computed before sent to cluster manager.
Should we make this consistent and also report them per member and cluster manager does the computation? 
This also gives more flexibility in case it's needed in the future by cluster manager.

2) In the other PR where you add the `StoragePoolUsage`, you already establish the concept of not populating it's `Name` field in case it's a cluster wide storage pool.
In this case if we have a map, can't we just set the maps index to an empty string to not set a member?

I am not too confident with both approaches, but trying to find the most suitable data structure here.

edlerd: > Mh I know thought about it a bit.
> 
> 1. `MemberStatues`, `InstancesStatuses` and `ServerMetrics` report per cluster member info. However the `CPU*` fields seem to be pre-computed before sent to cluster manager.
>    Should we make this consistent and also report them per member and cluster manager does the computation?
>    This also gives more flexibility in case it's needed in the future by cluster manager.

Both `MemberStatues` and `InstancesStatuses` are aggregates over the whole cluster. Counts per status, they are not cluster member specific.

`ServerMetrics` exists for each cluster member as each member gives a different subset of instance and local storage pool metrics. For cluster manager it is a bit special because this information is forwarded to Prometheus and not ingested by it.

> 2. In the other PR where you add the `StoragePoolUsage`, you already establish the concept of not populating it's `Name` field in case it's a cluster wide storage pool.
>    In this case if we have a map, can't we just set the maps index to an empty string to not set a member?

Yes, `StoragePoolUsage` will have some pools repeated for each cluster member. Those entries represent the local storage pool on that member, such as pools with the `directory` driver. Pools using the `ceph` driver will keep the `member` field empty and have only one entry in the `StoragePoolUsage`. I think pushing those cluster-wide pools into a map with a `<cluster member, member details>` structure might save some bits on the payload. But this seems less clear and more confusing.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -167,23 +174,35 @@ func enrichInstanceMetrics(lxdClient lxd.InstanceServer, result *types.ClusterMa
...
-	if err != nil {
-		return fmt.Errorf("Failed to get LXD metrics: %w", err)
-	}
+func enrichServerMetrics(ctx context.Context, lxdService *service.LXDService, lxdMembers []api.ClusterMember, result *types.ClusterManagerPostStatus) error {
+	for i := range lxdMembers {
+		member := lxdMembers[i]
+		u, err := url.Parse(member.URL)
+		if err != nil {
+			// If we can't parse the URL of a member, skip it but continue with others.
+			logger.Error("Could not parse URL for cluster member", logger.Ctx{"member": member, "err": err})
```
roosterfish: Let's also log the faulty `member.URL` to get some clarity in case this happens.

cmd/microcloudd/cluster_manager_task.go
roosterfish: Thanks for fixing this in the loop in `enrichClusterMemberMetrics` below. Do you mind also doing the change here?

roosterfish (changes requested):

cmd/microcloudd/cluster_manager_task.go
edlerd: Ah yes, there is another older occasion for looping through instances as well.

roosterfish:
@edlerd Please check the conflict caused by merging the other PR.

PR #1176, opened 2026-01-13
doc: how-to about storage.backups_volume and images_volume

minaelee:
Adds information about how to set `storage.backups_volume` and `storage.images_volume` manually, including when to do so.

Note: In testing the behavior of these settings, I found that they are not used for persistent storage. Instead, they are used as working directories when dealing with packing/unpacking/exporting/importing image and instance backup files. The actual image and instance tarballs are stored in other places. Thus, the main reason to recommend them is that, in the absence of local storage, they will prevent the host filesystem from being filled up during work. It does not seem to be a concern related to persistent storage of tarballs.

### DIFF
--- a/doc/how-to/initialize.md
+++ b/doc/how-to/initialize.md
@@ -104,6 +104,12 @@ Complete the following steps to initialize MicroCloud:
       ```
 
    1. You can choose to optionally set up a CephFS distributed file system.
+
+      ```{admonition} If you did not set up local storage
+      :class: note
+      If you did not set up local storage _and_ you set up CephFS storage, there will be an additional step for you at the end of the initialization process.
+      ```
+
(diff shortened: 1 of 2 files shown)


doc/how-to/initialize.md
```diff
@@ -218,3 +229,61 @@ ovn:
...
+:class: note
+If you initialized MicroCloud without local storage _and_ with CephFS storage, continue to the section below to complete your initialization.
+```
+
+(howto-initialize-images-backups)=
+## Configure `backups_volume` and `images_volume`
+
+```{admonition} When this applies
+:class: tip
+You only need to configure these options if you did not set up any local storage _and_ you set up CephFS storage.
```
roosterfish: "if you did not set up local storage"

The additional "any" suggests you can setup different types of local storage? Whilst this is possible in LXD, MicroCloud only offers ZFS which will cause the additional configuration of the `backups_volume` and `images_volume` config keys.

minaelee: Thanks, fixed!

PR #1207, opened 2026-02-02
doc: bump RTD go version

minaelee:
RTD docs build failing due to RTD attempting to use go version 1.25.4 when 1.25.5 is required for building the LXD client.

Bumped to 1.25.6 in doc/.readthedocs.yaml to prevent needing to bump it twice as LXD will shortly require 1.25.6.

### DIFF
--- a/doc/.readthedocs.yaml
+++ b/doc/.readthedocs.yaml
@@ -9,7 +9,7 @@ version: 2
 build:
   os: ubuntu-24.04
   tools:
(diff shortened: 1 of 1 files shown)


minaelee:
Closing this PR as it turns out it is not possible in RTD to specify a patch release in the .readthedocs.yaml file. Will submit a separate PR with an alternative method of ensuring the correct version of golang.

PR #1208, opened 2026-02-02
doc: set GOTOOLCHAIN=auto for RTD Go builds

minaelee:
Update `.readthedocs.yaml` to set `GOTOOLCHAIN=auto` when building the lxd integration and `make doc-html-rtd`. This fixes RTD build issues by allowing Go to auto-fetch the required Go patch version when the local toolchain is outdated.

### DIFF
--- a/doc/.readthedocs.yaml
+++ b/doc/.readthedocs.yaml
@@ -13,14 +13,14 @@ build:
     python: "3.12"
   commands:
       - git fetch --unshallow || true
-      - cd doc && make integrate
-      - cd doc/integration/lxd/ && go build -ldflags "-s -w" -o trimpath -o lxc.bin ./lxc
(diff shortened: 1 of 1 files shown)


doc/.readthedocs.yaml
```diff
@@ -13,14 +13,14 @@ build:
     python: "3.12"
   commands:
       - git fetch --unshallow || true
-      - cd doc && make integrate
-      - cd doc/integration/lxd/ && go build -ldflags "-s -w" -o trimpath -o lxc.bin ./lxc
+      - cd doc && make GOTOOLCHAIN=auto integrate
```
roosterfish, line 16: This means we still have to update the `golang: "1.25"` line above once we go to `1.26` if I understand correctly. But it ensures the right patch version is picked. LGTM.

PR #1209, opened 2026-02-02
doc: add cluster manager reference architecture

minaelee:
Add reference architecture documentation for MicroCloud Cluster Manager.


doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+  html_meta:
+    description: Reference architecture for MicroCloud Cluster Manager, a Kubernetes-based web application for viewing and managing multiple MicroCloud deployments. 
+---
+
+(ref-cluster-manager-architecture)=
+# Architecture of the MicroCloud Cluster Manager
+
+The MicroCloud Cluster Manager is a centralized tool that provides an overview of MicroCloud deployments. In its initial implementation, it provides an overview of resource usage and availability for all clusters. Future implementations will include centralized cluster management capabilities.
+
+Cluster Manager stores the data from registered clusters in Postgres and Prometheus databases. This data can be displayed in the Cluster Manager UI, which also links to Grafana dashboards for each MicroCloud.
```
edlerd: > which also links to Grafana dashboards for each MicroCloud

This is a possible extension. By default, the COS stack is not available. So a user will deploy cluster manager and get the manager UI without links to Grafana.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+(ref-cluster-manager-architecture-overview)=
+## Architecture overview
+
+Cluster Manager is a distributed web application that runs inside a Kubernetes cluster to achieve high availability. The diagram below illustrates its system architecture:
+
+```{figure} ../images/cluster_manager_architecture.png
+   :alt: A diagram of Cluster Manager architecture
+   :align: center
+```
```
edlerd: This diagram is from an earlier development environment. It is mostly correct, but some things have slightly changed.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+Inside the Kubernetes cluster are the following system components:
+
+DNS and static external IP
+: The Domain Name Server (DNS) must be set up by the user to resolve their domain names to a static external IP.
+That static external IP acts as the gateway to route user traffic to the appropriate Kubernetes load balancers.
+
+TCP load balancers
+: Two TCP load balancer services distribute traffic to the Management API and Cluster Connector deployments without terminating TLS. Instead, TLS termination is handled directly within each deployment application. This approach is particularly crucial for the Cluster Connector deployment, as it relies on mutual TLS (mTLS) authentication for secure communication.
```
edlerd: We are using a single Traefik instance that is dealing with the incoming requests, no two load balancers anymore.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+Certificate manager
+: Manages TLS/SSL certificates for secure communication within the Kubernetes cluster. It stores secrets in Kubernetes to be used by various components. The certificates are used by both the Management API and Cluster Connector deployments for HTTPS encryption.
```
edlerd: We now rely on a charm that implements the certificates interface to provide certificates.  This can be the `self-signed-certificates` charm, as suggested in the readme. We do not rely on the certificate manager k8s app anymore.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+Postgres deployment
+: A PostgreSQL database deployed within the Kubernetes cluster. It provides persistent storage for system data. Both the Management API and Cluster Connector will communicate with the Postgres database for CRUD operations.
+
+Persistent Volume (PV) and Persistent Volume Claim (PVC)
+: The Persistent Volume is the storage resource provisioned for the Postgres deployment. The Persistent Volume Claim is the request for storage by the Postgres deployment to ensure data persistence.
```
edlerd: We rely on the canonical Postgres charm. How that charm does persistent storage is outside our control.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+{ref}`ref-cluster-manager-architecture-management`
+: Responsible for serving the static UI assets, exposing API endpoints for the UI to communicate with the Cluster Manager backend. Requests from the UI are authenticated using OIDC.
+
+{ref}`ref-cluster-manager-architecture-connector`
+: Responsible for handling requests from remote clusters, authenticated using mTLS.
```
edlerd: Shall we say Microclouds instead of "remote clusters"?

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+(ref-cluster-manager-architecture-management)=
+## Management API deployment
```
edlerd:
```suggestion
## Management API
```

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+The management API deployment handles local operations in Cluster Manager, including:
```
edlerd:
```suggestion
The management API handles local operations in Cluster Manager, including:
```

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+- Listing active MicroCloud clusters
+- Creating cluster join tokens
+- Approving join requests from clusters
+- Serving the UI's static assets
+
+(ref-cluster-manager-architecture-management-ingress)=
+### Management API deployment ingress
```
edlerd:
```suggestion
### Management API ingress
```

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+Management API traffic passes through a TCP load balancer without TLS termination. The Management API terminates TLS itself using its certificate.
+
+Upon Cluster Manager startup, the certificate manager generates a server certificate, which is stored in a Kubernetes secret. This certificate secret is mounted as a volume within the service pods and also available as an environment variable.
```
edlerd: This might be too much detail, also not all correct, as our production deployment changed. This was true in the past for the dev environment, but not anymore.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+The Management API deployment uses the certificate to configure and run an HTTPS server.
+
+(ref-cluster-manager-architecture-management-oidc)=
+### OIDC authentication
+
+The Management API deployment is secured using OIDC authentication. Users need to provide the OIDC configurations using a Kubernetes `configMap` to inform the Cluster Manager which identity provider is used.
```
edlerd: We are using the charm configurations, see https://charmhub.io/microcloud-cluster-manager-k8s/actions

We do not expect users to interact with the k8s cluster or the config map directly.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+| Key         | Description |
+| ------------| ----------- |
+| secret      | A secret to be used by a MicroCloud cluster for creating a [HMAC](https://en.wikipedia.org/wiki/HMAC) signature for the join request.      |
+| expires_at  | Expiry date for the remote cluster join token. |
+| address     | The address at which the MicroCloud Cluster Manager is reachable. This address can be a domain name or external static IP. |
+| server_name | This is unique and stored for reference purposes in Cluster Manager to map which cluster the token belongs to. |
+| fingerprint | The public key from the Cluster Connector deployment certificate (secret in Kubernetes cluster). Used to establish mTLS between Cluster Manager and the cluster. |
+
+On a member of the registering cluster, the token is used in the command `microcloud cluster-manager join <token>`. The join request is sent to the Cluster Manager for approval. During this step, MicroCloud generates a dedicated certificate and a HMAC signature based on the join request payload, using the join token secret. The join request payload from MicroCloud includes the cluster name and certificate only.
```
edlerd: MicroCloud does not generate a new certificate, but uses its existing cluster certificate

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+The registering cluster begins sending periodic status updates to the Cluster Manager as soon as a join request response is received from the Manager. Until the join request is approved, the Manager will discard any status update data it receives from the cluster.
```
edlerd: There is no approval of the join request or discarding of status updates.
```suggestion
The registering cluster begins sending periodic status updates to the Cluster Manager as soon as a join request response is received from the Manager.
```

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+This is due to the unidirectional nature of communication initialization between clusters and the Cluster Manager. The Cluster Manager can only receive and respond to data from the cluster; it cannot initiate a message to tell the cluster when the join request is approved. For details about this, see {ref}`ref-cluster-manager-architecture-connector-ingress`.
+
+Once the Cluster Manager receives a join request, it tries to match the cluster name in the payload to an entry in its `remote_cluster_tokens` table. If it finds a match, it uses the corresponding token secret stored in that table to verify the join request HMAC signature. The validity of the remote cluster join token is also checked against its expiry date.
+
+If a valid match is found, the matched token is removed and the cluster is registered with the following information:
+
+- `name` - extracted from the token
+- `cluster_certificate` - received from the join request from the registering cluster. This is a dedicated certificate generated by MicroCloud for establishing mutual TLS authentication with the Cluster Manager.
```
edlerd: This is the MicroCloud cluster certificate, not a dedicated one.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+A corresponding entry in the remote_clusters table is created, adding the following information:
+
+- `status` - `PENDING_APPROVAL`
```
edlerd: status is `ACTIVE` right away.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+Finally, a user in the Cluster Manager approves or rejects the join request. Before approving, they must confirm that the fingerprint of the certificate received in the Cluster Manager matches that of the registering cluster. This ensures that the remote cluster join request was sent by the correct cluster.
```
edlerd:
```suggestion
```

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+If the join request is approved, then the cluster is marked as `ACTIVE` in the Cluster Manager. If the join request is rejected, then all data related to the cluster is removed, which causes an error response to be returned to the registering cluster when it tries to send data.
```
edlerd:
```suggestion
```

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+Once a cluster has been successfully registered, Cluster Manager begins storing status update data sent by the cluster.
+
+(ref-cluster-manager-architecture-management-ui)=
+### UI
+
+The Management API deployment handles serving static assets for the UI. Users access information about clusters through the UI. Through it, users can create remote cluster join tokens and view information about existing tokens, as well as approve or reject join requests.
```
edlerd:
```suggestion
The Management API deployment handles serving static assets for the UI. Users access information about clusters through the UI. Through it, users can create remote cluster join tokens and view information about existing tokens.
```

edlerd: We can expand here, we serve warnings and metric insights on a high level as well as a list of all registered clusters.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+### mTLS authentication
+
+During the initial join request of a cluster, each cluster presents a dedicated certificate to the Cluster Manager. The Manager uses this certificate to authenticate all subsequent requests from the cluster, using mTLS. For efficiency considerations, these certificates are cached after the first authenticated request.
+
+Due to the mTLS requirement, the TCP load balancer passes through TLS traffic and the Cluster Connector terminates TLS itself.
+
+(ref-cluster-manager-architecture-connector-status)=
+### Status signals
+
+MicroCloud clusters send periodic status signals along with data about resource usage and availability. A cluster status update includes the following information:
```
edlerd: It might be worth saying the status signals are sent by the db-leader of the cluster.

Also should we use the term "Heartbeat" for those?

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+
+- Cluster level details including:
+  - Number of cluster-wide instances and distribution of instance status (such as how many instances are stopped or started)
+  - Number of cluster members and distribution of member status (number of members online, number of members in error status, and so on)
+  - CPU, memory, and disk utilization for each cluster, as aggregated totals across all cluster members
+- MicroCloud certificate within the request context for mTLS authentication. The certificate fingerprint is used to look up the cluster ID in a cache for updating cluster details.
+
+For each status update request, the Cluster Manager performs the following tasks:
+
+- Match status update request by certificate fingerprint; check that the cluster exists and is marked as active (join approved)
```
edlerd:
```suggestion
- Match status update request by certificate fingerprint; check that the cluster exists and is marked as active
```

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,155 @@
...
+- mTLS authentication check against the matched certificate
+- Store and overwrite data in the `remote_cluster_details` table
+
+To avoid overwhelming the Cluster Connector deployment, the status endpoints are rate limited. The response sent to the originating cluster includes a delay period (in seconds) that must pass before the next status signal request.
```
edlerd: All endpoints are rate limited, not just this one.

doc/reference/cluster-manager-architecture.md
minaelee: Is there an updated diagram, or can you let me know what has changed and I can update it?

doc/reference/cluster-manager-architecture.md
minaelee: Fixed to:
>A TCP load balancer (using a [Traefik](https://traefik.io/) instance) distributes traffic to the Management API and Cluster Connector deployments without terminating TLS.

doc/reference/cluster-manager-architecture.md
minaelee: Does this update work, or would you prefer we did not mention Grafana at all?
>This data can be displayed in the Cluster Manager UI, which can be extended to link to Grafana dashboards for each MicroCloud.

Note: This information is from https://github.com/canonical/microcloud-cluster-manager/blob/main/ARCHITECTURE.md and likely should be updated there as well.

doc/reference/cluster-manager-architecture.md
minaelee: Should the Certificate manager section in lines 33-34 above be removed entirely?

doc/reference/cluster-manager-architecture.md
minaelee: What information should we provide in this section instead, or should we remove it entirely?

doc/reference/cluster-manager-architecture.md
minaelee: Done

doc/reference/cluster-manager-architecture.md
minaelee: >The Management API handles serving static assets for the UI. Users access information about clusters through the UI. Through it, users can create remote cluster join tokens and view information about existing tokens. The UI also serves warnings and metric insights on a high level.

I added this, but "on a high level" could bear more explanation. Do you mean through optional extension with Grafana, or something more/else?

doc/reference/cluster-manager-architecture.md
minaelee: I didn't want to use terms that I didn't already find in my references, but if you prefer it, I will adjust to using the word "heartbeat".

minaelee: Updated to:
```
(ref-cluster-manager-architecture-connector-heartbeats)=
### Heartbeats

The `db-leader` of each connected MicroCloud cluster sends periodic heartbeats to the Cluster Manager, along with data about resource usage and availability. A heartbeat update includes the following information:
```

doc/reference/cluster-manager-architecture.md
minaelee: Updated to:

```
(ref-cluster-manager-architecture-rate-limited)=
## Rate limited endpoints

To avoid overwhelming the Cluster Manager, all its endpoints are rate limited. When any endpoint receives a request from a cluster, the response from Cluster Manager includes a delay period (in seconds) that must pass before the next request to that endpoint.
```

Or did you mean all endpoints for the Cluster Connector deployment only? 

Also: do you want to change the term "Cluster Connector deployment" to "Cluster Connector" (like with "Management API deployment" to "Management API") or does it make sense to keep the word "deployment" here?

doc/reference/cluster-manager-architecture.md
minaelee: Updated to:
```
- `cluster_certificate` - received from the join request from the registering cluster. This is the MicroCloud cluster certificate for establishing mutual TLS authentication with the Cluster Manager.
```

doc/reference/cluster-manager-architecture.md
minaelee: Updated to:
```
On a member of the registering cluster, the token is used in the command `microcloud cluster-manager join <token>`. The join request is sent to the Cluster Manager for approval. The request payload includes the registering cluster's name and cluster certificate.
```

doc/reference/cluster-manager-architecture.md
minaelee: I can remove the whole section, or shorten it to something like this:

```
The Management API configures and runs an HTTPS server to make API endpoints available. Traffic to the server passes through a TCP load balancer.
```

doc/reference/cluster-manager-architecture.md
minaelee: Does this work?

```
The Management API is secured using OIDC authentication, using the [microcloud-cluster-manager-k8s](https://charmhub.io/microcloud-cluster-manager-k8s) charm configurations. The charm handles providing OIDC information to the Kubernetes cluster and its `configMap`.
```

doc/reference/cluster-manager-architecture.md
edlerd: Yes, suggestion sounds good to me. I'll take a note to update the architecture file.

doc/reference/cluster-manager-architecture.md
edlerd: I think we can remove it, yes.

doc/reference/cluster-manager-architecture.md
edlerd: I think we can remove it, yes.

doc/reference/cluster-manager-architecture.md
edlerd: Your suggestion above looks good to me :+1:

doc/reference/cluster-manager-architecture.md
edlerd: Yes, perfect :)

doc/reference/cluster-manager-architecture.md
edlerd: High level means aggregates of instances and microclouds cluster members. Like the number of instances and their status distribution (how many are started/stopped/etc). If the cluster manager is extended with COS/Grafana stack, then Grafana indeed holds detailed information about each instance in every cluster.

doc/reference/cluster-manager-architecture.md
edlerd: I think the suggestion is slightly confusing. We have rate limiting in place to avoid overwhelming the cluster manager, yes.

The functionality to signal to the Microcloud in a response when they should call in again is unrelated to the rate limiting, though.

doc/reference/cluster-manager-architecture.md
edlerd: We don't have  an updated diagram yet. Things that have changed:
- A single TCP load balancer instead of two, exposing two different domain names. One for the management-api and one for the cluster-connector.
- Postgres service / pg deployment and volume claims are "just" one thing: the postgres charm. The rest is detail of the PG charm. the diagram exposes to much detail of the PG charm internals with assumptions that might be wrong
- Cert manager is to be replaced by a charm implementing the "certificate" charm interface. We might just change the label here.
- k8s secrets/k8s config to be replaced by a juju config layer. Under the hood this is still true. I am not sure how to unify the levels of detail in the diagram to surface k8s internals and charm/juju internals.
- management-api and cluster-connector live together on the same container. Each container is running those two processes. there can be multiple containers to scale out.
- we might want to add Canonical observability stack as an optional extension to the diagram. With prometheus and grafana.

edlerd: I can create a task for myself to create an updated diagram.

minaelee: OK, I would volunteer to update it but it sounds like there are still some things to be decided (e.g., whether to add COS). Let me know if I can help - if you were to give me a finalized list of updates to make, I could give it a shot.

doc/reference/cluster-manager-architecture.md
minaelee: Updated to:

>The Management API serves the UI's static assets. Users can use the UI to access information about registered clusters, as well as create or view remote cluster join tokens.
>
>The UI also serves high-level metrics insights and warnings, including aggregated instance counts and status distributions (such as started vs. stopped). Detailed per-instance metrics can be viewed with Grafana dashboards if the Cluster Manager is extended to use the Canonical Observability Stack (COS).

doc/reference/cluster-manager-architecture.md
minaelee: I moved it to the end of the Cluster Connector deployment ingress section:

```
### Cluster Connector deployment ingress

We expect each cluster to be able to reach the Cluster Manager. However, we do not expect the Manager to be able to reach each remote cluster directly. This is because clusters might not be exposed on an internet-facing IP, or they might run behind a firewall or NAT. Therefore, operations consist of ingress traffic only.

All endpoints are rate limited to avoid overwhelming the Cluster Manager.
```

edlerd: LGTM :+1:

doc/reference/cluster-manager-architecture.md
edlerd: Oh, that would be great!

Let's include the COS stack, but mark it as optional in some way.

I suggest showing the flow from the cluster connector to Prometheus. And to show how the user will be able to connect to the COS stack through Grafana, reading the same Prometheus. COS has its own load balancer that is distinct from the cluster manager load balancer.

edlerd: The list above should be exhaustive as for the needed changes. Please let me know if you have a newer version, and we can review it together :)

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+Inside the Kubernetes cluster are the following system components:
+
+DNS and static external IP
+: The Domain Name Server (DNS) must be set up by the user to resolve their domain names to a static external IP.
+That static external IP acts as the gateway to route user traffic to the appropriate Kubernetes load balancers.
+
+TCP load balancers
+: A TCP load balancer (using a [Traefik](https://traefik.io/) instance) distributes traffic to the Management API and Cluster Connector deployments without terminating TLS. Instead, TLS termination is handled directly within each deployment application. This approach is particularly crucial for the Cluster Connector deployment, as it relies on mutual TLS (mTLS) authentication for secure communication.
```
edlerd:
```suggestion
: A TCP load balancer (using a [Traefik](https://traefik.io/) instance) distributes traffic to the Management API and Cluster Connector without terminating TLS. Instead, TLS termination is handled directly within each application. This approach is particularly crucial for the Cluster Connector, as it relies on mutual TLS (mTLS) authentication for secure communication.
```
A bit too much `deployment` in here ;)

minaelee: Agreed. Do you still want to use the phrase "Cluster Connector deployment" throughout, or can we shorten that to "Cluster Connector" like we did with the Management API deployment?

edlerd: Yes, I think we can use `Cluster Connector` without the "deployment" suffix everywhere.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,151 @@
...
+  html_meta:
+    description: Reference architecture for MicroCloud Cluster Manager, a Kubernetes-based web application for viewing and managing multiple MicroCloud deployments.
+---
+
+(ref-cluster-manager-architecture)=
+# Architecture of the MicroCloud Cluster Manager
+
+The MicroCloud Cluster Manager is a centralized tool that provides an overview of MicroCloud deployments. In its initial implementation, it provides an overview of resource usage and availability for all clusters. Future implementations will include centralized cluster management capabilities.
+
+Cluster Manager stores the data from registered clusters in Postgres and Prometheus databases. This data can be displayed in the Cluster Manager UI, which can be extended to link to Grafana dashboards for each MicroCloud.
```
edlerd: Prometheus is also part of the possible extension. In the default deployment without COS, there is only Postgres and no Prometheus.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,151 @@
...
+
+```{figure} ../images/cluster_manager_architecture.svg
+   :alt: A diagram of Cluster Manager architecture
+   :align: center
+```
+
+Inside the Kubernetes cluster are the following system components:
+
+DNS
+: The Domain Name Server (DNS) must be set up by the user to resolve their domain names to a static external IP. That static external IP acts as the gateway to route user traffic to the appropriate Kubernetes load balancers.
```
edlerd: DNS is "domain name system", not "server".

The actual domain name server is very likely outside kubernetes. So the leading sentence above gives a wrong impression of the setup.

Also we removed the static external IP blocks from the diagram. So maybe we simplify this to "A DNS name (such as cluster-manager.example.com) is configured to point to a public external IP of the load balancer in kubernetes .

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,151 @@
...
+
+TCP load balancer
+: A TCP load balancer (using a [Traefik](https://traefik.io/) instance) distributes traffic to the Management API and Cluster Connector without terminating TLS. TLS termination is handled directly within those two components. This approach is particularly crucial for the Cluster Connector, which relies on mutual TLS (mTLS) authentication for secure communication.
+
+Postgres charm
+: A PostgreSQL database deployed within the Kubernetes cluster using a charm. It provides persistent storage for system data. The Management API and Cluster Connector both communicate with the Postgres database for CRUD operations.
```
edlerd: Not only CRUD, we also store the heartbeat data.

Should we highlight here we are using the canonical postgresql charm, see https://canonical.com/data/postgresql?

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,151 @@
...
+
+{ref}`ref-cluster-manager-architecture-management`
+: Responsible for serving the static UI assets, exposing API endpoints for the UI to communicate with the Cluster Manager backend. Requests from the UI are authenticated using OpenID Connect (OIDC). Deployed along with the Cluster Connector inside one or more containers.
```
edlerd, line 33: Is this section missing a "Management API" heading?

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,151 @@
...
+
+{ref}`ref-cluster-manager-architecture-connector`
+: Responsible for handling requests from MicroCloud clusters, authenticated using mutual TLS (mTLS). Deployed along with the Management API inside a container inside one or more containers.
```
edlerd:
```suggestion
: Responsible for handling requests from MicroCloud clusters, authenticated using mutual TLS (mTLS). Deployed along with the Management API inside a container. Both can be scaled from one to many containers for high availability and load distribution.
```
Is this missing a heading like "Cluster connector"?

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,151 @@
...
+
+Certificate charm
+: The certificate charm manages TLS/SSL certificates for secure communication within the Kubernetes cluster. It stores the Kubernetes secrets managed by the Juju configuration layer. The certificates are used by the Management API and the Cluster Connector for HTTPS encryption
```
edlerd: I think there are no secrets stored by the charm.

The certificates are for external communication, not internal.
```suggestion
: The certificate charm manages TLS/SSL certificates for secure communication within the Kubernetes cluster. The certificates are used by the Management API and the Cluster Connector for HTTPS encryption
```

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,151 @@
...
+
+: Juju configuration layer
+Canonical's [Juju application orchestration engine](https://canonical.com/juju) is used to manage Kubernetes configuration and secrets.  
+
+Canonical Observability Stack
+: Optional. The Cluster Connector connects to the [Canonical Observability Stack](https://documentation.ubuntu.com/observability) (COS) through Grafana, reading from the Prometheus database. COS has its own load balancer that is distinct from the Cluster Manager load balancer.
```
edlerd: Cluster connector is writing to prometheus, but not grafana. Grafana is the UI for exploring metrics stored in prometheus.

doc/reference/cluster-manager-architecture.md
edlerd: I am slightly confused. Maybe not missing a heading, because there is another similar section below.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,151 @@
...
+(ref-cluster-manager-architecture-management-registering)=
+### Registering clusters
+
+To register a MicroCloud cluster with a Cluster Manager, the user first creates a remote cluster join token in the Manager. This token is base64-encoded and includes the following information:
+
+| Key         | Description |
+| ------------| ----------- |
+| secret      | A secret to be used by a MicroCloud cluster for creating a [HMAC](https://en.wikipedia.org/wiki/HMAC) signature for the join request.      |
+| expires_at  | Expiry date for the remote cluster join token. |
+| address     | The address at which the MicroCloud Cluster Manager is reachable. This address can be a domain name or external static IP. |
```
edlerd:
```suggestion
| address     | The address at which the Cluster Manager is reachable. This address can be a domain name or external static IP. |
```
Maybe simpler?

edlerd (commented):
This is great! Some thoughts on the new docs below.

doc/reference/cluster-manager-architecture.md
minaelee: I removed the mention of Prometheus here. It is mentioned later in the description of the optional COS component, under the diagram.

doc/reference/cluster-manager-architecture.md
minaelee: Agreed, I removed the DNS entry and added this sentence to the end of the TCP load balancer entry:

>A DNS name (such as `cluster-manager.example.com`) is configured to point to the public external IP of the load balancer.

doc/reference/cluster-manager-architecture.md
minaelee: The `{ref}` renders the heading as links like this:

image

See the rendered docs in the PR preview: https://canonical-ubuntu-documentation-library--1209.com.readthedocs.build/microcloud/1209/microcloud/reference/cluster-manager-architecture/

You can always access the PR preview for PRs that affect the docs by clicking on the RTD link in the tests:

image

doc/reference/cluster-manager-architecture.md
minaelee: Sounds good. Updated to:

>: A PostgreSQL database deployed within the Kubernetes cluster using the [Canonical PostgreSQL charm](https://canonical.com/data/postgresql). It provides persistent storage for system data. The Management API and Cluster Connector both communicate with the Postgres database for CRUD operations, and the Cluster Connector uses it to store {ref}`heartbeat data <ref-cluster-manager-architecture-connector-heartbeats>`.

doc/reference/cluster-manager-architecture.md
edlerd: Ah nice, that is much easier to review than the markdown source. Thanks for the hint!

edlerd: Can resolve this and the other similar question.

doc/reference/cluster-manager-architecture.md
minaelee: Got it, removed that sentence. Thanks!

doc/reference/cluster-manager-architecture.md
minaelee: Agreed, updated.

doc/reference/cluster-manager-architecture.md
minaelee: Updated to:

>The Cluster Connector can integrate with the [Canonical Observability Stack](https://documentation.ubuntu.com/observability) (COS) by exporting metrics to a Prometheus database, enabling visualization and analysis through the Grafana UI.

doc/reference/cluster-manager-architecture.md
minaelee: Updated the URL to be more specific and link to the PostgreSQL K8s charm here: https://charmhub.io/postgresql-k8s

At least, my understanding is that is the charm to be used

doc/reference/cluster-manager-architecture.md
minaelee: Does the above mean that the Cluster Manager always uses the self-signed-certificates charm as the "Certificates charm"? If so, I can link that charm (https://charmhub.io/self-signed-certificates) in the "Certificates charm" description. Or does the above mean that the self-signed-certificates charm is one option that can be used (among multiple other options)?

doc/reference/cluster-manager-architecture.md
minaelee: Is the Traefik charm (https://charmhub.io/traefik-k8s) always used for this? If so, we can mention that here and link to the charm (instead of the current link, which goes to traefik.io)

doc/reference/cluster-manager-architecture.md
edlerd: Good idea :+1:

doc/reference/cluster-manager-architecture.md
edlerd: Yes, it is always the traefik-k8s charm. Because we use a very specific juju interface that needs traefik.

doc/reference/cluster-manager-architecture.md
edlerd: No, the self-signed-certificates charm is one of the options. Deployment can use any other charm implementing the `certificates` and `send-ca-cert` interfaces. You can see a list of options here: https://charmhub.io/microcloud-cluster-manager-k8s/integrations#send-ca-cert and https://charmhub.io/microcloud-cluster-manager-k8s/integrations#certificates Any charm used must support both interfaces, so I see also "Vault" and "Lego" as options among others.

edlerd: Though other charms should work, arguably, we have only ever tested it with the self-signed-certificates charm.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,148 @@
...
+: Optional. The Cluster Connector can integrate with the [Canonical Observability Stack](https://documentation.ubuntu.com/observability) (COS) by exporting metrics to a Prometheus database, enabling visualization and analysis through the Grafana UI. COS uses its own load balancer, separate from the Cluster Manager load balancer.
+
+(ref-cluster-manager-architecture-management)=
+## Management API
+
+The management API handles local operations in Cluster Manager, including:
+
+- Listing active MicroCloud clusters
+- Creating cluster join tokens
+- Approving join requests from clusters
```
edlerd:
```suggestion
```
This workflow was simplified. We don't require approval anymore.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,148 @@
...
+- Serving the UI's static assets
```
edlerd:
```suggestion
- Serving the UI's static assets and dynamic data for the web interface
```
Maybe expand here a bit?

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,148 @@
...
+| server_name | This is unique and stored for reference purposes in Cluster Manager to map which cluster the token belongs to. |
+| fingerprint | The public key from the Cluster Connector certificate (secret in Kubernetes cluster). Used to establish mTLS between Cluster Manager and the cluster. |
+
+On a member of the registering cluster, the token is used in the command `microcloud cluster-manager join <token>`. The join request is sent to the Cluster Manager for approval. The request payload includes the registering cluster's name and cluster certificate.
+
+The registering cluster begins sending periodic heartbeats along with status updates to the Cluster Manager. These heartbeats and updates begin as soon as a join request response is received from the Manager.
+
+This is due to the unidirectional nature of communication initialization between clusters and the Cluster Manager. The Cluster Manager can only receive and respond to data from the cluster; it cannot initiate a message to tell the cluster when the join request is approved. For details about this, see {ref}`ref-cluster-manager-architecture-connector-ingress`.
```
edlerd:
```suggestion
This is due to the unidirectional nature of communication initialization between clusters and the Cluster Manager. The Cluster Manager can only receive and respond to data from the cluster. For details about this, see {ref}`ref-cluster-manager-architecture-connector-ingress`.
```
We simplified the approval process.  Maybe we can remove this whole paragraph?

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,148 @@
...
+  - Number of cluster-wide instances and distribution of instance status (such as how many instances are stopped or started)
+  - Number of cluster members and distribution of member status (number of members online, number of members in error status, and so on)
+  - CPU, memory, and disk utilization for each cluster, as aggregated totals across all cluster members
+- MicroCloud certificate within the request context for mTLS authentication. The certificate fingerprint is used to look up the cluster ID in a cache for updating cluster details.
+
+For each heartbeat it receives, the Cluster Manager performs the following tasks:
+
+- Match status update request by certificate fingerprint; check that the cluster exists and is marked as active.
+- mTLS authentication check against the matched certificate
+- Store and overwrite data in the `remote_cluster_details` table
```
edlerd, line 145: We could expand here, that if COS is deployed, detailed cluster metrics are received from the cluster as part of the heartbeat. Those metrics are forwarded to Prometheus by Cluster Manager.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,148 @@
...
+| ------------| ----------- |
+| secret      | A secret to be used by a MicroCloud cluster for creating a [HMAC](https://en.wikipedia.org/wiki/HMAC) signature for the join request.      |
+| expires_at  | Expiry date for the remote cluster join token. |
+| address     | The address at which the Cluster Manager is reachable. This address can be a domain name or external static IP. |
+| server_name | This is unique and stored for reference purposes in Cluster Manager to map which cluster the token belongs to. |
+| fingerprint | The public key from the Cluster Connector certificate (secret in Kubernetes cluster). Used to establish mTLS between Cluster Manager and the cluster. |
+
+On a member of the registering cluster, the token is used in the command `microcloud cluster-manager join <token>`. The join request is sent to the Cluster Manager for approval. The request payload includes the registering cluster's name and cluster certificate.
+
+The registering cluster begins sending periodic heartbeats along with status updates to the Cluster Manager. These heartbeats and updates begin as soon as a join request response is received from the Manager.
```
edlerd: One part of me wants to move this section down because of the timeline alignment. This only happens after the join request gets a success response. The below paragraphs explain what happens until the success response is generated.

The other part of me would like to keep this here to close the thought about the cluster side and continue below with the cluster manager part.

Not really sure what is best here. Both seem suitable options.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,148 @@
...
+
+```{figure} ../images/cluster_manager_architecture.svg
+   :alt: A diagram of Cluster Manager architecture
+   :align: center
+```
+
+Inside the Kubernetes cluster are the following system components:
+
+TCP load balancer
+: A TCP load balancer (using the [Traefik charm](https://charmhub.io/traefik-k8s)) distributes traffic to the Management API and Cluster Connector without terminating TLS. TLS termination is handled directly within those two components. This approach is particularly crucial for the Cluster Connector, which relies on mutual TLS (mTLS) authentication for secure communication. A DNS name (such as `cluster-manager.example.com`) is configured to point to the public external IP of the load balancer.
```
edlerd: I just now realize, that DNS is optional. We also support using public IP only without any domain in play. Shall we adjust this here slightly to mention both options?

doc/reference/cluster-manager-architecture.md
minaelee: Done

doc/reference/cluster-manager-architecture.md
minaelee: Done

doc/reference/cluster-manager-architecture.md
minaelee: In that case, should the diagram also be updated to show the possibility of either DNS or external static IP?

Something like this:

image

doc/reference/cluster-manager-architecture.md
minaelee: Are those detailed metrics _only_ received from the cluster if COS is deployed, and absent otherwise?

doc/reference/cluster-manager-architecture.md
minaelee: Removed, thanks.

minaelee: Also changed this further down:

>The join request is sent to the Cluster Manager for approval.

to

>The join request is sent to the Cluster Manager.

doc/reference/cluster-manager-architecture.md
edlerd: Metrics are always sent, but discarded if COS is not deployed. If COS is deployed they are forwarded to prometheus.

doc/reference/cluster-manager-architecture.md
edlerd: Yes, I was thinking of this too, but didn't have a good idea how to visualize "DNS or IP". I think you suggestion looks good 👍

doc/reference/cluster-manager-architecture.md
edlerd: Perfect, thanks.

edlerd:
Hey @minaelee did you maybe forget to push the latest changes?

doc/reference/cluster-manager-architecture.md
minaelee: I prefer moving it down, and it required reworking that section a little more. Here's the updated version:

```markdown
On a member of the registering cluster, the token is used in the command `microcloud cluster-manager join <token>`. The join request is sent to the Cluster Manager. The request payload includes the registering cluster's name and cluster certificate.

Once the Cluster Manager receives the join request, it tries to match the cluster name in the payload to an entry in its `remote_cluster_tokens` table. If it finds a match, it uses the corresponding token secret stored in that table to verify the join request HMAC signature. The validity of the remote cluster join token is also checked against its expiry date.

If a valid match is found, the matched token is removed and the cluster is registered with the following information:

- `name` - extracted from the token
- `cluster_certificate` - received from the join request from the registering cluster. This is the MicroCloud cluster certificate for establishing mutual TLS authentication with the Cluster Manager.

A corresponding entry in the remote_clusters table is created, adding the following information:

- `status` - `ACTIVE`

Once a cluster has been successfully registered, it begins to send periodic {ref}`heartbeats <ref-cluster-manager-architecture-connector-heartbeats>` that include status update data to the Cluster Manager.
```

doc/reference/cluster-manager-architecture.md
minaelee: Added this to below the list of the cluster-level details sent with the heartbeat:

>If the [Canonical Observability Stack](https://documentation.ubuntu.com/observability) (COS) is deployed, the Cluster Manager forwards the cluster-level details to the Prometheus database used by COS.

doc/reference/cluster-manager-architecture.md
minaelee: Updated to:

>Certificate charm
: The certificate charm manages TLS/SSL certificates for secure communication within the Kubernetes cluster. Any charm that implements both the [certificates](https://charmhub.io/microcloud-cluster-manager-k8s/integrations#certificates) and [send-ca-cert](https://charmhub.io/microcloud-cluster-manager-k8s/integrations#send-ca-cert) interfaces can be used, such as the [self-signed-certificates charm](https://charmhub.io/self-signed-certificates). The certificates are used by the Management API and the Cluster Connector for HTTPS encryption.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,145 @@
...
+
+Postgres charm
+: A PostgreSQL database deployed within the Kubernetes cluster using the [Canonical PostgreSQL K8s charm](https://charmhub.io/postgresql-k8s). It provides persistent storage for system data. The Management API and Cluster Connector both communicate with the Postgres database for CRUD operations, and the Cluster Connector uses it to store {ref}`heartbeat data <ref-cluster-manager-architecture-connector-heartbeats>`.
+
+{ref}`ref-cluster-manager-architecture-management`
+: Responsible for serving the static UI assets, exposing API endpoints for the UI to communicate with the Cluster Manager backend. Requests from the UI are authenticated using OpenID Connect (OIDC). Deployed along with the Cluster Connector inside one or more containers.
+
+{ref}`ref-cluster-manager-architecture-connector`
+: Responsible for handling requests from MicroCloud clusters, authenticated using mutual TLS (mTLS). Deployed along with the Management API inside a container inside one or more containers.
```
roosterfish: `Deployed along with the Management API inside one or more containers.`

roosterfish (approved):
Only a small nit, but we can also address this in a follow up PR. @edlerd waiting for your final review.

doc/reference/cluster-manager-architecture.md
minaelee: Fixed, thanks!

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,145 @@
...
+
+{ref}`ref-cluster-manager-architecture-management`
+: Responsible for serving the static UI assets, exposing API endpoints for the UI to communicate with the Cluster Manager backend. Requests from the UI are authenticated using OpenID Connect (OIDC). Deployed along with the Cluster Connector inside one or more containers.
+
+{ref}`ref-cluster-manager-architecture-connector`
+: Responsible for handling requests from MicroCloud clusters, authenticated using mutual TLS (mTLS). Deployed along with the Management API inside one or more containers.
+
+Certificate charm
+: The certificate charm manages TLS/SSL certificates for secure communication within the Kubernetes cluster. Any charm that implements both the [certificates](https://charmhub.io/microcloud-cluster-manager-k8s/integrations#certificates) and [send-ca-cert](https://charmhub.io/microcloud-cluster-manager-k8s/integrations#send-ca-cert) interfaces can be used, such as the [self-signed-certificates charm](https://charmhub.io/self-signed-certificates). The certificates are used by the Management API and the Cluster Connector for HTTPS encryption.
```
edlerd:
```suggestion
: The certificate charm manages TLS/SSL certificates for secure communication. Any charm that implements both the [certificates](https://charmhub.io/microcloud-cluster-manager-k8s/integrations#certificates) and [send-ca-cert](https://charmhub.io/microcloud-cluster-manager-k8s/integrations#send-ca-cert) interfaces can be used, such as the [self-signed-certificates charm](https://charmhub.io/self-signed-certificates). The certificates are used by the Management API and the Cluster Connector for HTTPS encryption.
```
Also for outside communication, not only internal.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,145 @@
...
+
+: Juju configuration layer
+Canonical's [Juju application orchestration engine](https://canonical.com/juju) is used to manage Kubernetes configuration and secrets.
```
edlerd:
```suggestion
Juju configuration layer
: Canonical's [Juju application orchestration engine](https://canonical.com/juju) is used to manage Kubernetes configuration and secrets.
```

Formatting is slightly bamboozled here. I guess there should be a heading.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,145 @@
...
+### Management API ingress
+
+The Management API configures and runs an HTTPS server to make API endpoints available. Traffic to the server passes through a TCP load balancer.
+
+(ref-cluster-manager-architecture-management-oidc)=
+### OIDC authentication
+
+The Management API is secured using OIDC authentication, using the [`microcloud-cluster-manager-k8s`](https://charmhub.io/microcloud-cluster-manager-k8s) charm configurations. The charm handles providing OIDC information to the Kubernetes cluster and its `configMap`.
+
+The OIDC flow is similar to the {ref}`approach implemented in LXD <lxd:authentication-openid>`:
```
edlerd, line 66: The link doesn't work on the dem server, is this correct?

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,145 @@
...
+- Subsequent requests use the session cookie to validate authentication.
+
+(ref-cluster-manager-architecture-management-enroll)=
+### Enroll clusters
+
+To enroll a MicroCloud cluster with a Cluster Manager, the user first creates a remote cluster join token in the Manager. This token is base64-encoded and includes the following information:
+
+| Key         | Description |
+| ------------| ----------- |
+| secret      | A secret to be used by a MicroCloud cluster for creating a [HMAC](https://en.wikipedia.org/wiki/HMAC) signature for the join request.      |
```
edlerd:
```suggestion
| secret      | A secret to be used by a MicroCloud cluster for creating an [HMAC](https://en.wikipedia.org/wiki/HMAC) signature for the join request.      |
```
Wikipedia also uses "an" here. I think that is the correct term, but not 100% sure.

doc/reference/cluster-manager-architecture.md
```diff
@@ -0,0 +1,145 @@
...
+| expires_at  | Expiry date for the remote cluster join token. |
+| address     | The address at which the Cluster Manager is reachable. This address can be a domain name or external static IP. |
```
edlerd:
```suggestion
| address     | The address at which the Cluster Manager is reachable. This address can be a domain name or static IP. |
```
As discussed earlier :)

doc/reference/cluster-manager-architecture.md
edlerd: - meaning the link to `lxd:authentication-openid`

doc/reference/cluster-manager-architecture.md
minaelee: Makes sense if it's pronounced Aich-MAC. I'll update.

doc/reference/cluster-manager-architecture.md
minaelee: Good catch, thanks!

doc/reference/cluster-manager-architecture.md
minaelee: It works for me on my local dev server and is not flagged as a bad reference target by Sphinx, so I think this is OK. If you mean that it does not work in the RTD preview, that is expected due to the way that MicroCloud integrates and links to the LXD docs set.

roosterfish (approved):
LGTM!

PR #1222, opened 2026-02-12
Add ceph status to cluster manager heartbeats

edlerd:
# Done

* add ceph statuses distribution for cluster manager
* enrich ceph statuses to heartbeat message for cluster manager

see https://github.com/canonical/microcloud-cluster-manager/pull/411

### DIFF
--- a/cmd/microcloudd/cluster_manager_task.go
+++ b/cmd/microcloudd/cluster_manager_task.go
@@ -129,6 +129,12 @@ func sendClusterManagerStatusMessage(ctx context.Context, sh *service.Handler, s
 		return nextUpdate
 	}
 
+	err = enrichCephStatuses(ctx, sh, &payload)
+	if err != nil {
+		logger.Error("Failed to enrich Ceph statuses", logger.Ctx{"err": err})
+		return nextUpdate
+	}
+
 	clusterCert, err := cloud.ClusterCert()
 	if err != nil {
 		logger.Error("Failed to get cluster certificate", logger.Ctx{"err": err})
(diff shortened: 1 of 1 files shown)


cmd/microcloudd/cluster_manager_task.go
```diff
@@ -129,6 +130,19 @@ func sendClusterManagerStatusMessage(ctx context.Context, sh *service.Handler, s
...
 	}
 
+	cephService := sh.Services[types.MicroCeph].(*service.CephService)
+	cephClient, err := cephService.Client("")
+	if err != nil {
+		logger.Error("Failed to get Ceph client", logger.Ctx{"err": err})
+		return nextUpdate
+	}
+
+	err = enrichCephStatuses(ctx, cephClient, &payload)
```
roosterfish: Why passing the client as arg? Can't the client be created directly inside `enrichCephStatuses`?

roosterfish (changes requested):

cmd/microcloudd/cluster_manager_task.go
edlerd: I was following the structure for the lxd client and assumed we might need the ceph client later for fetching ceph metrics. But we can easily pull it up later on once we need it in other functions. Will move it into the function.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -156,6 +162,37 @@ func sendClusterManagerStatusMessage(ctx context.Context, sh *service.Handler, s
 	return nextUpdate
 }
 
+func enrichCephStatuses(ctx context.Context, sh *service.Handler, result *types.ClusterManagerPostStatus) error {
+	if sh.Services[types.MicroCeph] == nil {
```
roosterfish, line 166: When the MicroCloud gets deployed without Ceph, is this the point where we just quit trying to attempt fetching information about the Ceph cluster?

roosterfish (approved):

cmd/microcloudd/cluster_manager_task.go
roosterfish: Looks like it, we are using the same pattern already in `askRemotePool`.

PR #1223, opened 2026-02-16
Doc: Fix broken links

roosterfish:
A link on snap management and the ref for LXD live migration have been changed.


tugbataluy (approved):
LGTM, thanks ❇️ . The links you fixed look correct.

PR #1239, opened 2026-02-23
Test: Fixes to unblock the pipeline

roosterfish:
Accommodate latest MicroOVN snap by installing latest core26 snap.

Also ensure LXD is started to prevent issues where MicroCloud exits because LXD didn't came up in time (slow runners). I saw these type of failures appear more often the last two weeks, e.g. https://github.com/canonical/microcloud/actions/runs/22100391546/job/64501597361#step:4:6490.

### DIFF
--- a/test/includes/microcloud.sh
+++ b/test/includes/microcloud.sh
@@ -861,6 +861,11 @@ reset_system() {
       lxc exec "${name}" -- ip link set "${iface}" up
       lxc exec "${name}" -- sysctl -wq "net.ipv6.conf.${iface}.disable_ipv6=1"
     done
+
+    # Force LXD to start up.
+    # Slower runners might cause longer LXD startup times.
+    # To not force bumping numbers in MicroCloud, ensure LXD is started before MicroCloud starts checking the socket.
+    lxc exec "${name}" -- lxc info >/dev/null
   )
 
   echo "==> Reset ${name}"


test/includes/microcloud.sh
```diff
@@ -861,6 +861,11 @@ reset_system() {
       lxc exec "${name}" -- ip link set "${iface}" up
       lxc exec "${name}" -- sysctl -wq "net.ipv6.conf.${iface}.disable_ipv6=1"
     done
+
+    # Force LXD to start up.
+    # Slower runners might cause longer LXD startup times.
+    # To not force bumping numbers in MicroCloud, ensure LXD is started before MicroCloud starts checking the socket.
+    lxc exec "${name}" -- lxc info >/dev/null
```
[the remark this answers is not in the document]
roosterfish: `lxc info` blocks until the server is started or returns right away in case the LXD daemon is already running. This is what we want to have.

test/includes/microcloud.sh
```diff
@@ -1299,6 +1304,9 @@ setup_system() {
     retry lxc exec "${name}" -- apt-get install --no-install-recommends -y ${packages}
     retry lxc exec "${name}" -- snap install snapd
 
+    # Install core26 to allow latest MicroOVN to be used.
+    retry lxc exec "${name}" -- snap install core26 --channel latest/edge
```
simondeziel, line 1308: FYI, this feels like pointless user-friction IMHO so I've asked the snap folks if this could be avoided by having `snapd` doing the right thing and pulling the needed base, even if in `devel` grade.

simondeziel (approved):
LGTM, thanks

test/includes/microcloud.sh
roosterfish: Nice thanks for follow up. I was wondering why snapd cannot download this by default when a snap depends on it. So it's just because the `core26` isn't yet released as a stable version?

simondeziel: That's my interpretation of the error message. I've yet to hear from the snap folks but you can follow the discussion in the `snappy-internal` channel.

PR #1240, opened 2026-02-24
Go: Bump to latest microcluster

roosterfish:
Used to validate the Microcluster v3 release candidate. It doesn't yet use the release tag of Microcluster v3 as it's not yet tagged.

With this change I have also removed the ability to run `microcloud cluster list <address>` against other cluster members as there is no point in offering/requiring this. Instead we use Microcluster's app func `GetClusterMembers` wich always talks against the local unix socket.

### DIFF
--- a/cmd/microcloudd/cluster_manager_task.go
+++ b/cmd/microcloudd/cluster_manager_task.go
@@ -10,7 +10,7 @@ import (
 	"github.com/canonical/lxd/client"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
-	"github.com/canonical/microcluster/v3/state"
+	microTypes "github.com/canonical/microcluster/v3/microcluster/types"
 
 	"github.com/canonical/microcloud/microcloud/api/types"
 	"github.com/canonical/microcloud/microcloud/client"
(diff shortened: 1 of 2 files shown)


api/services_auth.go
```diff
@@ -28,7 +27,7 @@ func authHandlerMTLS(sh *service.Handler, f endpointHandler) endpointHandler {
 
 		// Use certificate based authentication between cluster members.
 		if r.TLS != nil {
-			trustedCerts := s.Remotes().CertificatesNative()
+			trustedCerts := s.Truststore().RemoteCertificatesNative()
 			for _, cert := range r.TLS.PeerCertificates {
 				// First evaluate the permanent turst store.
```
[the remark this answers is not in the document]
roosterfish: Not a change in this PR, something for a follow up.

api/cluster_manager.go
```diff
@@ -13,9 +13,7 @@ import (
 
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
-	"github.com/canonical/microcluster/v3/microcluster/rest"
-	"github.com/canonical/microcluster/v3/microcluster/rest/response"
-	"github.com/canonical/microcluster/v3/state"
+	"github.com/canonical/microcluster/v3/microcluster/types"
```
tugbataluy, line 16: The `github.com/canonical/microcluster/v3/microcluster/types` package is imported with different conventions across files:

Some files use an aliased import: microTypes "github.com/canonical/microcluster/v3/microcluster/types"
Others use the unaliased/default import: "github.com/canonical/microcluster/v3/microcluster/types"

For consistency across the codebase, can we consider standardizing on one approach? Using microTypes as an alias is probably preferable here since it avoids potential name collisions with other types packages.

roosterfish: I have intentionally only used an alias in files where we also import MicroCloud's own `github.com/canonical/microcloud/microcloud/api/types`. 

To me it feels easier if we use an alias only if required. But in this case let's be consistent and stick to `microTypes`. We also use `cephTypes` and `ovnTypes` in other files.

PR #1241, opened 2026-02-25
Ask for initial UI access link during bootstrap

MusicDin:
Adds interactive question asking whether the initial UI access link should be created.

Requires update once the following PRs are merged:
- [x] https://github.com/canonical/lxd/pull/17742

### DIFF
--- a/test/suites/basic.sh
+++ b/test/suites/basic.sh
@@ -14,11 +14,18 @@ test_interactive() {
   export SETUP_OVN_EXPLICIT="no"
   export CEPH_CLUSTER_NETWORK="${microcloud_internal_net_addr}"
   export CEPH_PUBLIC_NETWORK="${microcloud_internal_net_addr}"
+  export CREATE_UI_ACCESS_LINK="yes"
   join_session init micro01 micro02 micro03
+
+  if check_api_extension auth_bearer micro01; then
+    # Ensure initial UI access link was printed by the bootstrap flow.
+    lxc exec micro01 -- cat out | grep -Fx "UI initial access link (expires in 1 day):"
+  fi
 
   lxc exec micro01 -- tail -1 out | grep "MicroCloud is ready" -q
   lxc exec micro02 -- tail -2 out | head -1 | grep "Successfully joined the MicroCloud cluster and closing the session" -q
   lxc exec micro03 -- tail -2 out | head -1 | grep "Successfully joined the MicroCloud cluster and closing the session" -q
+
   for m in micro01 micro02 micro03 ; do
     validate_system_lxd "${m}" 3
     validate_system_microceph "${m}"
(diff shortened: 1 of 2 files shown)


cmd/microcloud/ask.go
```diff
@@ -1869,3 +1871,71 @@ func (c *initConfig) askJoinConfirmation(gw *cloudClient.WebsocketGateway, servi
...
+	} else if uiAdminIdentity.Type != api.IdentityTypeBearerTokenInitialUI {
+		return "", fmt.Errorf("A bearer identity with name %q already exists but is not of type %q", uiAdminIdentityName, api.IdentityTypeBearerTokenInitialUI)
+	}
+
+	token, err := lxdClient.IssueBearerIdentityToken(uiAdminIdentityName, api.IdentityBearerTokenPost{})
+	if err != nil {
+		return "", fmt.Errorf("Failed to issue bearer token for initial UI access link: %w", err)
+	}
+
+	uiAccessLink := api.NewURL().Scheme("https").Host(address+":"+strconv.FormatInt(service.LXDPort, 10)).WithQuery("token", token.Token)
```
roosterfish: Just a nit, you can use `types.AddrPort` from Microcluster. It allows you to `address.String()` so you don't have to manually concat address and port.

test/suites/basic.sh
```diff
@@ -14,11 +14,18 @@ test_interactive() {
   export SETUP_OVN_EXPLICIT="no"
   export CEPH_CLUSTER_NETWORK="${microcloud_internal_net_addr}"
   export CEPH_PUBLIC_NETWORK="${microcloud_internal_net_addr}"
+  export CREATE_UI_ACCESS_LINK="yes"
   join_session init micro01 micro02 micro03
 
+  if check_api_extension "auth_bearer"; then
```
roosterfish: The check for the extension is already performed in the MicroCloud code. 
So if `CREATE_UI_ACCESS_LINK=yes`, shouldn't we always expect a token to be printed?

I understand that if we run this with `5.21/stable` the extension isn't present. 
So should we instead check for the extension inside the `microcloud_interactive` func that either appends a `yes` or `no` to the `setup` var depending on the list of available API extensions?

MusicDin: I agree that I should guard the yes/no resp.

However, for the output check, if the LXD does not have such extension, there will be no output, so avoiding this check would result in an error when searching for Initial UI access link in output.

cmd/microcloud/ask.go
MusicDin: Thanks

test/includes/check.sh
```diff
@@ -27,6 +27,17 @@ check_dependencies() {
 	fi
 }
 
+check_api_extension() {
+	ext="$1"
+
+	if ! lxc info | grep -qx "^- ${ext}$"; then
```
roosterfish: You are running `lxc info` against the LXD on the test runner, not against the `microX` instance running MicroCloud and a version of LXD which either does or does not have the requested extension.

Maybe you can pass the name of the `microX` member as an additional function argument?

test/suites/basic.sh
roosterfish: I see what you mean. But we should also not try to append a `yes` to the `setup` in case it's not even supported by the current version of LXD.

This maybe isn't problematic right now, but it might become an issue if we add another question (after the UI access link) which would be answered by the `yes`/`no` in case `CREATE_UI_ACCESS_LINK` is set.

roosterfish (approved):
Thanks for adding the extra API extension checks, LGTM!

PR #1336, opened 2026-04-16
doc: append version number in RTD version switcher flyout menu

minaelee:
We recently updated the default version of the docs to use the name/slug of "default", and it's set up to use the branch corresponding to the most current LTS. This way, we can permanently point to the same URL for the default docs, without having to change the URL when we have a new LTS.

Unfortunately, this means that in the ReadTheDocs version switcher flyout menu, the default version appears as "default" without anything to indicate the version number. This PR adds custom JavaScript that, along with a custom ReadTheDocs environment variable, updates the flyout to add the version after default, like this:

image

Note for the future: the RTD environment variable is configured as FLYOUT_DEFAULT_VERSION_LABEL at https://app.readthedocs.com/dashboard/canonical-microcloud/environmentvariables/ and must be updated when a new LTS is released, along with the change to the branch corresponding to the "default" version slug in the RTD version configuration. RTD does not allow you to edit environment variables, only delete and create them, so you must delete the existing one and create a new one with the same name. It _must_ be set as Public (via checkbox at creation) so that the custom javascript can access it when run from the browser.

## Checklist

### DIFF
--- a/doc/.sphinx/_templates/base.html
+++ b/doc/.sphinx/_templates/base.html
@@ -3,6 +3,7 @@
 {% block theme_scripts %}
 <script>
   const github_url = "{{ github_url }}";
+  window.flyoutDefaultVersionLabel = "{{ flyout_default_version_label }}";
 </script>
 {% endblock theme_scripts %}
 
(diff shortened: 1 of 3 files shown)


roosterfish (approved):

roosterfish:
Also added a note about `FLYOUT_DEFAULT_VERSION_LABEL` to the release notes.

PR #1344, opened 2026-04-20
doc: Pin MicroCeph squid (v2-edge)

roosterfish:
I just saw the branch is available so we should start pinning it to not slide any tentacle docs into 2 LTS.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -3,7 +3,7 @@ current_dir := $(dir $(abspath $(firstword $(MAKEFILE_LIST))))
 # Set the path prefix and MicroCloud component versions corresponding to each MicroCloud docs version.
 PATH_PREFIX      = /microcloud/v2/
 LXDVERSION       = origin/stable-5.21
-MICROCEPHVERSION = 64d02d9538a89be039707ecbf942ca1aeae29760 # origin/main.
+MICROCEPHVERSION = origin/squid
 MICROOVNVERSION  = origin/branch-24.03
 
 SPHINX_DIR       ?= .sphinx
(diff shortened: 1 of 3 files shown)


roosterfish:
@minaelee we can now pin MicroCeph.

As part of this I am including a commit which I have skipped previously as it failed the link check. 
Is the failure we can see in the spell check job related to the loaded objects.inv file?

roosterfish:
If we are not including this commit, the RTD job fails with:

```
/home/docs/checkouts/readthedocs.org/user_builds/canonical-microcloud/checkouts/1344/doc/how-to/initialize.md:16: WARNING: unknown document: 'microceph:explanation/security/full-disk-encryption' [ref.doc]
/home/docs/checkouts/readthedocs.org/user_builds/canonical-microcloud/checkouts/1344/doc/how-to/initialize.md:100: WARNING: unknown document: 'microceph:explanation/security/full-disk-encryption' [ref.doc]
/home/docs/checkouts/readthedocs.org/user_builds/canonical-microcloud/checkouts/1344/doc/reference/requirements.md:72: WARNING: unknown document: 'microceph:explanation/security/full-disk-encryption' [ref.doc]
```

minaelee:
> @minaelee we can now pin MicroCeph.
> 
> As part of this I am including a commit which I have skipped previously as it failed the link check. Is the failure we can see in the spell check job related to the loaded objects.inv file?

We discussed this f2f but noting here for posterity that this failure is due to the fact that for the GH CI check, to avoid having to clone and build all integrated docs for every PR, we use the MicroCeph objects.inv file from their https://canonical-microceph.readthedocs-hosted.com/en/v19.2.0-squid/ version. This file currently differs from the objects.inv created by cloning and building MicroCeph from its new `squid` branch (done by our RTD build), because MicroCeph has not yet updated its v19.2.0-squid docs version to use the `squid` branch; it still builds from a tag created 8 months ago.

This means that since the RTD build uses the `squid` branch (which has the updated URL for the about-fde page), we can ignore the GH CI failure here for now, as the actual docs on RTD will build correctly. Ideally, the `v19.2.0-squid` version on MicroCeph would point to the `squid` branch.

PR #1364, opened 2026-04-28
doc: add markdown linter

minaelee:
This adds the pymarkdownlnt markdown linter as part of standard documentation tooling. It includes adding a GH CI workflow to run the linter when files inside doc/ are changed, and it also updates a couple of Markdown files that fail the linting rule that requires blank lines around lists.

### DIFF
--- a/.github/workflows/markdown-style-checks.yml
+++ b/.github/workflows/markdown-style-checks.yml
@@ -1,0 +1,40 @@
+name: Markdown style checks
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    # Make sure bash is always invoked with `-eo pipefail`
+    # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell
+    shell: bash
+
+on:
+  workflow_call:
+  push:
+    branches:
+    - main
+    paths:
+    - 'doc/**'   # Only run on changes to the doc directory
+  pull_request:
+    paths:
+    - 'doc/**'   # Only run on changes to the doc directory
+
+jobs:
+  markdown-lint:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        fetch-depth: 0
+        persist-credentials: false
+    - name: Create venv
+      working-directory: "doc"
+      run: make install
+    - name: Lint markdown
+      working-directory: "doc"
+      run: make lint-md
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -15,6 +15,12 @@ LXDVERSION       = origin/main
 MICROCEPHVERSION = origin/main
 MICROOVNVERSION  = origin/main
 
+SPHINX_DIR       ?= .sphinx
+DOCS_SOURCEDIR   ?= .
+DOCS_VENVDIR     ?= $(SPHINX_DIR)/venv
+DOCS_VENV        ?= $(DOCS_VENVDIR)/bin/activate
+INTEGRATION_DIR  ?= integration
+
 # Put it first so that "make" without argument is like "make help".
 help:
 	@echo "\n" \
@@ -31,6 +37,7 @@ help:
         "* check accessibility:                       make pa11y \n" \
         "* check style guide compliance:              make vale \n" \
         "* check style guide compliance on target:    make vale TARGET=* \n" \
+        "* check markdown:                            make lint-md \n" \
         "* other possible targets:                    make <TAB twice> \n" \
         "------------------------------------------------------------- \n"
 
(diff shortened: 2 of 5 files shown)


.github/workflows/markdown-style-checks.yml
```diff
@@ -0,0 +1,28 @@
...
+    branches:
+    - '*'
+    paths:
+    - 'doc/**'   # Only run on changes to the doc directory
+
+jobs:
+  markdown-lint:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
roosterfish, line 31: Starting with https://github.com/canonical/microcloud/commit/be4f94b9a404d44a539e05e0a6e0611296656fc0 the `actions/checkout` jobs now set `persist-credentials: false`. 
Likely we should do the same here for consistency.

.github/workflows/markdown-style-checks.yml
```diff
@@ -0,0 +1,28 @@
+name: Markdown style checks
```
roosterfish, line 1: Can you please split into two commits, one that adds the new markdown check and config and one that fixes the lint issues in the actual doc files thanks.

roosterfish (changes requested):

.github/workflows/markdown-style-checks.yml
minaelee: Done!

.github/workflows/markdown-style-checks.yml
minaelee: Added

roosterfish (dismissed):
Ty!

minaelee:
Updated to include new lint-md target in output from `make help`.

roosterfish (dismissed):

roosterfish:
A fix for the linter is in https://github.com/canonical/microcloud/pull/1369.

roosterfish:
Can be rebased to unblock the merge.

.github/workflows/markdown-style-checks.yml
```diff
@@ -0,0 +1,29 @@
...
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        fetch-depth: 1
+        persist-credentials: false
+    - name: Create venv
+      working-directory: "doc"
+      run: make install
+    - name: Lint markdown
+      working-directory: "doc"
+      run: make lint-md
```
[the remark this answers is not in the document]
roosterfish: I assume the `permissions` is what is being flagged up here. 
Since we are doing a checkout of the code, I assume we would require at least read permissions, same as in https://github.com/canonical/microcloud/blob/main/.github/workflows/tests.yml#L19-L20.

@simondeziel is this the right assumption?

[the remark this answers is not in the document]
simondeziel: @roosterfish yeah, with the slight precision that since our repos are public, `permissions: {}` would also work. 

@minaelee I'd probably go with the usual boiler plate code that should please `zizmor`:
```
permissions:
  contents: read

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

defaults:
  run:
    # Make sure bash is always invoked with `-eo pipefail`
    # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell
    shell: bash
```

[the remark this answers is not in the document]
minaelee: Cheers, thanks. Added!

.github/workflows/markdown-style-checks.yml
```diff
@@ -0,0 +1,42 @@
...
+on:
+  workflow_call:
+  push:
+    branches:
+    - main
+    paths:
+    - 'doc/**'   # Only run on changes to the doc directory
+  pull_request:
+    branches:
+    - '*'
```
simondeziel: This filter doesn't filter out anything so it should probably be removed.

doc/Makefile
```diff
@@ -112,6 +119,35 @@ serve: html
 serve-microcloud: microcloud
 	cd $(current_dir)/_build/; python3 -m http.server --bind 127.0.0.1 8000
 
+pymarkdownlnt-install: install
+	@. $(DOCS_VENV); test -d $(DOCS_VENVDIR)/lib/python*/site-packages/pymarkdown || pip install pymarkdownlnt==0.9.35
```
simondeziel, line 123: Should this new dependency be added to `doc/custom_conf.py` instead?
```
custom_required_modules = [
    'sphinx-sitemap',
    'pyyaml',
    'pymarkdownlnt==0.9.35',
]
```

.github/workflows/markdown-style-checks.yml
```diff
@@ -0,0 +1,42 @@
...
+    paths:
+    - 'doc/**'   # Only run on changes to the doc directory
+
+jobs:
+  markdown-lint:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        fetch-depth: 1
```
simondeziel: Is `fetch-depth` needed here? By default shallow clones are used.

.github/workflows/markdown-style-checks.yml
```diff
@@ -0,0 +1,42 @@
...
+        persist-credentials: false
+    - name: Create venv
+      working-directory: "doc"
+      run: make install
+    - name: Lint markdown
+      working-directory: "doc"
+      run: make lint-md
```
simondeziel, line 40: So this new workflow is to run `make lint-md` but could the existing Documentation checks in the tests workflow be extended instead?

doc/Makefile
minaelee: custom_conf.py is removed in a separate PR, and the linter is not required for the docs build.

.github/workflows/markdown-style-checks.yml
minaelee: The existing docs checks are run from https://github.com/canonical/documentation-workflows/tree/main/.github/workflows which deliberately do not include the markdown linter because the default for Sphinx is reStructuredText (.rst). We could call the linter in the existing tests but that, as I noted above, takes us further out of alignment with the Canonical standard docs tooling instead of toward.

.github/workflows/markdown-style-checks.yml
minaelee: Sure. I can remove them here, but note that any changes to this file will cause it to fall out of alignment with the upstream file, and these changes could also be suggested upstream: https://github.com/canonical/sphinx-stack/blob/main/.github/workflows/markdown-style-checks.yml

.github/workflows/markdown-style-checks.yml
minaelee: Fixed.

doc/Makefile
simondeziel: Ack, thx.

.github/workflows/markdown-style-checks.yml
simondeziel: Ack, thx.

roosterfish (approved):

PR #1368, opened 2026-05-04
doc: docs tooling and configuration updates

minaelee:
Updates to MicroCloud docs tooling and configuration for improvements as well as to align with standard Canonical docs tooling (Sphinx Stack 1.6, formerly known as Sphinx Starter Pack), including:

- Remove build_requirements.py and related code to dynamically build requirements.txt. Static file provided instead.  
- Merge conf_py & custom_conf.py, and Makefile & Makefile.sp, and bring merged versions into alignment with Sphinx Stack. Includes using doc/.venv now instead of doc/.sphinx/venv for virtual environment directory at build time.
- Use Vale for spelling and woke checks; fix newly flagged issues by Vale.
- Use the [canonical-sphinx](https://github.com/canonical/canonical-sphinx) extension while preserving MicroCloud customizations; remove use of canonical-sphinx-extensions.
- Removing canonical-sphinx-extensions required adding back in some extensions previously bundled with it, including sphinx-terminal, which in its un-bundled form has a difference in usage (the `:input:` option is now handled differently). Update usage of `{terminal}` directive.
- To align with Sphinx Stack directory structure, move _templates and _static directories out from doc/.sphinx/ to doc/, and substitutions.yaml from doc/ to doc/reuse.
- Add SINGLE_BUILD=True to `linkcheck` make target to prevent GH CI from warning and exiting linkcheck when it tries and fails to find the integration files.
- Replace `sphinx-reredirects` extension with `sphinx-rerediraffe`, the new default URL redirecter.
- Update configuration paths for MicroCeph.
- Remove RTD search support (RTD search is already disabled in MicroCloud docs due to inconsistent behavior).
- Add ubuntu.com to the linkcheck ignore list (despite being up, the site has been failing every linkcheck recently and I suspect this is due to increased protection measures).
- Add `.sphinx/version` file to record last version of sphinx-stack this docs set was updated against. 
- Removes no longer needed workarounds in .readthedocs.yaml (the workarounds were only needed until PRs in MicroCeph and MicroOVN were merged, and they have been)

PR #1374, opened 2026-05-07
docs: homepage updates

minaelee:
Per https://github.com/canonical/microcloud/pull/1368#discussion_r3201253653, this PR updates the "Related links" section of the MicroCloud home page to display only a link to the product page. I removed the Discourse link (likely added originally before there was a product page) as well as the link to the Snap store (already linked inline in the homepage). Note that if the MicroCloud integration is removed in the future and the MicroCloud docs use the standard header (like LXD's current header), this related link can be removed entirely, since the standard header already includes a link to the product page. MicroCloud's current header is customized for the integration and does not show this.

As long as I was updating the homepage, I took the opportunity to increase its alignment with the standard homepage structure across Canonical docs by adding the required "How this documentation is organized" section. See [LXD's homepage](https://documentation.ubuntu.com/lxd/latest/#how-this-documentation-is-organized) for another example.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -2,8 +2,8 @@
 myst:
   html_meta:
     description: MicroCloud is a lightweight, open source cloud platform built on LXD, MicroCeph, and MicroOVN, ideal for private cloud, edge computing, and test labs.
-discourse: lxc:[Introducing&#32;MicroCloud](15871)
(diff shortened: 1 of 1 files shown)


roosterfish (approved):
Thanks!

PR #1395, opened 2026-05-29
fix: accept newer microceph versions

jamesbeedy:
## Summary
- change MicroCeph version validation from exact major/minor matching to minimum-version comparison

## Why
MicroCloud currently rejects tentacle MicroCeph even though it is newer than 19.2. This blocks deployments that need tentacle for dashboard OAuth2 support.


roosterfish (commented):
Hey, thanks for your PR. Are you aware of https://github.com/canonical/microcloud/pull/1338 which already adds support for the latest version of MicroCeph?

PR #1407, opened 2026-06-09
Init: Explicit MicroOVN warning

roosterfish:
This PR aims to help users understanding when MicroOVN cannot be configured.

An additional warning message is printed which explicitly mentions MicroOVN.
Also the default answer is changed from `true` to `false` so users rather stop the setup instead of falling back silently to FAN networking:

Screenshot from 2026-06-09 10-58-46


mionaalex (approved):
Thank you for adding this. LGTM

tugbataluy (approved):

PR #1421, opened 2026-06-23
demos/terraform: Added support for secure boot

mseralessandri:
## Checklist

### DIFF
--- a/demos/terraform/main.tf
+++ b/demos/terraform/main.tf
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     lxd = {
       source  = "terraform-lxd/lxd"
-      version = "~> 2.5.0"
+      version = "~> 2.6.2"
     }
     random = {
       source  = "hashicorp/random"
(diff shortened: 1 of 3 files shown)


demos/terraform/terraform.tfvars.example
```diff
@@ -21,6 +21,11 @@ session_timeout = 900
 lookup_timeout = 300
 instance_create_timeout = "25m"
 
+# Support security.secureboot for VMs.
+# Leave this disabled unless your host supports the key.
+enable_secureboot_config = false
```
roosterfish, line 26: Can this be one variable? Meaning if `vm_secureboot` is set to false, we don't populate the LXD `security.secureboot` config key? Why do we need an extra `enable_secureboot_config`?

demos/terraform/main.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     lxd = {
       source  = "terraform-lxd/lxd"
-      version = "~> 2.5.0"
+      version = "~> 2.6.2"
```
roosterfish, line 5: Can the update of the terrafrom provider please be in its own commit as it doesn't seem to be related to adding the secure boot option.

roosterfish (changes requested):

demos/terraform/terraform.tfvars.example
mseralessandri: One variable controls whether the secure-boot key is added to the config (`enable_secure_boot_config`), and the other controls its value (`vm_secureboot`, true/false).
However, I realized that LXD’s boot-related instance config keys have changed (https://canonical.com/lxd/docs/latest/reference/instance_options/#boot-related-options), so the current implementation is only valid for LXD 5.21. The issue is not the config keys themselves, but rather the approach for handling different config keys across LXD versions. Since this is not urgent, I’ll close the PR and revisit it later with a cleaner, more maintainable solution.

PR #1422, opened 2026-06-24
demos/terraform: Remove incompatible secure boot option

mseralessandri:
## Checklist
Removed `security.secureboot` from `main.tf` since it is not compatible with [LXD 6.8](https://canonical.com/lxd/docs/latest/reference/instance_options/#instance-boot:boot.mode) 

### DIFF
--- a/demos/terraform/main.tf
+++ b/demos/terraform/main.tf
@@ -90,7 +90,6 @@ resource "lxd_instance" "microcloud" {
   }
 
   config = {
-    "security.secureboot" = "false"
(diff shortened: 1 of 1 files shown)


demos/terraform/main.tf
```diff
@@ -90,7 +90,6 @@ resource "lxd_instance" "microcloud" {
   }
 
   config = {
-    "security.secureboot" = "false"
```
roosterfish, line 93: Was there a reason this was added in the first place? AFAIK the `ubuntu:24.04` which is used as default in the tf var anyway comes with secure boot support.

roosterfish (approved):

PR #1425, opened 2026-06-25
Doc: Remove outdated statement about 3 track

roosterfish:
There are feature releases on the 3 track.

### DIFF
--- a/doc/reference/releases-snaps.md
+++ b/doc/reference/releases-snaps.md
@@ -94,7 +94,7 @@ MicroCloud LTS tracks use the format _x.y_, corresponding to the major and minor
 
 The MicroCloud feature track uses the major number of the current {ref}`feature release <ref-releases-microcloud-feature>` series. Feature releases within the same major version are published to the same track, replacing the previous release. This simplifies updates, as you don't need to switch channels to access new feature releases within the same major version.
 
-The current feature track is {{current_feature_track}}. The most recent development updates can be found in the {{current_feature_track}}/`edge` channel, for testing purposes only.
+The current feature track is {{current_feature_track}}. The most recent development updates can be found in the {{current_feature_track}}/edge channel, for testing purposes only.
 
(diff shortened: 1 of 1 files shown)


doc/reference/releases-snaps.md
```diff
@@ -94,7 +94,7 @@ MicroCloud LTS tracks use the format _x.y_, corresponding to the major and minor
 
 The MicroCloud feature track uses the major number of the current {ref}`feature release <ref-releases-microcloud-feature>` series. Feature releases within the same major version are published to the same track, replacing the previous release. This simplifies updates, as you don't need to switch channels to access new feature releases within the same major version.
 
-The current feature track is {{current_feature_track}}. No feature release has yet been published to this track. The most recent development updates can be found in the {{current_feature_track}}/`edge` channel, for testing purposes only.
+The current feature track is {{current_feature_track}}. The most recent development updates can be found in the {{current_feature_track}}/`edge` channel, for testing purposes only.
```
simondeziel: We usually have the backdicks span the whole channel name like `{{current_feature_track}}/edge`. Is this avoided due to the special treatment needed for the `{{current_feature_track}}` expansion?

simondeziel (dismissed):

doc/reference/releases-snaps.md
roosterfish: Mh good spot, I'll try aligning it to cover all of the channel name.

roosterfish: I haven't found a way to do this that doesn't cause the `{{current_feature_track}}` substitution to break. Therefore I have pushed another commit that doesn't use backticks, similar to a few lines above in the same file.

simondeziel (approved):

doc/reference/releases-snaps.md
simondeziel: Thanks for trying

PR #1444, opened 2026-07-07
doc: update LXD urls (v2-edge)

elijahgreenstein:
Update LXD documentation URLs following the move to `canonical.com/lxd/docs`.

## Checklist

### DIFF
--- a/doc/.sphinx/_integration/add_config.py
+++ b/doc/.sphinx/_integration/add_config.py
@@ -10,7 +10,7 @@ html_context['microovn_path'] = "../microovn"
 html_context['microovn_tag'] = "../microovn/_static/microovn.png"
 
 if project == "LXD":
-    html_baseurl = "https://documentation.ubuntu.com/lxd/stable-5.21/"
+    html_baseurl = "https://canonical.com/lxd/docs/v5.21/"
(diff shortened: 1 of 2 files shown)


doc/.sphinx/_integration/add_config.py
```diff
@@ -10,7 +10,7 @@
 html_context['microovn_tag'] = "../microovn/_static/microovn.png"
 
 if project == "LXD":
-    html_baseurl = "https://documentation.ubuntu.com/lxd/stable-5.21/"
+    html_baseurl = "https://canonical.com/lxd/docs/default/"
```
roosterfish: I guess `default` always points to the latest LTS? So what happens if we create the next LTS, I think for MicroCloud 2 this should stay as 5.21, not 6.

Only MicroCloud 3 should the default to LXD 6 LTS.

roosterfish (changes requested):

doc/.sphinx/_integration/add_config.py
elijahgreenstein: Sure, that makes sense. Updated to point to `v5.21`. This currently redirects to `default`, but, as you noted, that will change in the future.

roosterfish (approved):
Thanks!

PR #1450, opened 2026-07-09
doc: add decommissioning guide

elijahgreenstein:
This PR adds a decommissioning guide to comply with SSDLC requirements.

This guide is based off of the guide to LXD introduced in https://github.com/canonical/lxd/pull/18507, but with several changes:

1. The structure is reorganized: this guide is largely organized around decommissioning an entire cluster. A paragraph at the beginning of the guide instructs user wishing to decommission a single cluster member to first remove the cluster, and provides cross references to relevant pages in the LXD documentation. This paragraph then directs users to proceed to destroy the data on the removed member after removing the cluster.
2. Deleting replicators and cluster links is moved into the "Delete data" section, in order to follow the note about projects and the need to use the `--project` flag in order to delete project-scoped entities.
3. `microcloud` commands are used wherever possible, in place of `lxc` commands.
4. References to MicroCeph logging is included.
5. Details about removing MicroCeph OSDs is included.
6. Instructions about removing storage buckets is removed from this guide, on the assumption that MicroCloud users would not use Ceph Object buckets.
7. Cluster removal includes details about removing the Ceph `monmap` when reducing the cluster down to one member.
8. Snap removal covers MicroCloud, LXD, MicroCeph, and MicroOVN.
9. A note about ACME services is included, alongside the note about OIDC providers, in the final section.

## Checklist

### DIFF
--- a/doc/how-to/decommission.md
+++ b/doc/how-to/decommission.md
@@ -1,0 +1,293 @@
+---
+myst:
+  html_meta:
+    description: Follow these steps to securely decommission a MicroCloud cluster member or cluster.
+---
+
+(howto-decommission)=
+# How to securely decommission a MicroCloud deployment
+
+```{important}
+This process will erase all data associated with your MicroCloud deployment.
+Make copies of any data that you need to preserve before proceeding.
+Refer to {ref}`lxd:instances-backup` and {ref}`lxd:howto-storage-backup-volume` for relevant details.
+```
+
+This guide walks you through the steps to decommission an entire MicroCloud cluster.
+
+If you only need to decommission a single cluster member, first {ref}`remove the member from the cluster <howto-member-remove>`.
+After removing the member, {ref}`update the certificate <lxd:cluster-manage-update-certificate>` on the cluster remaining in production.
+Then, return to this guide, skip ahead to {ref}`howto-decommission-remove-microcloud`, and follow the instructions in the sections that follow.
+
+Some commands used to decommission MicroCloud are LXD or MicroCeph commands.
+Refer to the {ref}`LXD decommissioning guide <lxd:howto-decommission>` and the {ref}`MicroCeph guide to removing disks <microceph:remove-disk>` for related information.
+
+(howto-decommission-remove-offline-member)=
+## Remove offline cluster members
+
+Use the `--force` flag to {ref}`remove any offline cluster members <howto-member-remove>` (you will remove online cluster members later in the process):
+
+```bash
+sudo microcloud remove --force <offline_member_name>
+```
+
+(howto-decommission-revoke-remote)=
+## Revoke remote access
+
+List all identities that have access to LXD, then delete each identity:
+
+```bash
+lxc auth identity list
+lxc auth identity delete <type>/<name_or_identifier>
+```
+
+(howto-decommission-list-projects)=
+## List projects
+
+Replicators, instances, profiles, and custom volumes are scoped by {ref}`project <lxd:projects>`.
+For deployments with more than one project, you must repeat some steps for **each** project, each time using the `--project` flag.
+You do not need to use the `--project` flag to decommission deployments with only one project.
+
+Run this command to get a list of all projects:
+
+```bash
+lxc project list
+```
+
+````{note}
+You can also delete a project (except the `default` project) and all of its project-level entities with:
+
+```bash
+lxc project delete <project_name> --force
+```
+````
+
+(howto-decommission-delete-replicators)=
+## Delete replicators and cluster links
+
+For each project, list all replicators, then delete each replicator:
+
+```bash
+lxc replicator list --project <project_name>
+lxc replicator delete <replicator_name> --project <project_name>
+```
+
+Likewise, list all cluster links, then delete each cluster link (cluster links are not scoped by project, so you do not need to use the `--project` flag):
+
+```bash
+lxc cluster link list
+lxc cluster link delete <cluster_link_name>
+```
+
+(howto-decommission-delete-data)=
+## Delete data
+
+You can run the commands in this section on any online cluster member to delete data.
+
+```{important}
+Data deleted by LXD physically remains on disks and can be recovered by users with access to the disks.
+To prevent unauthorized data recovery, you must {ref}`destroy and sanitize your data <howto-decommission-destroy-data>`.
+```
+
+(howto-decommission-delete-instances)=
+### Stop and delete instances
+
+For each project, stop all instances:
+
+```bash
+lxc stop --all --project <project_name>
+```
+
+Next, for each project, list all instances, then delete each instance:
+
+```bash
+lxc list --project <project_name>
+lxc delete <instance_name> --project <project_name>
+```
+
+If you are unable to stop or delete an instance, use the `--force` flag:
+
+```bash
+lxc stop --force <instance_name> --project <project_name>
+lxc delete --force <instance_name> --project <project_name>
+```
+
+(howto-decommission-delete-profiles)=
+### Delete profiles
+
+For each project, list all profiles:
+
+```bash
+lxc profile list --project <project_name>
+```
+
+Each project has a `default` profile that cannot be deleted.
+Delete all other profiles:
+
+```bash
+lxc profile delete <profile_name> --project <project_name>
+```
+
+(howto-decommission-remove-disk-devices)=
+### Remove disk devices from `default` profiles
+
+You cannot delete a storage pool used by an instance, profile, or custom volume.
+You must, therefore, remove any disk devices used by the `default` profiles in order to delete any storage pools or custom volumes referenced by those devices.
+
+At a minimum, the `default` profile of the `default` project has a disk device named `root` that references a storage pool.
+Remove this device with:
+
+```bash
+lxc profile device remove default root --project default
+```
+
+To check for additional disk devices, view information about the `default` profile of each project:
+
+```bash
+lxc profile show default --project <project_name>
+```
+
+Remove any remaining disk devices that reference storage pools:
+
+```bash
+lxc profile device remove default <device_name> --project <project_name>
+```
+
+(howto-decommission-delete-volumes)=
+### Delete custom volumes
+
+To delete {ref}`custom volumes <lxd:storage-volume-types>`, you must specify the storage pools used by the volumes.
+First, list all storage pools across projects:
+
+```bash
+lxc storage list
+```
+
+Next, for each storage pool, list the custom volumes.
+Use the `--all-projects` flag to view all custom volumes across projects:
+
+```bash
+lxc storage volume list <pool_name> type=custom --all-projects
+```
+
+Use the `PROJECT` column in the output to identify the project associated with each custom volume.
+Then delete each custom volume, specifying both the storage pool and the project:
+
+```bash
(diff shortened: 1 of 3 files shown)


elijahgreenstein:
@mseralessandri , @roosterfish -- I added the section on MicroCeph OSDs based on my reading of the [MicroCloud guide to removing cluster members](https://documentation.ubuntu.com/microcloud/latest/how-to/member_remove/), which included a reference to the [MicroCeph guide to removing disks](https://documentation.ubuntu.com/microcloud/latest/microceph/snap/how-to/remove-disk/). However, would these steps be necessary when decommissioning an entire cluster?

doc/how-to/decommission.md
```diff
@@ -0,0 +1,266 @@
...
+```{important}
+This process will erase all data associated with your MicroCloud deployment.
+Make copies of any data that you need to preserve before proceeding.
+Refer to {ref}`lxd:instances-backup` and {ref}`lxd:howto-storage-backup-volume` for relevant details.
+```
+
+This guide walks you through the steps to decommission an entire MicroCloud cluster.
+
+If you only need to decommission a single cluster member, first {ref}`remove the member from the cluster <howto-member-remove>`.
+After removing the member, {ref}`update the certificate <lxd:cluster-manage-update-certificate>` on the cluster remaining in production.
```
roosterfish, line 19: If we recommend updating the cluster certificate in the reamining cluster for LXD, likely we should also propose doing this for MicroCloud, MicroCeph and MicroOVN (Microclusters).

@mseralessandri I wonder if we should grow support in MicroCloud to generally trigger this across LXD and Micro* from a single MicroCloud command like `microcloud cluster update-certificate`

roosterfish: It seems MicroCeph has a `microceph certificates` command used to set RGW cert. Also MicroOVN comes with a `microovn certificates` command but it also seems to be used solely for OVN services, not Microcluster.

So it seems this is a general gap currently in all Micro* products to be able to rotate the underlying Microcluster certificate with straightforward commands. This is supported by Microcluster but not directly exposed. 
In case a user would hit this today, we can instruct using curl against the Microcluster API though.

roosterfish: I have added https://github.com/canonical/microcloud/issues/1453 with a note to update this guide.

doc/how-to/decommission.md
```diff
@@ -0,0 +1,266 @@
...
+
+```bash
+lxc storage volume delete <pool_name> <volume_name> --project <project_name>
+```
+
+### Delete storage pools
+
+Storage pools cannot be deleted if they are used by an instance, profile, or custom volume.
+The `default` profile cannot be deleted; therefore, the storage pool used by the `default` profile cannot be deleted.
+To identify this storage pool, view information about the `default` profile and find the pool listed under `devices` > `root` > `pool`:
```
roosterfish: This is not true. LXD blocks the pool deletion because the `default` profile has a `disk` device which references the pool. You can remove this device which will then allow you to remove the pool:

```bash
root@micro01:~# lxc storage rm remote
Error: The storage pool is currently in use
root@micro01:~# lxc profile show default
name: default
description: Default LXD profile
config: {}
devices:
  eth0:
    name: eth0
    network: default
    type: nic
  root:
    path: /
    pool: remote
    type: disk
used_by: []
project: default
root@micro01:~# lxc profile device rm default root
root@micro01:~# lxc storage rm remote
```

doc/how-to/decommission.md
```diff
@@ -0,0 +1,266 @@
...
+
+
+(howto-decommission-remove-microceph-osds)=
+## Remove MicroCeph OSDs
+
+To {ref}`remove MicroCeph OSDs <microceph:remove-disk>`, you need to determine the OSD IDs.
+Run this command to view the Ceph OSD tree:
+
+```
+ceph osd tree
```
roosterfish: Any reason for not using a MicroCeph command directly?

```bash
root@micro01:~# microceph disk list
Disks configured in MicroCeph:
+-----+----------+----------------------------------------------------+
| OSD | LOCATION |                        PATH                        |
+-----+----------+----------------------------------------------------+
| 1   | micro01  | /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_disk2 |
+-----+----------+----------------------------------------------------+
```

There is also a `--host-only` flag to return only OSDs configured on this member (requires running the command on the to be decommissioned member).

doc/how-to/decommission.md
```diff
@@ -0,0 +1,266 @@
...
+```bash
+microcloud cluster list
+sudo microcloud remove <member_name>
+```
+
+````{note}
+To {ref}`reduce the cluster down to one member <howto-member-remove-reduce-cluster>`, you must first clean up the Ceph monitor map (`monmap`) while there are still two cluster members:
+
+```bash
+sudo microceph.ceph mon remove <name>
```
roosterfish: Should we instead link to the doc page where those commands are already listed? Just to prevent duplicating this.

doc/how-to/decommission.md
```diff
@@ -0,0 +1,266 @@
...
+## Remove snaps
+
+```{important}
+Run these commands on **every** machine that you decommission.
+
+Removing MicroCloud **does not** erase {ref}`ZFS pools (zpools) <lxd:storage-zfs>` or dedicated disks used by MicroCeph as Ceph object storage daemons (OSDs).
+To securely decommission MicroCloud, you must {ref}`destroy and sanitize your data <howto-decommission-destroy-data>`.
+```
+
+Remove the MicroCloud, LXD, MicroCeph, and MicroOVN snaps.
```
roosterfish, line 258: Worth adding a command that MicroCeph and MicroOVN might not be installed if the user chose to deploy a MicroCloud without those components?

doc/how-to/decommission.md
```diff
@@ -0,0 +1,266 @@
...
+sudo snap remove lxd --purge
+sudo snap remove microceph --purge
+sudo snap remove microovn --purge
+```
+
+Verify that the snaps and associated data were removed.
+The following commands should report that none of these snaps are installed and that the `/var/snap/microcloud/`, `/var/snap/lxd/`, `/var/snap/microceph/`, and `/var/snap/microovn` directories do not exist:
+
+```bash
+snap list microcloud lxd microceph microovn
```
roosterfish, line 276: Worth running `microcloud status` on any of the remaining members to verify everything got cleaned up and the member is gone?

roosterfish (changes requested):
Thanks, left a few comments.

doc/how-to/decommission.md
elijahgreenstein: Ok, so leaving this as-is for now, with an update planned after #1453 is addressed.

doc/how-to/decommission.md
elijahgreenstein: Nice, it was bothering me to leave that profile and its associated storage pool untouched. I've updated the section with the instructions to remove the device.

Two notes:
1. I changed the command to `lxc profile device remove` (rather than `rm`), since that is what appears in the man pages.
2. It seems like the device is always named `root`. Is that correct? I've used that term in the instructions, but, if necessary, I can update the instructions to instruct users to find the device name.

doc/how-to/decommission.md
elijahgreenstein: Thanks -- I was following the instructions in the MicroCeph [Remove a disk](https://documentation.ubuntu.com/microceph/latest/snap/how-to/remove-disk/) guide, but that is a much simpler way to do things and a big improvement to the instructions.

Updated

elijahgreenstein: I think `--host-only` would be useful if the instructions were specific to removing a single cluster member. At the moment, the top of this decommissioning guide instructs users to follow our MicroCloud "Remove a member" guide when decommissioning a single member. That guide, in turn links over to the MicroCeph docs for the OSD removal. If we ever incorporated the MicroCeph OSD removal into our "Remove a member" guide, possibly we could include the `--host-only` flag. But I don't think we need it here, since we're removing all OSDs at once.

doc/how-to/decommission.md
elijahgreenstein: That makes sense. Done.

doc/how-to/decommission.md
elijahgreenstein: Seems like it wouldn't hurt to add that. I'll include it as a note.

Seems like it's also worth adding to the "How to remove a member" guide. I've updated it too.

doc/how-to/decommission.md
elijahgreenstein: Thanks, updated.

doc/how-to/decommission.md
elijahgreenstein: Another note: it seems as if the `default` profile is always associated with the `default` project. Is that correct? I've added a flag with the goal of making the command work regardless of the currently active project:

```
lxc profile show default --project default
```

elijahgreenstein:
Updated based on @roosterfish 's comments. I've also changed the title that appears in the navigation menu from "Decommission MicroCloud" to "Decommission a MicroCloud." Wondering what your thoughts on this are. I've also tried out "Decommission a cluster or cluster member."

elijahgreenstein:
Made an additional update to direct users decommissioning a single cluster member to proceed to the "Remove the snaps" section after removing the member, rather than skipping all the way to "Destroy and sanitize data".

doc/how-to/decommission.md
roosterfish: Profiles belong to projects. When you create a new project, it also creates a new profile `default`. So it could happen that you have multiple projects, and each has their own `default` profile.

> It seems like the device is always named root. Is that correct?

When it comes to MicroCloud, we configure the `default` profile in the `default` project to with two devices. One called `eth0` (network) and the other called [`root`](https://github.com/canonical/microcloud/blob/main/cmd/microcloud/main_init.go#L712). On top of this the user can ofc create additional entities we are not aware of when it comes to decommissioning.

doc/how-to/decommission.md
roosterfish: Makes sense thanks.

doc/how-to/decommission.md
elijahgreenstein: Thanks for the clarification. I've updated the guide as follows to address this by adding a subsection on removing disk devices from `default` profiles --- I am assuming here that the issue centers on **disk** devices. If there are other devices that would be of concern, let me know, but I currently understand the issue as:
- `default` profiles cannot be deleted
-  `default` profiles *may* have disk devices that reference storage pools (and the `default` profile in the `default` project will *definitely* have that `root` device).
- Disk devices reference storage pools and possibly custom volumes (in the case of the `root` device, it does not have a value for `source` as would be the case in a custom volume attached to the profile)
- Hence: users must remove **disk** devices from the `default` profiles; if those reference custom volumes, the volumes will be deleted in the next section.

elijahgreenstein:
Updated to address @roosterfish 's comments about the `default` profiles and the `root` device.

Also reorganized following a conversation with @mseralessandri to pull the "List projects" and the "Delete replicators and cluster links" subsections out of the "Delete data" section, and reorganize them *above* "Delete data".

roosterfish (approved):
LGTM thanks!

mseralessandri (approved):
LGTM, thanks!

PR #1481, opened 2026-08-10
Backports (v2-edge)

roosterfish:
Backports until c7de0c05f397ab77caeb14e9adc61f07a55065f8.

Missing only https://github.com/canonical/microcloud/pull/1344 (depending on MicroCeph doc updates, non blocking, just a reference for the next backports). This one is already carried since the last few backports.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -75,9 +75,9 @@ html_context = {
 }
 
 # Enables the pencil icon to edit pages on GitHub, shown at the top of each page
-html_theme_options = {
-    'source_edit_link': html_context['github_url']
-}
+# html_theme_options = {
+#     'source_edit_link': html_context['github_url']
+# }
 
 #######################
 # Sitemap configuration: https://sphinx-sitemap.readthedocs.io/
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -75,9 +75,9 @@ html_context = {
 }
 
 # Enables the pencil icon to edit pages on GitHub, shown at the top of each page
-# html_theme_options = {
-#     'source_edit_link': html_context['github_url']
-# }
+html_theme_options = {
+    'source_edit_link': html_context['github_url']
+}
 
 #######################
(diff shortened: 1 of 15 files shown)


doc/conf.py
```diff
@@ -132,10 +132,14 @@
...
     # These links may fail from time to time
     'https://ceph.io',
+    # Rejects us always with a "Connection aborted"
+    r'https://charmhub\.io/.*',
     # Cloudflare protection on SourceForge domains often block linkcheck
     r'https://.*\.sourceforge\.(net|io)/.*',
     # These links often fail due to infra issues or protection measures 
     r'https://ubuntu\.com.*',
+    r'https://canonical\.com.*',
+    r'https://snapcraft\.io.*',
```
elijahgreenstein, line 142: If v2-edge docs have been building successfully since we introduced these links to the linkcheck_ignore in main, then I'm wondering if this backport is necessary. Likewise, we could consider removing these from main to see if the link check works on these URLs again. The infra issues / protection measures on the canonical.com / snapcraft.io -side that was causing the link check to fail may have been resolved since we introduced these back in June. I'm less familiar with the charmhub.io case, but it may have been the same issue.

elijahgreenstein (commented):
Hey @roosterfish , I already backported both the "edit" button commits as part of the migration process, in PRs #1398 and #1458, so they may be unneeded in this PR.

I also have a comment below about the `linkcheck_ignore` updates.

doc/conf.py
roosterfish: I am trying to keep `v2-edge` as close as possible to `main`. This causes less friction with future backports as the commits tend to apply cleanly this way.
 
We can revert this in `main` first and then backport it to `v2-edge`. But we should never start experimenting in `v2-edge` first. There is an exception for cases where the change isn't viable to be in `main`.

elijahgreenstein: That makes sense. We can test out the linkchecker separately in main, then.

elijahgreenstein (approved):

PR #1345, opened 2026-04-20
Doc: 3.2 release notes

roosterfish:
Adds the release notes for the upcoming 3.2 release.

PR #1343, opened 2026-04-20
Doc: 2.1.3 release notes

roosterfish:
Adds the release notes for the upcoming 2.1.3 release.

PR #1482, opened 2026-08-10
Backport MicroCloud 3 release notes to MicroCloud 2 docs (v2-edge)

elijahgreenstein:
This PR:

- backports the release notes for MicroCloud 3.1 and MicroCloud 3.2 to the documentation for MicroCloud 2 (from PRs https://github.com/canonical/microcloud/pull/1274, https://github.com/canonical/microcloud/pull/1343, and https://github.com/canonical/microcloud/pull/1345), and
- updates the MicroCloud 2 release notes TOC to use the pages, rather than links out to MicroCloud 3 (`latest`) docs.

Motivation: if a user is viewing the `default` version of the docs and navigates to the 3.1 or 3.2 release notes, then the user jumps to the `latest` version. If the user then uses the version selector flyout menu to navigate back to `default`, the user lands on a 404 page.

We could, alternatively, put in place a redirect for the `/reference/release-notes/release-notes-3.{1|2}/` path back to `latest`, but if that additional step of adding a redirect is required anyway, it seems simpler to backport the MicroCloud 3 release notes. This also ensures that users do not unintentionally switch from the `default` to `latest` docs versions.

### DIFF
--- a/doc/reference/release-notes/index.md
+++ b/doc/reference/release-notes/index.md
@@ -25,8 +25,8 @@ For full instructions on updating or upgrading MicroCloud, refer to {ref}`howto-
 
 ```{toctree}
 :titlesonly:
-MicroCloud 3.2 <https://documentation.ubuntu.com/microcloud/latest/reference/release-notes/release-notes-3.2/>
-MicroCloud 3.1 <https://documentation.ubuntu.com/microcloud/latest/reference/release-notes/release-notes-3.1/>
+MicroCloud 3.2 </reference/release-notes/release-notes-3.2>
+MicroCloud 3.1 </reference/release-notes/release-notes-3.1>
 MicroCloud 2.1.3 LTS </reference/release-notes/release-notes-2.1.3>
 MicroCloud 2.1.2 LTS <https://discourse.ubuntu.com/t/microcloud-2-1-2-lts-has-been-released/73771/2>
 MicroCloud 2.1.1 LTS <https://discourse.ubuntu.com/t/microcloud-2-1-2-lts-has-been-released/73771>
--- a/doc/reference/release-notes/release-notes-3.1.md
+++ b/doc/reference/release-notes/release-notes-3.1.md
@@ -29,7 +29,7 @@ how to enroll a MicroCloud cluster.
 
 ### Temporary UI access link
 
-It's fast to deploy a MicroCloud and gain the ability to spawn resources with it from the command line.
+It's fast to deploy a MicroCloud and gain the ability to spawn resources from the command line.
 To provide the same experience when {ref}`using the UI <howto-ui>`, we added a new question to the end of the
 interactive `microcloud init` questionnaire:
 
@@ -37,7 +37,7 @@ interactive `microcloud init` questionnaire:
 
 If you answer `yes` (default), MicroCloud creates a temporary UI access link that allows you to immediately enter
 the LXD UI after MicroCloud is installed.
-The temporary access lasts for 24 hours and is intended to perform further setups (such as configuring permanent access).
+The temporary access lasts for 24 hours and is intended to allow you to perform further setup (such as configuring permanent access).
 
 This requires a 6+ version of LXD, which is currently a feature track (non-LTS).
 
--- a/doc/reference/release-notes/release-notes-3.1.md
+++ b/doc/reference/release-notes/release-notes-3.1.md
@@ -1,0 +1,63 @@
+---
+myst:
+  html_meta:
+    description: Release notes for MicroCloud 3.1, including highlights about new features, bugfixes, and other updates from the MicroCloud project.
+---
+
+(ref-release-notes-3.1)=
+# MicroCloud 3.1 release notes
+
+This is a {ref}`feature release <ref-releases-microcloud-feature>` and is not recommended for production use.
+It's the first feature release of the new track 3.
+
+(ref-release-notes-3.1-highlights)=
+## Highlights
+
+This section highlights new and improved features in this release.
+
+### Cluster Manager
+
+Cluster Manager allows you to manage and monitor multiple MicroCloud clusters in a single application.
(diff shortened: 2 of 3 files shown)


elijahgreenstein:
@roosterfish -- another option is to simply limit the MicroCloud 2 docs release notes to MicroCloud 2 and earlier. This is the approach in Juju and Snapcraft for example:

- Snapcraft 8 docs only provide release notes up through 8: https://ubuntu.com/docs/snapcraft/8/release-notes/
- Snapcraft 9 docs provide release notes up through 9: https://ubuntu.com/docs/snapcraft/9/release-notes/
- Juju 2.9 provides release notes for 2.9 and earlier: https://canonical.com/juju/docs/juju-cli/2.9/
- Juju 3.6 provides release notes for 3.6, and < 2.9 (does not link to 2.9 itself): https://canonical.com/juju/docs/juju-cli/3.6/releasenotes/

elijahgreenstein:
RTD build is failing, since the release notes include links to MicroCloud 3 content that is unavailable in MicroCloud 2. Possibly further supports the approach of limiting release notes to past versions, rather than adding future versions. Or adding redirects to `latest`.

roosterfish:
> RTD build is failing, since the release notes include links to MicroCloud 3 content that is unavailable in MicroCloud 2. Possibly further supports the approach of limiting release notes to past versions, rather than adding future versions. Or adding redirects to `latest`.

So should we continue with the current approach of simply linking to the versions > the current/present version?

elijahgreenstein:
>> RTD build is failing, since the release notes include links to MicroCloud 3 content that is unavailable in MicroCloud 2. Possibly further supports the approach of limiting release notes to past versions, rather than adding future versions. Or adding redirects to latest.
>
> So should we continue with the current approach of simply linking to the versions > the current/present version?

That's one option, but then we'd need to add redirects in MicroCloud 2 docs to send users trying to access those release notes to MicroCloud 3 docs. That's hardly a big maintenance problem, but it does mean that once a user is on a MicroCloud 3 release note page, they won't be able to use the version selector to switch back to MicroCloud 2 docs --- they'll keep getting looped back to MicroCloud 3 docs. To change versions, they'll have to navigate to another page, and then switch versions, which might not be that intuitive. That set up would also mean that a user could end up switching docs versions without realizing it.

What do you think about the approach of only including release notes up through MicroCloud 2 series and earlier in MicroCloud 2, and adding a note on the [Release notes index](https://canonical.com/microcloud/docs/default/reference/release-notes/) along the lines of: "Refer to the [latest version of the documentation](https://canonical.com/microcloud/docs/latest/reference/release-notes/) for MicroCloud 3 release notes." ? The MicroCloud 3 release notes won't show up in the sidebar navigation menu anymore in MicroCloud 2 docs, but there will at least be a signpost in the MicroCloud 2 docs that directs users to the MicroCloud 3 release notes.

If we want the MicroCloud 3 release notes in the sidebar menu, we could also include an entry in the TOC to the MicroCloud 3 release notes index, and clearly mark it as "latest" -- i.e. "MicroCloud 3 (latest)". Since this would link to the release notes index page, and the same page exists in MicroCloud 2 docs, it wouldn't create that loop effect locking users into the `latest` docs version.

elijahgreenstein:
@roosterfish -- another option I thought of to avoid locking the user in `latest`: we could add redirects for the MicroCloud 3 release note URLs in MicroCloud 2, but instead of redirecting to the release note page in `latest`, we could redirect to the `default` release notes index page. So, if a user is on the release notes page for MicroCloud 3.1, and switches the version to `default` (MicroCloud 2), the user lands on that `default`  index page.

This does create the possibility for an odd path through the docs: `default` release notes index, click on MicroCloud 3.1 -> `latest` page for MicroCloud 3.1 release notes, click to switch versions -> back to `default` release notes index. This option also means that a user may unintentionally switch versions, from `default` to `latest`. But at least this option prevents the user from landing on a 404 page or getting "stuck" on `latest` if they're on the MicroCloud 3.1 release notes page, and the user can also see all of the release notes while in the `default` docs.

What do you think about this option, versus the option I described above that would limit the `default` docs release notes to MicroCloud 2 and earlier?

roosterfish:
I have the feeling the more redirects we can omit, the better. Is it a technical limitation that the 404 page cannot have a version picker? So it could be a more convenient shortcut to "go back" as the user would simply revert its previous action.

But in general, should there always be a guarantee that when you are visiting page x and switch the version to something older, that you can never end up in a situation in which you see a 404. And instead we always account for this by having redirects? 
For example when I view [cluster manager architecture](https://canonical.com/microcloud/docs/latest/reference/cluster-manager-architecture/) which is only in v3, there isn't any possible redirect target in v2. So in this case we have to show the 404. 
This makes we wonder if we shouldn't accept the same also for the release notes.

elijahgreenstein:
> I have the feeling the more redirects we can omit, the better.

I agree.

> Is it a technical limitation that the 404 page cannot have a version picker? So it could be a more convenient shortcut to "go back" as the user would simply revert its previous action.

It's technically possible, I believe, but in this case it would essentially serve as a "back" button, so I'm not sure if it would be worth the modification and departure from how we set up other docs sets.

> But in general, should there always be a guarantee that when you are visiting page x and switch the version to something older, that you can never end up in a situation in which you see a 404.

I agree on this, and the Cluster Manager is a good example. Not every page in MicroCloud 3 docs needs a redirect if it doesn't exist in MicroCloud 2.

The reason that the release notes might be a different case, is that we currently link to the MicroCloud 3.1 and 3.2 release notes from *within* the MicroCloud 2 docs. That's what inclined me to make some sort of adjustment to the release notes organization or add redirects. But, you're also right that another option is to just let those pages go to 404s.

Based on this discussion, let's leave the release notes organization as-is for now, and refrain from redirects. I'll continue to monitor if the release notes are a frequent source of 404s, and we can revisit if it seems to be a persistent problem.

PR #1487, opened 2026-08-12
Test: Wait in e2e test for restarted intances

roosterfish:
In https://github.com/canonical/microcloud/actions/runs/31585772271/job/94079902223 I saw that it sometimes might take the instances longer than the 20s which are currently in place to get an IPv4 address. 

Instead of bumping to 30s, we wait until they actually got an IP which might also speed up the test sometimes instead of always waiting 20s. But its capped at 30s max.

### DIFF
--- a/test/e2e/run
+++ b/test/e2e/run
@@ -289,14 +289,23 @@ evacuation() {
         fi
 
         lxc cluster restore --project e2e-testing --force "${REMOTE}:${member}"
-    done
 
-    echo -n "Allow time for stopped instances to start back again "
-    for _ in $(seq 20); do
-        echo -n "."
-        sleep 1
+        echo -n "Allow time for stopped instances on ${member} to start back again and have an IPv4 address"
+        for instance in $(instanceListByMember "${member}"); do
+            attempts=0
+            while [ -z "$(getIPv4 "${instance}")" ]; do
+                echo -n "."
+                sleep 1
+                attempts=$((attempts + 1))
+                if [ "${attempts}" -ge 30 ]; then
+                    echo " FAILED"
+                    exit 1
+                fi
+            done
+        done
+
+        echo " DONE"
     done
-    echo " DONE"
 }
 
 
--- a/test/e2e/run
+++ b/test/e2e/run
@@ -202,6 +202,11 @@ instanceList() {
     lxc list --project e2e-testing -f csv -c n "${REMOTE}:"
 }
 
+instanceListByMember() {
+    local member="${1}"
+    lxc list --project e2e-testing -f csv -c n "${REMOTE}:" location="${member}"
+}
+
 clusterMembers() {
     local state="${1}"
     [ "${state}" = "ALL" ] && state=".*"


test/e2e/run
```diff
@@ -202,6 +202,11 @@ instanceList() {
     lxc list --project e2e-testing -f csv -c n "${REMOTE}:"
 }
 
+instanceListByMember() {
+    local member="${1}"
+    lxc list --project e2e-testing -f csv -c nL "${REMOTE}:" | awk -F, "/,${member}/ {print \$1}"
+}
```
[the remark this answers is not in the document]
roosterfish: Fixed.

test/e2e/run
```diff
@@ -202,6 +202,11 @@ instanceList() {
     lxc list --project e2e-testing -f csv -c n "${REMOTE}:"
 }
 
+instanceListByMember() {
+    local member="${1}"
+    lxc list --project e2e-testing -f csv -c nL "${REMOTE}:" | awk -F, -v member="${member}" '$2 == member {print $1}'
```
simondeziel: I think native `lxc list` filtering by location could also have been used here.

simondeziel (dismissed):

test/e2e/run
roosterfish: Thanks for the suggestion, wasn't aware that we can do `location=x` in the `lxc ls` command. Fixed.

simondeziel (dismissed):
LGTM, thanks!

test/e2e/run
```diff
@@ -289,14 +289,22 @@ evacuation() {
...
 
         lxc cluster restore --project e2e-testing --force "${REMOTE}:${member}"
-    done
 
-    echo -n "Allow time for stopped instances to start back again "
-    for _ in $(seq 20); do
-        echo -n "."
-        sleep 1
+        echo -n "Allow time for stopped instances on ${member} to start back again and have an IPv4 address"
+        for instance in $(instanceListByMember "${member}"); do
```
tugbataluy, line 294: The script runs under `set -eu -o pipefail`, but a command substitution in a `for` word list is one of the places `set -e` does not cover; if `lxc list` errors here (which is plausible right after a `cluster restore`), the loop just gets an empty list, the wait is skipped entirely, and the test carries on as if everything were ready:

```console
$ bash -c 'set -eu -o pipefail; for i in $(exit 3); do echo body; done; echo "no abort"'
no abort
```

`connectivity()` already avoids this by assigning first (`instances="$(instanceList)"`)  could we do the same here?

```bash
local instances
instances="$(instanceListByMember "${member}")"
for instance in ${instances}; do
```

That way a failing `lxc list` aborts the run instead of quietly disarming the new check.

test/e2e/run
```diff
@@ -289,14 +289,22 @@ evacuation() {
...
+            attempts=0
+            while [ -z "$(getIPv4 "${instance}")" ]; do
+                echo -n "."
+                sleep 1
+                attempts=$((attempts + 1))
+                if [ "${attempts}" -ge 30 ]; then
+                    echo " FAILED"
```
tugbataluy, line 301:
```bash
echo " FAILED"
exit 1
```

From a CI log this gives us `FAILED` and nothing else. Since the whole point is to diagnose slow-to-address instances, could we name the instance (and send it to stderr)?

```bash
echo " FAILED: ${instance} has no IPv4 address after ${attempts} attempts" >&2
exit 1
```

test/e2e/run
```diff
@@ -289,14 +289,22 @@ evacuation() {
...
+                    exit 1
+                fi
+            done
+            echo " DONE"
```
tugbataluy: The `echo -n "Allow time for stopped instances on ${member} ..."` runs once per member, but `echo " DONE"` runs once per instance, so with several instances per member the output ends up as:

```
Allow time for stopped instances on micro1 to start back again and have an IPv4 address DONE
 DONE
 DONE
```

…where lines 2+ have no context. And if a member has no instances, that header line never gets a terminating newline and the next output is appended to it.

Simplest fix is to move `DONE` out of the inner loop (one per member), or move the header in and include the instance name. Purely cosmetic, but the logs are what we read when this fails.

test/e2e/run
roosterfish: The format already is the following with one `.` added on every retry:

```
Allow time for stopped instances on ${member} to start back again and have an IPv4 address ......... (DONE|FAILED)
```

I don't think we should make the message any longer in this case.

test/e2e/run
roosterfish: AFAIK evacuating the member shouldn't influence the behavior of `lxc ls` as the member itself is still part of the cluster.
The actual reboot of the member happened before the restore so we can be sure at this stage the the member is up and will respond to `lxc ls`.

test/e2e/run
roosterfish: Nice one thanks, I didn't spot it in the test output as we run with three instances evenly distributed across three cluster members so it will always only log a single `DONE` per member.

roosterfish: Fixed.

test/e2e/run
tugbataluy: That's fair; you're right that the member is back up by the time we reach the restore, so the scenario I had in mind isn't really a risk here.

My concern was less about how likely `lxc list` is to fail and more about what happens if it ever does: because the command substitution sits in the `for` word list, a non-zero exit doesn't trip `set -e`, so we'd get an empty list, skip the wait entirely, and continue as if every instance were ready. The check would disappear rather than fail. `connectivity()` sidesteps this just by assigning first (`instances="$(instanceList)"`, `test/e2e/run:259`),
which is why it caught my eye.

Genuinely non-blocking though; happy to leave it as is if you'd rather keep the diff small.

test/e2e/run
tugbataluy: Agreed, and with `DONE` moved out of the inner loop in `2dd3c2e` that line reads much better than what I was looking at, no argument on keeping the message short :).

tugbataluy (approved):
LGTM, thanks 🚀

test/e2e/run
roosterfish: That is ok, afterwards we run actual connections checks which would then fail anyway when the instances are not ready. 

The purpose of this PR is to accommodate slow runners envs where the instances take a long time to come up and get and address. 
So instead of just bumping the number of retries, I have modified the code to wait until the instances have an address, so that the later connection checks don't fail just because of that.

PR #1500, opened 2026-08-13
test/e2e/run: Adjust tests to make script and TF plan work in offline environments

ggouzi:
In offline environments, the e2e tests do not work because we relied on:
- A default hardcoded remote `ubuntu-minimal-daily`
- A default hardcoded project/profile

In offline envs, we have an additional remote pointing to custom simplestreams LXD mirrors. We also have custom profiles with `cloud-init-user-data` so the images apt sources and snap proxy are overridden to point to local mirrors. We may also share images across projects (`features.images=false`)

This PR aims to tackle down those changes to make the script and TF plan work in offline environments. Tested and confirmed to be working in an offline production environment.

Added the following variables
- `image_remote` (string)
- `image_project` (string)
- `copy_image_aliases` (bool)
- `project` (string)
-  `create_project` (bool) 
- `profile` (string)
- `create_profile` (bool)

- ## Checklist

- [X] I have read the [contributing guidelines](https://github.com/canonical/microcloud/blob/main/CONTRIBUTING.md) and attest that all commits in this PR are [signed off](https://github.com/canonical/microcloud/blob/main/CONTRIBUTING.md#including-a-signed-off-by-line-in-your-commits), [cryptographically signed](https://github.com/canonical/microcloud/blob/main/CONTRIBUTING.md#commit-signature-verification), and follow this project's [commit structure](https://github.com/canonical/microcloud/blob/main/CONTRIBUTING.md#commit-structure).
- [X] I have checked and added or updated relevant documentation.

### DIFF
--- a/test/e2e/main.tf
+++ b/test/e2e/main.tf
@@ -3,17 +3,31 @@ data "lxd_info" "cluster" {
   remote = var.remote
 }
 
-# Project
+# Check whether the project already exists
+data "external" "project_exists" {
+  program = ["bash", "-c", "lxc project list \"${var.remote}:\" -f csv | awk -F, '{gsub(/ \\(current\\)$/, \"\", $1); print $1}' | grep -qxF \"${var.project}\" && echo '{\"exists\": \"true\"}' || echo '{\"exists\": \"false\"}'"]
+}
+
+# Check whether the profile already exists
+data "external" "profile_exists" {
+  program = ["bash", "-c", "lxc profile list --project \"${var.project}\" \"${var.remote}:\" -f csv 2>/dev/null | awk -F, '{gsub(/ \\(current\\)$/, \"\", $1); print $1}' | grep -qxF \"${var.profile}\" && echo '{\"exists\": \"true\"}' || echo '{\"exists\": \"false\"}'"]
+}
+
+# Project - created only when it does not already exist
 resource "lxd_project" "e2e" {
-  name   = "e2e-testing"
+  count  = data.external.project_exists.result["exists"] == "true" ? 0 : 1
+  name   = var.project
   remote = var.remote
 }
 
-# Profile
+# Profile - created only when it does not already exist
 resource "lxd_profile" "e2e" {
-  name    = "e2e-testing"
-  project = lxd_project.e2e.name
-  remote  = lxd_project.e2e.remote
+  count   = data.external.profile_exists.result["exists"] == "true" ? 0 : 1
+  name    = var.profile
+  project = var.project
+  remote  = var.remote
+
+  depends_on = [lxd_project.e2e]
 
   # Configuration
   config = {
@@ -44,20 +58,24 @@ resource "lxd_profile" "e2e" {
 # Images
 resource "lxd_cached_image" "ctn" {
   count         = var.containers_per_host > 0 ? 1 : 0
-  project       = lxd_project.e2e.name
-  remote        = lxd_project.e2e.remote
-  source_remote = "ubuntu-minimal-daily"
+  project       = coalesce(var.image_project, var.project)
+  remote        = var.remote
+  source_remote = var.image_remote
   source_image  = "24.04"
   type          = "container"
+  copy_aliases  = var.copy_image_aliases
+  depends_on    = [lxd_project.e2e]
 }
 
 resource "lxd_cached_image" "vm" {
   count         = var.vms_per_host > 0 ? 1 : 0
-  project       = lxd_project.e2e.name
-  remote        = lxd_project.e2e.remote
-  source_remote = "ubuntu-minimal-daily"
+  project       = coalesce(var.image_project, var.project)
+  remote        = var.remote
+  source_remote = var.image_remote
   source_image  = "24.04"
   type          = "virtual-machine"
+  copy_aliases  = var.copy_image_aliases
+  depends_on    = [lxd_project.e2e]
 }
 
 # Containers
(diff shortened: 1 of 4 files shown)


test/e2e/variables.tf
```diff
@@ -7,6 +7,48 @@ variable "remote" {
...
+  default     = false
+}
+
+variable "project" {
+  description = "LXD project to use for e2e testing resources"
+  type        = string
+  default     = "e2e-testing"
+}
+
+variable "create_project" {
```
roosterfish: Same for project and profile: Can we please check if the project exists and if not simply create it? 
Otherwise when you want to reuse an existing one, you have to not only provide its name, but also have to set `create_*` var to `false`.

test/e2e/main.tf
```diff
@@ -5,15 +5,18 @@ data "lxd_info" "cluster" {
 
 # Project
 resource "lxd_project" "e2e" {
-  name   = "e2e-testing"
+  count  = var.create_project ? 1 : 0
```
roosterfish: Instead let's check whether or not the project/profile already exists.

test/e2e/main.tf
```diff
@@ -5,15 +5,18 @@ data "lxd_info" "cluster" {
...
   remote = var.remote
 }
 
 # Profile
 resource "lxd_profile" "e2e" {
-  name    = "e2e-testing"
-  project = lxd_project.e2e.name
-  remote  = lxd_project.e2e.remote
+  count      = var.create_profile ? 1 : 0
+  name       = var.profile
```
roosterfish: I guess we can still derive the project/remote from the `lxd_project`? It's id doesn't change, only its display name based on the var.

test/e2e/main.tf
```diff
@@ -69,11 +76,12 @@ resource "lxd_instance" "e2e-ctn" {
...
   type             = "container"
-  remote           = lxd_project.e2e.remote
-  project          = lxd_project.e2e.name
-  profiles         = [lxd_profile.e2e.name]
+  remote           = var.remote
+  project          = var.project
+  profiles         = [var.profile]
   image            = lxd_cached_image.ctn[0].fingerprint
   wait_for_network = true
+  depends_on       = [lxd_project.e2e, lxd_profile.e2e]
```
roosterfish, line 95: Is there a particular reason for switching from `lxd_project.e2e.*` to `var` so it requires the `depends_on`?

test/e2e/terraform.tfvars.custom-image-server.example
```diff
@@ -0,0 +1,24 @@
+# LXD remote corresponding to the MicroCloud deployment to test.
```
roosterfish, line 1: I wonder instead of "offline", this should be labeled differently to indicate this can be used with a custom image server? Might be also a use case for online environments which want to use a different image server.

roosterfish (changes requested):
Thanks for this, please have a look at my comments.

Also please make sure the commits are squashed accordingly, I saw there were some from the copilot suggestions I assume. They should be integrated with the actual code changes.
It might also help to commit the changes in smaller chunks. One commit that adds new vars, one that uses it and so on.

roosterfish:
Please also `commit -s` to ensure each commit as a `Signed-off-by` line to pass the DCO stage.

test/e2e/main.tf
ggouzi: Checking if profile/project exist implies using local-exec bash commands no ? Or is there a better way to check ?

roosterfish: My thinking was you could maybe use the profile or [project](https://registry.terraform.io/providers/terraform-lxd/lxd/latest/docs/resources/project) resource to check if they exist?

test/e2e/main.tf
ggouzi: Consequence of using count on `lxd_project.e2e.` to see if the project needs to be created or not.

When count = 0 (existing project), `lxd_project.e2e` is an empty list, so referencing `lxd_project.e2e.name` directly would fail.
We would need to use `lxd_project.e2e[0].name`, but that errors when count = 0 :/

roosterfish: Ah I see, thanks for clarifying.

test/e2e/main.tf
ggouzi: data `lxd_project` and data `lxd_profile` are plain read data sources but the real objects in LXD API can have been created outside TF (not managed by TF plan)

Won't they error if the resource doesn't exist ?

Switched to `external` type with bash commands for now

test/e2e/main.tf
```diff
@@ -3,17 +3,31 @@ data "lxd_info" "cluster" {
   remote = var.remote
 }
 
-# Project
+# Check whether the project already exists
+data "external" "project_exists" {
+  program = ["bash", "-c", "lxc project list \"${var.remote}:\" -f csv | awk -F, '{gsub(/ \\(current\\)$/, \"\", $1); print $1}' | grep -qxF \"${var.project}\" && echo '{\"exists\": \"true\"}' || echo '{\"exists\": \"false\"}'"]
```
ggouzi, line 8: Used AI for this ugly command FYI.

roosterfish: Mh, if there is no easy way to achieve this, I would say let's have an additional TF var as you have proposed right from the start. 
Doing this seems to be not the right way.

roosterfish: @MusicDin do you know if there is a way to check for an existing project/profile and create if it doesn't yet exist (in terraform)?

PR #1505, opened 2026-08-18
Init: Set OVN connection string only if necessary

### LINKED ISSUE #1286: Don't set `network.ovn.northbound_connection` if not required
Once https://github.com/canonical/lxd/pull/17874 lands in LXD, MicroCloud doesn't need to set the  `network.ovn.northbound_connection` config key anymore as LXD can retrieve it directly from MicroOVN. 

MicroCloud should check for the `ovn_dynamic_northbound_connection` API extension and skip setting `network.ovn.northbound_connection` if present.
In case it's already set, unset the `network.ovn.northbound_connection` config key when the API extension is present.

roosterfish:
Fixes https://github.com/canonical/microcloud/issues/1286

Starting the with the ovn_dynamic_northbound_connection API extension in LXD, we don't anymore have to manually set the OVN northbound DB connection string. This also has the benefit that in case it requires a change, LXD always picks up the latest info from  file setup by MicroOVN.

### DIFF
--- a/cmd/microcloud/main_init.go
+++ b/cmd/microcloud/main_init.go
@@ -885,8 +885,18 @@ func (c *initConfig) setupCluster(s *service.Handler) error {
 
 	fmt.Println("Configuring cluster-wide devices ...")
 
-	var ovnConfig string
-	if s.Services[types.MicroOVN] != nil {
+	// Update LXD's global config.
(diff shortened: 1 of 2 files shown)


nmezhenskyi (approved):
LGTM!

PR #1519, opened 2026-08-21
client, test/suites: forward session timeout cause over websocket

### LINKED ISSUE #1245: Session expiry isn't forwarded to the client
When the join session expires, the server hangs up on the client (websocket) and logs the following error `Session timeout exceeded`.
This exact error should also be visible on the client side. But instead the client seems to see only `⨯ Error: Failed to read join confirmation: websocket: close 1006 (abnormal closure): unexpected EOF` which is misleading and indicates we don't properly handle close on the websocket client side.

shardool-patil:
### Description
When a trust establishment session expires, the client previously received a generic connection error (`websocket: close 1006: unexpected EOF`) instead of the actual cause (`Session timeout exceeded`)[cite: 5].

### Changes
* **`client/websocket.go`**:
  * Updated `WriteClose` to write `ControlClose` directly over the WebSocket under `writeLock` so close frames are sent even when `gwCtx` is canceled on timeout[cite: 5].
  * Updated `ReceiveWithContext` to check channel closure and return `context.Cause(w.ctx)`[cite: 5].
* **`test/suites/basic.sh`**:
  * Added a test case in `test_interactive` with `SESSION_TIMEOUT="5"` verifying that both the initiator and joiner receive `"Session timeout exceeded"`[cite: 2].

## Checklist

### Related issues
- Fixes #1245

@roosterfish i have taken into consideration the insights you provided and resolved the issue, so take a look at it when you get time.

### DIFF
--- a/client/websocket.go
+++ b/client/websocket.go
@@ -82,8 +82,13 @@ func NewWebsocketGateway(ctx context.Context, conn *websocket.Conn) *WebsocketGa
 			decoder.DisallowUnknownFields()
 			err = decoder.Decode(&controlClose)
 			if err == nil {
-				// Cancel the inner context with the respective error.
-				defer gwCancel(errors.New(controlClose.ControlMessage))
+				var closeErr error
+				if controlClose.ControlMessage != "" {
+					closeErr = errors.New(controlClose.ControlMessage)
+				}
+
+				// Cancel the inner context too with the respective error.
+				defer gwCancel(closeErr)
 				return
 			}
 
@@ -109,8 +114,16 @@ func (w *WebsocketGateway) Receive() <-chan []byte {
 func (w *WebsocketGateway) ReceiveWithContext(ctx context.Context, v any) error {
 	var err error
 	select {
-	case bytes := <-w.Receive():
-		err = json.Unmarshal(bytes, v)
+	case data, ok := <-w.Receive():
+		if !ok {
+			if w.ctx.Err() != nil {
+				return context.Cause(w.ctx)
+			}
+
+			return errors.New("WebSocket connection closed")
+		}
+
+		err = json.Unmarshal(data, v)
 	case <-w.ctx.Done():
 		err = context.Cause(w.ctx)
 	case <-ctx.Done():
@@ -131,20 +144,20 @@ func (w *WebsocketGateway) Write(v any) error {
 	w.writeLock.Lock()
 	defer w.writeLock.Unlock()
 
-	if w.ctx.Err() != nil {
-		return context.Cause(w.ctx)
-	}
-
 	return w.conn.WriteJSON(v)
 }
 
 // WriteClose sends our websocket control close message.
-// Unlike the actual websocket control close message this supports message longer than 125 bytes
+// Unlike the actual websocket control close message this supports messages longer than 125 bytes
 // as well as special characters.
-// It waits for the other side to hang up or the gateway's context being cancelled.
 func (w *WebsocketGateway) WriteClose(err error) error {
+	var msg string
+	if err != nil {
+		msg = err.Error()
+	}
+
 	writeErr := w.Write(ControlClose{
-		ControlMessage: err.Error(),
+		ControlMessage: msg,
 	})
 	if writeErr != nil {
 		return fmt.Errorf("Failed to write control message: %w", writeErr)
--- a/test/suites/basic.sh
+++ b/test/suites/basic.sh
@@ -440,6 +440,22 @@ test_interactive() {
   lxc exec micro01 -- tail -1 out | grep "User aborted" -q
   lxc exec micro04 -- tail -1 out | grep "Failed waiting during join: Initiator aborted the setup" -q
 
+  # Initiate a MicroCloud cluster with a session timeout and verify both initiator and joiner receive "Session timeout exceeded".
+  reset_systems 3 0 0
+  unset_interactive_vars
+
+  echo "Initiate a MicroCloud cluster with a short session timeout and verify both initiator and joiner receive the timeout error"
+  export MULTI_NODE="yes"
+  export LOOKUP_IFACE="enp5s0"
+  export EXPECT_PEERS=2
+  export SETUP_ZFS="no"
+  export SETUP_CEPH="no"
+  export OVN_WARNING="no"
+
+  ! join_session init micro01 micro02 || false
+  lxc exec micro01 -- tail -1 out | grep "Session timeout exceeded" -q
+  lxc exec micro02 -- tail -1 out | grep "Session timeout exceeded" -q
+
   echo "Try to initiate a MicroCloud with left over networking resources and check it fails to configure OVN"
   reset_systems 3 2 1
 
@@ -622,7 +638,6 @@ _test_case() {
       validate_system_microovn "${name}"
     done
 }
-
 
 test_interactive_combinations() {
   # Test with 2 systems, no disks, no interfaces.


test/suites/basic.sh
```diff
@@ -440,6 +440,25 @@ test_interactive() {
...
   lxc exec micro04 -- tail -1 out | grep "Failed waiting during join: Initiator aborted the setup" -q
 
+  # Initiate a MicroCloud cluster with a short session timeout and verify the joiner receives "Session timeout exceeded".
+  reset_systems 2 3 0
+  unset_interactive_vars
+
+  echo "Initiate a MicroCloud cluster with a short session timeout and verify the joiner receives the timeout error"
+  export MULTI_NODE="yes"
+  export LOOKUP_IFACE="enp5s0"
+  export EXPECT_PEERS=1
```
roosterfish: I guess what you would need here is `export EXPECT_PEERS=2`, then let only `micro02` attempt to join. As `micro01` is the initiator, it will wait for two peers, but only `micro02` reaches out so it will cause the session to expire.

test/suites/basic.sh
```diff
@@ -1500,4 +1518,4 @@ EOF
     validate_system_microceph ${m} 1 "disk2" "disk3"
     validate_system_microovn ${m}
   done
-}
+}
```
roosterfish: Newline missing.

test/suites/basic.sh
```diff
@@ -440,6 +440,25 @@ test_interactive() {
   lxc exec micro01 -- tail -1 out | grep "User aborted" -q
   lxc exec micro04 -- tail -1 out | grep "Failed waiting during join: Initiator aborted the setup" -q
 
+  # Initiate a MicroCloud cluster with a short session timeout and verify the joiner receives "Session timeout exceeded".
+  reset_systems 2 3 0
```
roosterfish: 2 is the number of peers, 3 number of disks, 0 number of extra interfaces.

As we never reach the config stage and don't require a disk, you can also do `2 0 0` here.

roosterfish (changes requested):
Thanks for taking a look at this, please see my comments on the test.

shardool-patil:
@roosterfish i have addressed all the requested changes:

* **`test/suites/basic.sh`**: Set `EXPECT_PEERS=2`, clarified the test comment and echo message to verify both initiator and joiner, and added the missing EOF newline[cite: 2, 3].
* **`test/includes/microcloud.sh`**: Updated `microcloud_interactive` to pass `${SESSION_TIMEOUT:-60}` dynamically and added `SESSION_TIMEOUT` to `unset_interactive_vars`[cite: 2].

take a look whenever possible

roosterfish:
See the test failure https://github.com/canonical/microcloud/actions/runs/32477357735/job/96766897819?pr=1519#step:3:45500. 
I suspect what is happening is that the session expires too early now. Maybe let's leave the 60s default (no need to make it configurable in this case).

shardool-patil:
> See the test failure https://github.com/canonical/microcloud/actions/runs/32477357735/job/96766897819?pr=1519#step:3:45500. I suspect what is happening is that the session expires too early now. Maybe let's leave the 60s default (no need to make it configurable in this case).

@roosterfish I have worked and handled the requested changes so take a look at the code.

shardool-patil:
@roosterfish I switched the test to `reset_systems 3 0 0` because with `reset_systems 2 0 0`, `micro03` is paused by default. Since `EXPECT_PEERS=2` triggers `table:expect 2`, `micro01` hangs during mDNS peer discovery waiting for a nonexistent second peer before the session ever starts. Running `reset_systems 3 0 0` allows `micro01` to discover both peers, select them, connect with `micro02`, and wait for `micro03` until the session timeout triggers. Does `reset_systems 3 0 0` work for you here?

shardool-patil:
@roosterfish I've updated `test/suites/basic.sh`:
- Switched to `reset_systems 3 0 0` and disabled storage setup (`SETUP_ZFS="no"`, `SETUP_CEPH="no"`).
- Kept `EXPECT_PEERS=2` with `join_session init micro01 micro02` so `micro01` connects with `micro02` and waits for `micro03` until the default 60s session timeout triggers.
With `reset_systems 2 0 0`, `micro03` is paused by default, causing `micro01` to hang during mDNS peer discovery (`table:expect 2`) before the session can start. `3 0 0` keeps `micro03` running so discovery succeeds without configuring any unused storage devices. 

Please take a look at this setup before re-running the tests.

client/websocket.go
```diff
@@ -51,7 +51,10 @@ func NewWebsocketGateway(ctx context.Context, conn *websocket.Conn) *WebsocketGa
 
 		// Send close control message.
 		// Try to send the cause from the outer context if present.
-		_ = gw.WriteClose(context.Cause(gwCtx))
+		cause := context.Cause(gwCtx)
```
roosterfish: Don't we always have to write the close message? From what I remember it's important that it's of type `ControlClose`, but we don't have to populate the `ControlMessage` if there isn't any error.

roosterfish: Here we will only ever send the control message if `context.Cause` returns a non-nil error.

client/websocket.go
```diff
@@ -139,20 +150,22 @@ func (w *WebsocketGateway) Write(v any) error {
 }
 
 // WriteClose sends our websocket control close message.
-// Unlike the actual websocket control close message this supports message longer than 125 bytes
+// Unlike the actual websocket control close message this supports messages longer than 125 bytes
 // as well as special characters.
-// It waits for the other side to hang up or the gateway's context being cancelled.
 func (w *WebsocketGateway) WriteClose(err error) error {
-	writeErr := w.Write(ControlClose{
+	w.writeLock.Lock()
```
roosterfish: Why the extra lock? By invoking `w.Write`, it already obtains the lock for us. I don't see why we have to use `w.conn.WriteJSON` directly here.

test/includes/microcloud.sh
```diff
@@ -891,7 +891,7 @@ cluster_reset() {
   (
     set -eu
     if [ "${SKIP_SETUP_LOG}" = 1 ]; then
-      exec > /dev/null 2>&1
+      exec > /dev/null
```
roosterfish: Why is this necessary?

test/includes/microcloud.sh
```diff
@@ -935,8 +935,8 @@ cluster_reset() {
             microceph.ceph osd out \${osd}
             microceph.ceph osd down \${osd} --definitely-dead
             pkill -f \"ceph-osd .* --id \${osd}\"
-            microceph.ceph osd purge \${osd} --yes-i-really-mean-it --force
-            microceph.ceph osd destroy \${osd} --yes-i-really-mean-it --force
+            microceph.ceph osd purge \${osd} --yes-i-really-really-mean-it --force
```
roosterfish: Why is this necessary? According to `ceph osd purge -h` it's `--yes-i-really-mean-it`.

roosterfish (changes requested):

client/websocket.go
shardool-patil: thanks for pointing that out, i have removed the manual lock and reverted to write close

test/includes/microcloud.sh
shardool-patil: this was an accidental edit left from local debugging, reverted microcloud.sh back to main.

test/includes/microcloud.sh
shardool-patil: same here.

test/suites/basic.sh
```diff
@@ -440,6 +440,22 @@ test_interactive() {
   lxc exec micro01 -- tail -1 out | grep "User aborted" -q
   lxc exec micro04 -- tail -1 out | grep "Failed waiting during join: Initiator aborted the setup" -q
 
+  # Initiate a MicroCloud cluster with a session timeout and verify both initiator and joiner receive "Session timeout exceeded".
```
roosterfish, line 443: Please make sure the changes are separated into multiple commits (as you did before).

shardool-patil: i have separated the commits so take a look @roosterfish

client/websocket.go
```diff
@@ -131,28 +144,24 @@ func (w *WebsocketGateway) Write(v any) error {
...
+
 	writeErr := w.Write(ControlClose{
-		ControlMessage: err.Error(),
+		ControlMessage: msg,
 	})
 	if writeErr != nil {
 		return fmt.Errorf("Failed to write control message: %w", writeErr)
 	}
 
-	// Wait on the other end to hang up.
```
roosterfish, line 153: Why don't we have to wait anymore for the other site to hang up in case we signal close of the websocket?

shardool-patil: While debugging earlier test failures, I removed that trying to simplify WriteClose, but realized in hindsight that was a mistake. We definitely still need to wait for the other end to hang up so the connection isn't severed before the close frame is processed. I've restored the wait logic, so take a look at this when you get time @roosterfish

shardool-patil:
@roosterfish just a ping, to check the status

PR #1522, opened 2026-08-24
Test: Add release branches

roosterfish:
Add `release-*` branches to the various tests and also add the version branches for the sake of completeness. They are [already set](https://github.com/canonical/microcloud/blob/v2-edge/.github/workflows/tests.yml#L7) on the respective files in the version branch. But this makes backporting easier.


tomponline (approved):

tugbataluy (approved):
LGTM, thanks ❇️

PR #1538, opened 2026-09-03
doc: configure docs to generate llms.txt

elijahgreenstein:
Adds a Sphinx extension to auto-generate `llms.txt` and `llms-full.txt` files. Extension documentation is available here:
https://sphinx-llms-txt.readthedocs.io/en/latest/


doc/conf.py
```diff
@@ -128,6 +129,22 @@
...
 
+############################
+# sphinx-llm configuration #
+############################
+
+# This description is included in llms.txt to provide some initial context for your
+# product docs.
+llms_txt_description = textwrap.dedent(
+    """\
+    This is the documentation for MicroCloud, an open source cloud platform.
```
roosterfish, line 140: Worth using any of the short descriptions we already actively use? 

* GH project desc: "Automated private cloud based on LXD, Ceph and OVN"
* Product page: "MicroCloud: your open source cloud platform"
* Docs: "Deploy a low-touch, open source cloud platform in minutes with MicroCloud."

roosterfish (changes requested):

doc/conf.py
elijahgreenstein: I think we want to keep the "This is the documentation for" part of the description, to provide context for what is in the `llms.txt` file (though I don't believe there's consensus about how LLMs actually "use" these files). This `llms_txt_description` only appears at the top of the `llms.txt` file, which is a list of the docs pages, with titles and descriptions from the page metadata. You can see it in the RTD build here: https://canonical-microcloud--1538.com.readthedocs.build/1538/llms.txt. 

It might be worth bringing in a few more keywords, though. How about an adaptation of the GH project desc?

> This is the documentation for MicroCloud: automated private cloud based on LXD, Ceph, and OVN.

roosterfish: Your proposal sounds good to me.

elijahgreenstein: Updated

roosterfish (approved):
Thanks

tugbataluy (approved):

PR #1539, opened 2026-09-09
feat: allow override to lxd ui url sent to cluster manager

### LINKED ISSUE #1527: Cluster Manager "LXD UI" link points to a single member's internal address — No existing path to advertise an HA/load-balancer URL
The "LXD UI" button on the Cluster Manager's cluster details page links to the first cluster member's internal cluster address (the cluster.https_address visible in lxc cluster list). This was found while running MicroCloud demo environments on GCP, where each cluster sits behind a TCP load balancer with a static address across all members (the address operators actually use), but the button links to one member's private IP instead.

Tracing the behavior, the value appears to originate in the connector's status heartbeat: enrichClusterMemberMetrics in cmd/microcloudd/cluster_manager_task.go sets ui_url from lxdMembers[0].URL, and the Cluster Manager stores and renders it as-is (in microcloud-cluster-manager: RemoteClusterStatusPost → remote_cluster_detail.go, refreshed on each heartbeat → ClusterUiButton.tsx). No configuration point was found along that path where an operator-supplied URL could be provided.

Concerns related to current limitations:
1. HA: production deployments front the UI with a VIP/load balancer across all members. Linking to one specific member means the button dies when that member is down for maintenance, even though the cluster and its LB remain healthy.

2. TLS: a member's private IP cannot carry the certificate an operator has issued for their DNS-named LB endpoint, so the link always produces browser warnings, even when the deployment has a perfectly clean HTTPS entry point.

3. Segmented networks: deployments that keep cluster-internal addresses on an isolated management network (exposing only the LB/DNS endpoint to operators) end up with a link that isn't reachable from the browsers meant to use it.

Request: a way to set the UI URL a cluster reports to the Cluster Manager, for example the load balancer's address or DNS name. If nothing is set, the current behavior stays the default. This could potentially be implemented in a preseed field or join field, or another mechanism that would work best.

omarelkashef:
## Done

Allow override to lxd ui url sent to cluster manager

Fixes #1527 

## Checklist


api/types/cluster_manager.go
```diff
@@ -60,6 +62,26 @@ type ClusterManagerPut struct {
 	// Enables or disables the reverse tunnel to the cluster manager
 	// Example: true, false
 	ReverseTunnel *bool `json:"reverse_tunnel" yaml:"reverse_tunnel"`
+
+	// The URL of the LXD UI
+	// Example: https://example.com/ui
+	LxdUIURL *string `json:"lxd_ui_url" yaml:"lxd_ui_url"`
```
roosterfish: I wonder if this new config setting should (and can) be used to override in general how to reach the cluster. Meaning don't include the "ui" bits in there, it will likely also have an effect on how to reach LXD's API? 

What do you think about `LXDURL`?

api/types/cluster_manager.go
```diff
@@ -60,6 +62,26 @@ type ClusterManagerPut struct {
...
+}
+
+// ValidateLxdUIURL checks that value is an absolute HTTP or HTTPS URL with a host.
+// A bare host or IP (e.g. "192.168.1.10:8443") without an "http://" or "https://" prefix is
+// rejected.
+func ValidateLxdUIURL(value string) error {
+	u, err := url.ParseRequestURI(value)
+	if err != nil || u.Scheme == "" || u.Host == "" {
+		return fmt.Errorf("Invalid value %q for lxd_ui_url: expected an absolute URL starting with http:// or https://, for example https://192.168.1.10:8443/ui", value)
```
roosterfish: Can we show the actual parsing error?

`return fmt.Errorf("Invalid value %q: %w", value, err)`

roosterfish: Also make the `ValidateLxdUIURL` generic so that the caller can then add further context like `lxd_ui_url` to the error message.

api/cluster_manager.go
```diff
@@ -196,16 +196,28 @@ func clusterManagerPut(state types.State, r *http.Request) types.Response {
...
 	}
 
 	if args.ReverseTunnel != nil {
-		reverseTunnelValue := strconv.FormatBool(*args.ReverseTunnel)
-		err = database.StoreClusterManagerConfig(state, r.Context(), name, database.ReverseTunnelKey, reverseTunnelValue)
+		changedConfigs[database.ReverseTunnelKey] = strconv.FormatBool(*args.ReverseTunnel)
+	}
+
+	if len(changedConfigs) > 0 {
+		err = database.StoreClusterManagerConfigs(state, r.Context(), name, changedConfigs)
```
roosterfish, line 220: I see the reasoning behind this change to call `StoreClusterManagerConfigs` only once, please move this into a separate commit.

cmd/microcloud/cluster_manager.go
```diff
@@ -303,6 +310,13 @@ func (c *cmdClusterManagerSet) run(_ *cobra.Command, args []string) error {
 		payload.CertificateFingerprint = &value
 	case "update_interval_seconds":
 		payload.UpdateIntervalSeconds = &value
+	case "lxd_ui_url":
+		err := types.ValidateLxdUIURL(value)
```
roosterfish: I think it's ok to run the validation only on the API.

cmd/microcloudd/cluster_manager_task.go
```diff
@@ -8,7 +8,7 @@ import (
 	"slices"
 	"time"
 
-	"github.com/canonical/lxd/client"
+	lxd "github.com/canonical/lxd/client"
```
roosterfish: We usually don't commit this in any of the files. Please remove.

api/types/cluster_manager.go
roosterfish: Or we follow the already existing `ClusterManagerPostStatus` which uses `UIURL`.

database/cluster_manager_config.go
```diff
@@ -147,31 +169,37 @@ func StoreClusterManagerConfig(state types.State, ctx context.Context, name stri
 		return err
 	}
 
-	var existingConfig *ClusterManagerConfig = nil
+	existingConfigs := make(map[string]ClusterManagerConfig, len(clusterManagerConfig))
```
roosterfish, line 172: Thanks, can this please also be in a separate commit?

database/cluster_manager_config.go
```diff
@@ -147,31 +169,37 @@ func StoreClusterManagerConfig(state types.State, ctx context.Context, name stri
...
-				ClusterManagerID: clusterManager.ID,
-				Key:              key,
-				Value:            value,
-			})
-		} else if value != "" && existingConfig != nil {
-			// update
-			existingConfig.Value = value
-			err = UpdateClusterManagerConfig(ctx, tx, existingConfig.ID, *existingConfig)
+		for key, value := range configs {
+			existingConfig, hasExistingConfig := existingConfigs[key]
```
roosterfish, line 179: If I read this right, is there a bug currently with resetting config keys as we are not checking the specific config key?

test/suites/cluster-manager.sh
```diff
@@ -87,6 +87,34 @@ PYEOF"
   lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager set update_interval_seconds 60
   lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager show | grep "certificate_fingerprint:" -q
 
+  echo "==> LXD UI URL config for cluster manager"
```
roosterfish, line 90: Please move this also into a separate commit.

test/suites/cluster-manager.sh
```diff
@@ -87,6 +87,34 @@ PYEOF"
...
+
+  echo "==> Reject lxd_ui_url values with a disallowed scheme"
+  ! lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager set lxd_ui_url ftp://example.com || false
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager set lxd_ui_url ftp://example.com 2>&1 | grep "URL scheme must be http or https" -q
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager set lxd_ui_url https://example.com/ui
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager get lxd_ui_url | grep "https://example.com/ui" -q
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager show | grep "lxd_ui_url: https://example.com/ui" -q
+
+  echo "==> Accept http lxd_ui_url values"
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager set lxd_ui_url http://example.com/ui
```
roosterfish: This seems to be redundant, there is a set/get already a few lines before. Worth removing these but adding the `| grep "lxd_ui_url: https://example.com/ui" -q` case here underneath `  echo "==> Accept http lxd_ui_url values"` ?

roosterfish (changes requested):
Thanks for this and adding tests. Please also make sure to split the changes into separate commits based on the package in which you are introducing the change. As an example see https://github.com/canonical/microcloud/pull/1529/commits.

roosterfish:
Also the tests you have added seem to be failing atm.

api/types/cluster_manager.go
omarelkashef: It might be a good idea to keep general and set it as `LXDURL`.

database/cluster_manager_config.go
omarelkashef: Can you elaborate? We check the incoming configs, retrieve any existing configs with matching keys, and clear/reset if the value is "".

roosterfish: Nvm, you changed the functions signature to pass multiple config keys and before it was only one.

roosterfish (changes requested):
Commits are not signed and the Signed-off-by line is missing which causes the DCO check to fail.

api/cluster_manager.go
```diff
@@ -196,16 +196,17 @@ func clusterManagerPut(state types.State, r *http.Request) types.Response {
...
```
roosterfish, line 240: Just a nit, its easier to review when the change to the `StoreClusterManagerConfigs` func is introduced before calling it with the updated argument list passing a map instead of a single key/value pair.

api/types/cluster_manager.go
```diff
@@ -60,6 +62,30 @@ type ClusterManagerPut struct {
...
+
+	// The URL of the LXD UI
+	// Example: https://example.com/ui
+	LXDURL *string `json:"lxd_url" yaml:"lxd_url"`
+}
+
+// ValidateLXDURL checks that value is an absolute HTTP or HTTPS URL with a host.
+// A bare host or IP (e.g. "192.168.1.10:8443") without an "http://" or "https://" prefix is
+// rejected.
+func ValidateLXDURL(value string) error {
```
roosterfish: Can you please define this func closer to where it is actually used? Also does it need to be exported?

api/types/cluster_manager.go
```diff
@@ -60,6 +62,30 @@ type ClusterManagerPut struct {
...
+	u, err := url.ParseRequestURI(value)
+	if err != nil {
+		return err
+	}
+
+	if u.Scheme == "" || u.Host == "" {
+		return fmt.Errorf("Invalid value %q: expected an absolute URL starting with http:// or https://, for example https://192.168.1.10:8443/ui", value)
```
roosterfish: We usually don't put examples in the error message, can we please just do `fmt.Errorf("URL %q is missing a scheme or host", value)` to just indicate the problem?

api/types/cluster_manager.go
```diff
@@ -60,6 +62,30 @@ type ClusterManagerPut struct {
...
+	}
+
+	if u.Scheme != "http" && u.Scheme != "https" {
+		return fmt.Errorf("Invalid value %q: URL scheme must be http or https", value)
```
roosterfish: return fmt.Errorf("URL scheme %q must be http or https", value)

api/cluster_manager.go
```diff
@@ -197,6 +197,17 @@ func clusterManagerPut(state types.State, r *http.Request) types.Response {
 	}
 
 	changedConfigs := make(map[string]string)
+	if args.LXDURL != nil {
```
roosterfish, line 220: Why not just `if args.LXDURL != nil {`? What if you want to unset the URL?

database/cluster_manager_config.go
```diff
@@ -117,6 +120,20 @@ func LoadClusterManagerReverseTunnel(state types.State, ctx context.Context, clu
...
 
+// LoadLXDURL loads the LXD UI URL configuration from the database.
+func LoadLXDURL(state types.State, ctx context.Context, clusterManagerId int64) (string, error) {
+	uiUrlConfig, err := LoadClusterManagerSingleConfig(state, ctx, clusterManagerId, LXDURLKey)
+	if err != nil {
+		return "", err
+	}
+
+	if uiUrlConfig == nil {
+		return "", api.StatusErrorf(http.StatusNotFound, "UI URL not found")
```
roosterfish: "`LXD URL` not found" to stay consistent with the key?

api/types/cluster_manager.go
```diff
@@ -60,6 +62,30 @@ type ClusterManagerPut struct {
 	// Enables or disables the reverse tunnel to the cluster manager
 	// Example: true, false
 	ReverseTunnel *bool `json:"reverse_tunnel" yaml:"reverse_tunnel"`
+
+	// The URL of the LXD UI
```
roosterfish: Based on https://github.com/canonical/microcloud/pull/1539#discussion_r3976282865, should we also drop mentioning UI here?

Do we expect users to add `.../ui` or only set the LXD address without any path involved? I suspect we want the latter so that this config key is true for both API and UI.

roosterfish (changes requested):

api/types/cluster_manager.go
omarelkashef: Yeah, there is no need for /ui at the end. It is added automatically when the users are redirected in the browser. So it is a good idea to keep it general and drop the UI from here.

api/cluster_manager.go
omarelkashef: Unsetting uses an empty string.

cmd/microcloud/cluster_manager.go
```diff
@@ -242,6 +243,12 @@ func (c *cmdClusterManagerGet) run(_ *cobra.Command, args []string) error {
 			fmt.Printf("%s\n", value)
 		}
 
+	case "lxd_url":
+		value, ok := clusterManager.Config[database.LXDURLKey]
+		if ok {
+			fmt.Printf("%s\n", value)
```
roosterfish, line 249: Juts a nit for the future, I think we can simplify this and do `fmt.Println(value)` for all of the config keys.

roosterfish: Not a blocker though as this pattern is already used.

api/types/cluster_manager.go
```diff
@@ -100,7 +100,7 @@ type ClusterManagerPostStatus struct {
 	MemberStatuses    []StatusDistribution `json:"member_statuses"`
 	InstanceStatuses  []StatusDistribution `json:"instance_statuses"`
 	ServerMetrics     []ServerMetrics      `json:"server_metrics"`
-	UIURL             string               `json:"ui_url"`
+	LXDURL            string               `json:"lxd_url"`
```
roosterfish, line 103: Is there an equivalent change on cluster manager for this? Maybe you can check for an API extension?

@omarelkashef @edlerd I think there isn't yet a LTS release of cluster manager, so do you think we should just change this field here and on cluster manager side?

api/cluster_manager.go
```diff
@@ -161,6 +161,26 @@ func clusterManagerPost(sh *service.Handler) func(state types.State, r *http.Req
...
+// validateLXDURL checks that value is an absolute HTTP or HTTPS URL with a host.
+// A bare host or IP (e.g. "192.168.1.10:8443") without an "http://" or "https://" prefix is
+// rejected.
+func validateLXDURL(value string) error {
+	u, err := url.ParseRequestURI(value)
+	if err != nil {
+		return err
+	}
+
+	if u.Scheme == "" || u.Host == "" {
```
roosterfish: In your tests I see that `url.ParseRequestURI` already errors in case there is no scheme. Is this true also for `u.Host`? In this case maybe you can omit this check if its already covered.

roosterfish (changes requested):
Looking good, just two more comments.

api/types/cluster_manager.go
omarelkashef: Yes, there is already a pr that will be merged after this https://github.com/canonical/microcloud-cluster-manager/pull/641

roosterfish: Ok thanks.

api/cluster_manager.go
omarelkashef: Good catch, but actually both checks are necessary because ParseRequestURI does not validate everything we need. The comment for one of the tests was misleading which I fixed and added two more test cases.

roosterfish (dismissed):
Nice thanks!

api/cluster_manager.go
```diff
@@ -196,16 +216,28 @@ func clusterManagerPut(state types.State, r *http.Request) types.Response {
...
-		err = database.StoreClusterManagerConfig(state, r.Context(), name, database.UpdateIntervalSecondsKey, *args.UpdateIntervalSeconds)
-		if err != nil {
-			return types.SmartError(err)
+	changedConfigs := make(map[string]string)
+	if args.LXDURL != nil {
+		if *args.LXDURL != "" {
+			err = validateLXDURL(*args.LXDURL)
+			if err != nil {
+				return types.BadRequest(fmt.Errorf("Invalid lxd_url: %w", err))
+			}
```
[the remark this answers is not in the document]
omarelkashef: Addressed: The PUT handler now validates all fields (including  lxd_url ) before persisting anything, so an invalid request returns 400 with no side effects. Both the cluster manager record and its config changes are applied atomically in a single database transaction via a new  StoreClusterManagerWithConfigs  helper, which also skips the record update when only configs changed.

test/suites/cluster-manager.sh
```diff
@@ -87,6 +87,49 @@ PYEOF"
...
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager get lxd_url | grep -F "http://example.com:8443" -q
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager show | grep -F "lxd_url: http://example.com:8443" -q
+
+  echo "==> Accept https lxd_url values"
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager set lxd_url https://example.com:8443
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager get lxd_url | grep -F "https://example.com:8443" -q
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager show | grep "lxd_url: https://example.com:8443" -q
+
+  echo "==> Reject lxd_url values that are not absolute URLs"
+  lxc exec micro01 --env TEST_CONSOLE=0 -- microcloud cluster-manager set lxd_url example.com/ui 2>&1 | grep -F "Invalid lxd_url: parse \"example.com/ui\": invalid URI for request" -q
```
edlerd: Can we ensure the error message guides the user to the correct url format?

omarelkashef: This is one of the cases that come from `url.ParseRequestURI` and we return the error directly.

edlerd: Can we amend the raw message so a user is guided to how to resolve the error?

omarelkashef: We do not know what the exact error is: it might be "not absolute path" or "no scheme", etc. The only way we can provide a helpful error messages in these cases is to give an example of the valid url, but Julian said that this pattern is not followed anywhere else. https://github.com/canonical/microcloud/pull/1539#discussion_r4014074058

edlerd: Can we add our own custom check? If the string does not starts with `http` or `https` then throw the error `url must start with http:// or https://`?

omarelkashef: yeah, we can move code around and put our checks before `url.ParseRequestURI` to be able to provide better error messages.

edlerd (approved):
LGTM, thanks.

